PR #12, opened 2016-04-05
Push-button deployment script

ivan-c:
Here's my first pass at a deployment script, please don't hesitate to offer any critique or suggestions.

Notes:

- Runnable from any directory (pass git repo path with `-p`)
- Rudimentary detection of changes requiring:
    - Python package installation
    - Database seeding, migration
    - Webserver restart

### DIFF
--- a/bin/deploy.sh
+++ b/bin/deploy.sh
@@ -98,7 +98,7 @@ if [[ -n $(git diff $old_head $new_head -- ${GIT_WORK_TREE}/migrations) && $? -e
 fi
 
 # Restart apache if application is served by apache
-if [[ "${GIT_WORK_TREE}" == "/srv/www/"* && -n $(git diff $old_head $new_head) && $? -eq 0 ]]; then
+if [[ "${GIT_WORK_TREE}" == "/srv/www/"* ]]; then
     if [[ $VERBOSE ]]; then
         echo "Restarting apache"
     fi
--- a/bin/deploy.sh
+++ b/bin/deploy.sh
@@ -60,7 +60,7 @@ update_repo
 new_head=$(git rev-parse origin/$BRANCH)
 
 # New modules, or new seed data
-if [[ -n $(git diff $old_head $new_head -- ${GIT_WORK_TREE}/setup.py) ]]; then
+if [[ -n $(git diff $old_head $new_head -- ${GIT_WORK_TREE}/setup.py) && $? -eq 0 ]]; then
     activate_once
 
     if [[ $VERBOSE ]]; then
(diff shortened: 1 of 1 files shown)


bin/deploy.sh
```diff
@@ -0,0 +1,99 @@
...
+    git pull origin $BRANCH
+}
+
+# Prevent reading virtualenv environmental variables multiple times
+activate_once(){
+    if [[ $(which python) != "${GIT_WORK_TREE}"* ]]; then
+        if [[ $VERBOSE ]]; then
+            echo "Activating virtualenv"
+        fi
+        source "${GIT_WORK_TREE}/bin/activate"
```
pbugni: we started with this path, and now typically have ${GIT_WORK_TREE}/env/bin/activate - should either make all virtual environments live in a nested 'env' directory, or make this tool smart enough to look in both locations.

bin/deploy.sh
```diff
@@ -0,0 +1,99 @@
...
+
+    if [[ $VERBOSE ]]; then
+        echo "Running database migrations"
+    fi
+    cd "${GIT_WORK_TREE}"
+    python "${GIT_WORK_TREE}/manage.py" db upgrade
+fi
+
+# Restart apache if application is served by apache
+if [[ "${GIT_WORK_TREE}" == "/srv/www/"* && -n $(git diff $old_head $new_head) && $? -eq 0 ]]; then
```
pbugni: I think we should restart apache even if old and new don't match - to pick up config or db or other changes

pbugni:
One more requirement I just remembered.

IFF the files celery uses are modified (namely portal/tasks.py), we need to reload celeryd.

bin/deploy.sh
```diff
@@ -0,0 +1,99 @@
...
+    if [[ $(which python) != "${GIT_WORK_TREE}"* ]]; then
+        if [[ $VERBOSE ]]; then
+            echo "Activating virtualenv"
+        fi
+        source "${GIT_WORK_TREE}/bin/activate"
+    fi
+}
+
+# Defaults
+BRANCH="develop"
```
pbugni: Not sure how to do this, but ideally BRANCH would default to 'develop' everywhere except on prod, where it defaults to 'master'

bin/deploy.sh
ivan-c: Would a single repo have multiple virtual environments? 
Should [the README](https://github.com/uwcirg/true_nth_usa_portal#create-a-virtual-environment) be updated with this new path (ie <code>$ virtualenv $PROJECT_HOME**/env**</code>)?

pbugni: Excellent point - I'd overlooked that till now.  There is a new README in the develop branch.

ivan-c:
@pbugni I think I've addressed all of your suggestions (thanks!), let me know if I've missed anything.

PR #15, opened 2016-04-15
Whitelist shared services URL for redirections

ivan-c:
This will allow redirections to shared services without having to add an additional entry to `clients._redirect_uris`


pbugni:
I like it.  Wish it was easy to get the scheme - but I think hardcoding to https is fine.

PR #16, opened 2016-04-15
Formalize Assessment Engine as Intervention

ivan-c:
These changes leverage existing intervention/client fields for generating URLs to the Assessment Engine. Previously, URLs were hard-coded for each instrument. 

As usual, please comment if there are better ways to store/query this information.


portal/views/assessment_engine.py
```diff
@@ -1095,16 +1096,21 @@ def present_assessment(instrument_id):
         description: if missing valid OAuth token or bad `next` parameter
 
     """
+    AE_intervention = Intervention.query.filter_by(name='assessment_engine').one()
```
pbugni: I should probably provide better access to the static list of interventions by object - this query will certainly work for now.  Might replace the string 'assessment_engine' with INTERVENTION.ASSESSMENT_ENGINE

Replace the import of STATIC_INTERVENTIONS with INTERVENTION

portal/views/assessment_engine.py
```diff
@@ -1095,16 +1096,21 @@ def present_assessment(instrument_id):
...
     clients_instruments = current_app.config['CLIENTS_INSTRUMENTS']
 
     if not instrument_id or not instrument_id in instruments:
         abort(404, "No matching assessment found: %s" % instrument_id)
 
-    for client_id, instrument_urls in clients_instruments.items():
-        if instrument_id in instrument_urls:
-            assessment_url = instrument_urls[instrument_id]
+    for client_id, instruments in clients_instruments.items():
+        if instrument_id in instruments and client_id == AE_intervention.client_id:
```
pbugni: This seems a little more fragile than I believe we need.  Why duplicate locations for setting the client_id of the assessment engine?  It lives in the database tables, do we also need to set it in the config (as I see in staging):
`ASSESSMENT_ENGINE_CLIENT_ID = 'UVAh1naZO8tztccxb98yj4XRSanAr8Zhzu4Zx5Kc'`

If we do need such flexibility, should it move to the database?

ivan-c:
@pbugni 
Fixed [issue in first comment](https://github.com/uwcirg/true_nth_usa_portal/pull/16#diff-for-comment-60319826)

[Removed most of this](https://github.com/uwcirg/true_nth_usa_portal/pull/16#diff-for-comment-60320814) - will likely replace with Questionnaires model/table when more mature

PR #28, opened 2016-06-02
Plugging security hole in login_url

pbugni:
Aligning behavior of protected and portal-header-wrapper.  Now identical
except that protected-portal-header-wrapper uses CORS.


pbugni:
Please don't merge this yet - conversation taking place over email!

PR #34, opened 2016-06-10
Use project metadata for populating documentation boilerplate

ivan-c:
Let me know if this seems like a bad idea!


pbugni:
Looks like a step in the right direction to me.  Thanks.

PR #35, opened 2016-06-14
Separate test dependencies

ivan-c:
This change moves the python modules that are only required for testing Shared Services into their own section in `setup.py` so they are no longer installed by default. 

Relevant files (eg `.travis.yml`) have also been updated.

Please let me know if I should move any other modules into this new section.


pbugni:
Very nice.  Can you please also update the README instructions?

I would think these packages are also only needed for tests:
- swagger_spec_validator
- Flask-WebTest

PR #41, opened 2016-06-20
Group endpoint spec

ivan-c:
This should fix the swagger documentation for the PUT `/user/<int:user_id>/groups` endpoint.

### DIFF
--- a/portal/views/user.py
+++ b/portal/views/user.py
@@ -176,6 +176,8 @@ def set_user_groups(user_id):
               type: array
               items:
                 type: object
+                required:
+                  - name
                 properties:
                   name:
                     type: string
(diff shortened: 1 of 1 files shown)


portal/views/user.py
```diff
@@ -170,15 +170,18 @@ def set_user_groups(user_id):
...
           properties:
-            name:
-              type: string
-              description:
-                The string defining the name of each group the user should
-                belong to.  Must exist as an available group in the system.
+            groups:
+              type: array
+              items:
+                type: object
```
ivan-c: You may be able to reference the `group_definition` definition/schema here instead of redefining it (ie `$ref: "#/definitions/group_definition"`). The only reason I didn't is the `group_definition` schema requires the `description` attribute while the example you gave me does not.

pbugni:
Perfect - thanks Ivan.

PR #42, opened 2016-07-01
Freeze requirements

ivan-c:
These changes pin the project dependencies to a known working state, preventing the newest, potentially unstable versions from being automatically installed.

To check if the newest dependencies are compatible a new tox virtual environment (`dependencies`) has been created.

[requires.io](https://requires.io/github/uwcirg/true_nth_usa_portal/requirements/) or [other similar services](https://github.com/integrations/versioneye) will also list if any dependencies are out of data


ivan-c:
Just learned of [pyup.io](pyup.io), a service that will [make pull requests](https://github.com/ivan-c/true_nth_usa_portal/pulls/pyup-bot) when individual dependencies are updated. 

Because pull requests are checked by CI we should know if we can integrate the update without error.

pbugni:
I'm not opposed to this change, it is probably the wise step for reliable, repeatable installs across systems.

Here's my concern:  once a requirements.txt file is in place, the project's dependencies tend to languish and avoid updating to new versions as they haven't been tested. Therefore, bitrot tends to increase as the project uses increasingly obsolete versions of 3rd party libraries.

Do you think we can enforce policy to keep our requirements.txt file as up to date as possible?

Can you also modify the README if we're going this route?

ivan-c:
>Do you think we can enforce policy to keep our requirements.txt file as up to date as possible?

Yes, I would be more than willing to take responsibility and pyup.io should help quite a bit. 

The service creates a PR every time a dependency is updated and should make updating dependencies truly a push-button operation.

pbugni:
Deal!

PR #43, opened 2016-07-01
Do not require application.cfg

ivan-c:
These changes allow a shared services instance to be introspected without an application.cfg file. Previously, documentation generation was producing errors as some config values were missing.

This also has the added benefit of obviating the need for a TravisCI-specific application.cfg file.

To prevent documentation errors strict error checking ([upgrading warnings into errors](http://www.sphinx-doc.org/en/stable/invocation.html#cmdoption-sphinx-build-W)) and [nitpick mode](http://www.sphinx-doc.org/en/stable/invocation.html#cmdoption-sphinx-build-n) have also been enabled on the tox `docs` virtual environment, but documentation can still be generated locally without such constraints.


pbugni:
Nice changes.  Perhaps we should move more values into environment variables and kill off the need for application.cfg all together?

I really don't care for the early_app_config_access() hack in particular (my own invention).  Tools like docker also play well with environment variables.  Do you (Ivan) think this would help or should we keep an application.cfg file?  We'd still want a clean way to organize environment vars for installations.

ivan-c:
*...As discussed earlier*

>Perhaps we should move more values into environment variables and kill off the need for application.cfg all together?

I like the idea of relying more on environmental variables but to simplify debugging I would prefer them to be opportunistically loaded, allowing values defined in `application.cfg` to have the highest precedence.


>I really don't care for the early_app_config_access() hack in particular (my own invention). Tools like docker also play well with environment variables. Do you (Ivan) think this would help or should we keep an application.cfg file? We'd still want a clean way to organize environment vars for installations.

I can't think of an elegant solution to this problem but would it be a start to allow any value in in `BaseConfig` to be overridden with an environmental variable of the same name? I should be able to do that with some `__init__()` magic.

PR #90, opened 2016-09-30
refactored patient_profile.html and profile.html to use same macros f…

achen2401:
I refactored the html content for both patient_profile.html and profile.html to use the common macros created in a new file named profile_macros.html in the flask_user folder within the templates main folder.    I am open to suggestions if these changes aren't ideal or need modifications.

Thank you!

Amy


pbugni:
Nice work, Amy.  I especially like how you packaged the relevant JS with the macro in question.

I'd suggest we remove the commented out blocks left behind (we always have git), such as those in the document ready block in patient_profile.html

I'm curious about the location choice for the new macros file.  The 'flask_user' directory exists, and is named as such, to enable customizing the 3rd party library Flask-User, which we use for local account management (i.e. outside of OAuth).  I think that directory should continue to house only changes related to that library.

Thank you!

achen2401:
Paul, I removed the commented out code from the html files and moved the macro file to the same directory as that for the patient_profile.html and profile.html.  Let me know if you see anything else that needs changed.  -Amy

PR #121, opened 2016-10-18
Feature/install instructions

ivan-c:
These commits add some of the missing steps in initially setting up a Shared Services instance. I've also fixed the deployment script so it can be run seamlessly after creating the DB and configuring the instance.


pbugni:
Looks great.  Merging.

PR #126, opened 2016-10-24
fix banner visibility issue in client apps

achen2401:


achen2401:
This fix addresses the issue of the visibility of banner in clients app pages in Truenth.

PR #131, opened 2016-10-26
fix datepicker browser incompatibility issue in challenge form

achen2401:
Fix the datepicker browser incompatibility issue - use boostrap datepicker object for all browsers since native date input type is not recognized by all
Add responsive styling to the challenge form


pbugni:
Nice job.

PR #133, opened 2016-10-26
fix datepicker format in challenge form to mm-dd-yyyy

achen2401:
Fix the datepicker field format to mm-dd-yyyy in boostrap datepicker object as well as in portal.py where birthdate field is now a Flask-WTF String Field - and converted to datetime field before the fuzzy match comparison.


pbugni:
Guess I missed this one earlier.  Merging now...

PR #136, opened 2016-10-27
Use requests for HL7 namespace fetch

OptimusRhine:
During db seeding, #fetch_HL7_V3_Namespace seems to fail on certain machines with a socket error (errno 104), because of the use of urllib. Switching from urllib to requests, which seems to resolve the issue.

### DIFF
--- a/portal/models/fhir.py
+++ b/portal/models/fhir.py
@@ -11,6 +11,7 @@ import urllib
 from ..extensions import db
 from .lazy import lazyprop
 from ..system_uri import TRUENTH_CLINICAL_CODE_SYSTEM
+import requests
 
 
 def as_fhir(obj):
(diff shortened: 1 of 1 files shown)


portal/models/fhir.py
```diff
@@ -451,8 +452,9 @@ def fetch_HL7_V3_Namespace(valueSet):
     """Pull and parse the published FHIR ethnicity namespace"""
     src_url = 'http://hl7.org/fhir/v3/{valueSet}/v3-{valueSet}.json'.format(
         valueSet=valueSet)
-    response = urllib.urlopen(src_url)
```
pbugni: As we're no longer using urllib - please remove the import line as well

OptimusRhine: Sure thing

PR #154, opened 2016-10-31
Changing cell/mobile in templates to apptext

OptimusRhine:
Changing any instances of 'cell' in the template files to an app_text call, where the custom_text can be modified to 'Cell' or 'Mobile' depending on if the instance is TrueNTH (USA) or ePROMS (AU). 

This is just temporary, until we have a functional i18n system in place.

### DIFF
--- a/portal/templates/profile_macros.html
+++ b/portal/templates/profile_macros.html
@@ -267,7 +267,7 @@
 
 {% macro profilePhone(person) -%}
     <div class="form-group float-input-label profile-section">
-        <label for="phone">Cell phone <span class="text-muted smaller-text">(Optional)</span></label>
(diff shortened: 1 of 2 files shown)


OptimusRhine:
@pbugni Ah, I get how the individual instance templates work now - I'll update those repos separately. For now, removed the tmp file change, and changed app_text name from 'Cell' to 'Cellphone'.

pbugni:
@OptimusRhine 

that looks great.

now, if you’ll add an AppText with name ‘Cellphone’ to both site_persistence files, it’ll all work!

eproms: https://github.com/uwcirg/ePROMs-site-config/blob/master/site_persistence_file.json
truenth: https://github.com/uwcirg/TrueNTH-USA-site-config/blob/master/site_persistence_file.json

OptimusRhine:
@pbugni Gotcha. It should be added to both site persistence files now.

PR #168, opened 2016-11-02
Revert "Update kombu to 4.0.0"

ivan-c:
Reverts uwcirg/true_nth_usa_portal#149

Some kind of compatibility issue between celery/kombu:

[`python manage db seed` output](https://github.com/uwcirg/true_nth_usa_portal/files/567112/db_seed.log.txt)

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -44,7 +44,7 @@ itsdangerous==0.24
 jedi==0.9.0
 Jinja2==2.8
 jsonschema==2.5.1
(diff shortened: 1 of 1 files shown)


ivan-c:
It looks like [kombu no longer provides OrderedDict for compatibility](https://github.com/celery/kombu/issues/632) (for older versions of python), but celery still expects it. Will need to pin this until celery is updated

PR #209, opened 2016-11-14
Updating locale methods + LR urls

OptimusRhine:
- Adding locale getter/setter methods, and modifying test_user accordingly
- Adding babel.get_locale() localeselector method
- Adding languageId parameter to liferay urls app_text calls
- Adding languageId parameter to ConsentATMA.permanent_url, and modifying test_app_text accordingly


portal/extensions.py
```diff
@@ -154,3 +154,9 @@ def init_app(self, app):
 # Bable is used for i18n
 from flask_babel import Babel
 babel = Babel()
+
+@babel.localeselector
```
pbugni: Hate to sound so particular, but the extensions module should remain as small as possible.  It exists to work around compile time dependency lookups - where you need for example a database, or library instance variable for inclusion.

I would assume get_locale() might better be associated with the user object?

portal/models/app_text.py
```diff
@@ -94,7 +94,7 @@ def name_key(**kwargs):
         return "{} organization consent URL".format(organization.name)
 
     @staticmethod
-    def permanent_url(generic_url, version):
+    def permanent_url(generic_url, version, languageId='en_US'):
```
pbugni: I would think passing language into every app_text call is a little cumbersome.  Would it not make more sense for app_text to be clever enough to determine this for itself?

PR #211, opened 2016-11-15
Feature/feature/delete user ui

achen2401:
- add delete button to Admin page for users who do not have admin or provider role

- bug fix for Patient List - to not add/change due date field and status field for deleted users

- validation email check for new user account

### DIFF
--- a/portal/templates/admin.html
+++ b/portal/templates/admin.html
@@ -73,14 +73,14 @@ function deleteUser(event, userId) {
        var c = confirm("Are you sure you want to delete this user?");
        if (c) {
          $.ajax ({
-              type: "GET",
-              url: "/admin/" + userId + "/delete",
+              type: "DELETE",
+              url: "/api/user/" + userId,
               contentType: 'application/json; charset=utf-8',
               dataType: 'json'
           }).done(function(data) {
-                if (data["error"] === "") {
+                if (data["message"] === "deleted") {
                     $("#data_row_" + userId).fadeOut();
-                } else alert("System error: " + data["error"]);
+                } else alert("System error: " + data["message"]);
           }).fail(function(xhr) {
                 console.log("response Text: " + xhr.responseText);
                 console.log("response status: " +  xhr.status);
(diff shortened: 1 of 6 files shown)


portal/views/portal.py
```diff
@@ -324,6 +325,19 @@ def admin():
         u.rolelist = ', '.join([r.name for r in u.roles])
     return render_template('admin.html', users=users, wide_container="true")
 
+@portal.route('/admin/<int:user_id>/delete', methods=['GET'])
+@oauth.require_oauth()
```
pbugni: please swap the order of these two lines.  if  @roles_required() comes before @oauth.require_oauth() it'll properly redirect when a call is attempted prior to authentication.

portal/views/patients.py
```diff
@@ -50,18 +50,20 @@ def patients_root():
...
-            if timedelta_days < 0:
-                desc = 'overdue'
-            elif timedelta_days < 30:
-                desc = 'due'
-            else:
-                desc = 'not due'
-            user.random_due_date_status = desc
-            user.due_date = user.due_date.strftime('%d %b %Y')
-            user.due_date = user.due_date.lstrip('0') # remove any leading 0 from day
+            if not user.deleted:
```
pbugni: This check isn't necessary - what is necessary is to avoid sending any deleted users into the patient list (my oversight).  I'll submit this change to this branch shortly.

portal/templates/admin.html
```diff
@@ -64,4 +66,26 @@ <h3 class="tnth-headline">User Administration</h3>
...
 });
 **/
+
+function deleteUser(event, userId) {
+    if (event) event.stopPropagation();
+    if (userId) {
+       var c = confirm("Are you sure you want to delete this user?");
+       if (c) {
+         $.ajax ({
+              type: "GET",
```
pbugni: We are attempting to stick with a RESTful API layout - I think a better endpoint would be to use the DELETE verb (in place of GET) on /api/user/{user_id}

PR #230, opened 2016-11-16
Feature/user l10n

OptimusRhine:
Full Babel implementation for user-specific translation:
- Removed previous gettext implementation (app.jinja_env.install_gettext_translations), as it was overwriting Babel with one set language
- Created babel config file
- Moving .mo & .po translation files to portal/translations/<lang_code>/LC_MESSAGES/messages.* (babel apparently requires a very specific filepath)
- Fixing user.locale_code method
- Added additional user locale method tests


portal/babel.cfg
```diff
@@ -0,0 +1,4 @@
+[python: **.py]
```
pbugni: As we're now setting (instance_relative_config=True)[https://github.com/uwcirg/true_nth_usa_portal/blob/develop/portal/app.py#L66] - shouldn't this config also be in the instance directory?

OptimusRhine: Sure, that makes sense. I'll move it.

PR #249, opened 2016-11-23
Fix CORS/same-origin error on SS

ivan-c:


ivan-c:
I tested this out on truenth-dev and it appears to fix the issue of including the `origin` header. 

@achen2401 was using GET instead of POST necessary to get it working? 
Do we still need to allow POST?

pbugni:
It does look to fix the crossdomain problem.

It does not however look to be sending pings anymore?

@achen2401 is it difficult to use POST? 
@ivan-c the failing tests were due to a change in site_persistence - I rebased this branch to fix.

ivan-c:
@achen2401 Sorry, it looks like we were asking for the impossible. Apparently the [Origin header may be included for same-origin requests](https://stackoverflow.com/questions/15512331/chrome-adding-origin-header-to-same-origin-request/16442333#16442333) but chrome does not include it for GET requests.

I think adding the Shared Service `SERVER_NAME` as an allowed origin by default is a better approach. I'll  try implementing it in this PR too.

PR #260, opened 2016-11-29
invoking validation check on form fields in profile

achen2401:
- invoking  call to validate form fields in profile to include name, birthday and email fields - appeared not to be invoked previously
- adjust validation check on email field to update field only after ajax call to check unique email address is done
- remove obsolete code from profile.html
- fixed hard coded instrument id code in assessment session list section to get it dynamically from api

### DIFF
--- a/portal/static/js/main.js
+++ b/portal/static/js/main.js
@@ -323,6 +323,7 @@ var assembleContent = {
             "given": $("input[name=firstname]").val(),
             "family": $("input[name=lastname]").val()
         };
+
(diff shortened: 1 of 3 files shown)


pbugni:
Wonderful catch.  Explains that log message we saw.

I'd be in favor of removing the blocks of commented out code, unless you anticipate needing them again soon?

achen2401:
Paul - yes - I have removed the commented out code - since it no longer applies.

PR #261, opened 2016-11-29
Multi-instrument queuing

ivan-c:
* Allow multiple instruments to be queued for the Assessment Engine at once
  * Instrument code path-parameter now accepts CSVs for instrument code

### DIFF
--- a/portal/models/intervention_strategies.py
+++ b/portal/models/intervention_strategies.py
@@ -182,7 +182,7 @@ def update_card_html_on_completion():
             """.format(user.display_name)
             link_label = 'Begin questionnaire'
             link_url = url_for('assessment_engine_api.present_assessment',
-                               instrument_id='epic26')
+                               instrument_ids='epic26')
         if authored:
             card_html = """
             <h2 class="tnth-subhead">Thank you,</h2>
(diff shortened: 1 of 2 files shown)


portal/models/intervention_strategies.py
```diff
@@ -182,7 +182,7 @@ def update_user_card_html(intervention, user):
             """.format(user.display_name)
             link_label = 'Begin questionnaire'
             link_url = url_for('assessment_engine_api.present_assessment',
-                               instrument_id='epic26')
+                               instrument_ids='epic26')
```
pbugni: If you're expecting an array, shouldn't this be a "list" of instruments?  i.e.
`instrument_ids=['epic26']`

pbugni: Ah - I see.  You are actually parsing the comma delimited string.  This would introduce the third way for us to handle multiple values for a non-scalar.  In /api/coredata/acquire we take multiple 'require' parameters, which I thought was the preferred CGI approach for this.  Elsewhere, we use JSON arrays.  Can you convince me this approach is better?  I'd like our API to be as consistent as possible.

ivan-c: The main `present-assessment` API should now use querystring parameters exclusively.

I moved the old route to a new, deprecated function that can be easily removed once we get everyone on the new API/route.

PR #279, opened 2016-12-02
Feature/provider consent check

pbugni:


pbugni:
This restored the reverted commits from pull request #275 via cherry-picks.

PR #287, opened 2016-12-06
add maintenance notice to landing page

pbugni:


pbugni:
@achen2401 I think we should include the timezone in the message, namely 'Eastern Standard Time'

PR #295, opened 2016-12-07
fix permission for pca-localized section in profile - only visible to users with role of provider

achen2401:
- restrict visibility of pca-localized section (i.e. clinical question inquired of the patient at account creation) in profile to only users having role of provider
- pca-localized section is currently visible in EPROMs, hidden in Truenth


pbugni:
Not sure why we need to make sure the current_user isn't the target person being shown in the profile, but okay :)  The rest of the check looks better, thanks.

PR #302, opened 2016-12-08
Removing unused template files + profile_test ref

OptimusRhine:
Based on team input, removing the following deprecated template files:
access.html
patient_profile.html
portal_old.html
profile_test.html

Also removing the view function for profile_test (this was apparently a proof-of-concept view, and is no longer needed).


pbugni:
Thank you!  Clean up is vital.

PR #316, opened 2016-12-15
User document upload API (POST only)

OptimusRhine:
- Created new model UserDocument, a generic model to handle any user-specific document uploads (patient reports, user avatars, etc)
- Created migration for user_document db table
- Created portal/uploads folder, where actual files are stored using the uuid as the filename (included a .gitignore for folder file contents)
- Created specific patient model API POST endpoint ( /user/<int:user_id>/patient_report ), which allows for PDF file uploads. Endpoint only requires a PDF file, no additional information necessary.
- Created test for above endpoint (and subsequently the UserDocument model as well)


portal/models/user_document.py
```diff
@@ -0,0 +1,63 @@
...
+        return d
+
+    @classmethod
+    def from_post(cls, upload_file, data):
+        user = User.query.get(data['user_id'])
+        if not user:
+            raise ValueError("user not found")
+        if not data['document_type']:
+            raise ValueError('must provide document type')
+        if upload_file.filename == '':
```
pbugni: this will miss some things, like whitespace.  i'd suggest 

`if not (upload_file.filename and upload_file.filename.strip()):`

portal/views/user.py
```diff
@@ -1235,3 +1236,75 @@ def unique_email():
...
+        in: path
+        description: TrueNTH user ID
+        required: true
+        type: integer
+        format: int64
+      - in: body
+        name: body
+        schema:
+          id: file
+          required:
```
ivan-c: I think you need to add `file` as a property here (see [failing test on TravisCI](https://travis-ci.org/uwcirg/true_nth_usa_portal/jobs/184323681#L378)). 

Something like:
```
properties:
  file:
    type: file
    description: File to upload

```

Also, adding it under a schema may not be strictly necessary. Schemas are mostly used to de-duplicate common elements by referring to existing ones (eg `"$ref": "#/definitions/FHIRPatient"`)

tests/test_user_document.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+class TestUserDocument(TestCase):
+    """User Document tests"""
+
+    def test_post_patient_report(self):
+        #tests whether we can successfully post a patient report -type user doc file
+        service_user = self.add_service_user()
+        self.login(user_id=service_user.id)
+        tmpfile = NamedTemporaryFile(suffix='.pdf')
```
ivan-c: I might add using `with` on resources ([see example](https://stackoverflow.com/questions/3924117/how-to-use-tempfile-namedtemporaryfile-in-python/21944302#21944302)) that need to be explicitly closed (like files) is a good habit to be in but not strictly necessary.

OptimusRhine: Both good points. Added additional file contents check, and moved temp file to within a with statement (just for consistency/cleaner code). For the upload file, I'm currently just deleting it after checking the file contents.

portal/config.py
```diff
@@ -52,6 +52,7 @@ class BaseConfig(object):
     GOOGLE_CONSUMER_SECRET = os.environ.get('GOOGLE_CONSUMER_SECRET', '')
 
     DEFAULT_LOCALE = 'en_US'
+    FILE_UPLOAD_DIR = 'uploads'
```
pbugni: brilliant - now if you could add a different value for FILE_UPLOAD_DIR to the TestConfig (below in same file), you're set.

OptimusRhine: Done.

pbugni:
Not to step on your toes, @OptimusRhine, but when I tried this from my demo intervention, it failed.  Seems that the flask docs don't model real world, or at least aren't consistent with a simple and common use for uploading files via http.  Figured it'd be easier to just fix than get a demo intervention up, etc.

The whitespace and extra imports show in my editor with a pyflakes/pep8 plugin.

OptimusRhine:
No problem @pbugni , thanks for your fix. Like you said, probably much easier than trying to set up a new demo intervention so I could test from my end.

PR #324, opened 2016-12-19
portal wrapper timeout fixes

achen2401:
- only start timeout session when user is logged in

- display timeout message as such if user is logged out due to timeout

### DIFF
--- a/portal/templates/landing.html
+++ b/portal/templates/landing.html
@@ -34,7 +34,7 @@
                     <a href="/go" class="btn feature-btn feature-btn-primary">{{ _("Register") }}</a> &nbsp;&nbsp; <a href="/user/sign-in" class="btn feature-btn feature-btn-primary">{{ _("Login") }}</a>
                     {% if timed_out %}
                         <div id="timeOutMessageContainer">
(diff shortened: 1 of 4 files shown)


portal/templates/landing.html
```diff
@@ -17,13 +27,18 @@
...
-                  <div class="headline">
+                    <div class="headline">
                         <div style="margin-bottom: 1em">{{ app_text('landing title') }}</div>
                         {{ app_text('landing sub-title') }}
                     </div>
                     <a href="/go" class="btn feature-btn feature-btn-primary">{{ _("Register") }}</a> &nbsp;&nbsp; <a href="/user/sign-in" class="btn feature-btn feature-btn-primary">{{ _("Login") }}</a>
-
+                    {% if timed_out %}
+                        <div id="timeOutMessageContainer">
+                            You have been logged out due to inactivity. Please log in again to continue.
```
pbugni: We should wrap this in a get text call for i18n, namely:
`{{ _("You have been logged out due to inactivity. Please log in again to continue.") }}`

PR #338, opened 2016-12-21
Improve Event Callback Logging

ivan-c:
This change adds all of the context passed to the event callback `notify()` to the debugging log

### DIFF
--- a/portal/models/auth.py
+++ b/portal/models/auth.py
@@ -151,8 +151,9 @@ class Client(db.Model):
 
         context = {
             "id": res.task_id,
-            "data": kwargs['data'],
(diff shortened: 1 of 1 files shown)


ivan-c:
@pbugni Looks like something broke in the [site persistence](https://github.com/uwcirg/TrueNTH-USA-site-config/). 

I can rebase and re-run tests but it shouldn't be necessary for this minor of a change.

pbugni:
Known issue.  I opted to merge the site persistence change for a different pull request - which has since become complicated for external reasons.  I'll roll back the site persistence change for now, and then trigger a rebuild here.

PR #342, opened 2016-12-22
initial queries fix to skip questions for caregiver user

achen2401:
-  change in the logic so a user who is specified as a caregiver will not be prompted for further questions


pbugni:
@achen2401 please see if you still need the front end changes, now that coredata no longer requires organization info for ROLE.PARTNER (aka caregiver).

achen2401:
Paul - thank you for the fix - I have removed the front-end code to update org for the caregiver user and I tested and worked.

PR #363, opened 2016-12-30
add deceased status/date section to patient profile

achen2401:
- add deceased status/date section to patient profile, viewable and editable only for provider

Paul - to give you a heads up, I found and fixed an issue I encountered while updating deceasedBoolean (see user.py, it seems that the "else" clause had one more indent than it should be).  Let me know if you see any issue with that.


ivan-c:
@achen2401 That un-indentation change looks correct but might be better as a separate commit; merging
@pbugni Are any of those changes in being persisted in `update_deceased()` without calling `db.session.commit()`?

PR #379, opened 2017-01-05
unescape ampersand character in assessment links

achen2401:
- fix escaped ampersand link in assessment links in profile


ivan-c:
Looks good; merging
Thanks!

PR #384, opened 2017-01-06
fix validation bugs for birthday field

achen2401:
- was able to pinpoint that it was the custom boostrap validation check in place (see main.js) that prevented user from submitting form and/or updating field even when all birthdate fields are valid

- replace those custom checks with a function call to validate on change/blur field event of field instead

- fix birthday validation error message displays on challenge form


portal/static/js/main.js
```diff
@@ -123,9 +123,9 @@ var fillContent = {
                 }
             };
             // Display clinics if any value is false (except localized) or if all are answered
-            if ((clinicalValue == "false" && clinicalItem != "pca_localized") || i == 3) {
-                $("#clinics").fadeIn();
-            }
+            //if ((clinicalValue == "false" && clinicalItem != "pca_localized") || i == 3) {
```
pbugni: Did you intend to leave these statements commented out?  If they aren't needed, can we remove them?

achen2401:
Paul - I removed the commented out code - it shouldn't be there as that should be the logic within initial queries itself.

achen2401:
Paul - I  removed commented out code (some from other areas in main.js) as well.

PR #407, opened 2017-01-10
Fuzzy match name only, DOB strict match

OptimusRhine:
Modifying our identity challenge logic - now we still allow for fuzzy matching of firstname and lastname (average of the two fuzzy match ratios must be above configurable threshold, default 85%), but DOB must be a strict match.

### DIFF
--- a/portal/models/user.py
+++ b/portal/models/user.py
@@ -802,7 +802,7 @@ class User(db.Model, UserMixin):
             return '<div>' + '</div><div>'.join(
                 [ui.provider_html for ui in uis]) + '</div>'
 
-    def fuzzy_match(self, first_name, last_name, birthdate=None):
+    def fuzzy_match(self, first_name, last_name, birthdate):
         """Returns probability score [0-100] of it being the same user"""
         # remove case issues as it confuses the match
         scores = []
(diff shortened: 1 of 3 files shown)


portal/models/user.py
```diff
@@ -812,9 +812,11 @@ def fuzzy_match(self, first_name, last_name, birthdate):
 
         # birthdate is trickier - raw delta doesn't make sense.  treat
         # it like a string, assuming only typos for a mismatch
```
pbugni: can you fix the comment to represent current code?

portal/views/portal.py
```diff
@@ -271,9 +271,8 @@ def challenge_identity(user_id=None, next_url=None, merging_accounts=False):
     birthdate = datetime.strptime(form.birthdate.data, '%m-%d-%Y');
 
     score = user.fuzzy_match(first_name=first_name,
-                             last_name=last_name,
-                             birthdate=birthdate)
-    if score > current_app.config.get('IDENTITY_CHALLENGE_THRESHOLD', 85):
+                             last_name=last_name)
+    if (user.birthdate.strftime('%d%m%Y') == birthdate.strftime('%d%m%Y')) and (score > current_app.config.get('IDENTITY_CHALLENGE_THRESHOLD', 85)):
```
pbugni: as per PEP 8, please avoid lines > 79 chars long.  breaking after an open paren works well

tests/test_user.py
```diff
@@ -592,6 +592,10 @@ def test_fuzzy_match(self):
                                  birthdate=user.birthdate)
         self.assertEquals(score, 100)  # should be perfect match
 
+        score = user.fuzzy_match(first_name=user.first_name,
+                                 last_name=user.last_name)
+        self.assertEquals(score, 100)  # should still be perfect match w/o birthday
```
pbugni: Don't agree - we always need a perfect match on birthday as I read it, right?

ivan-c:
@OptimusRhine you might want to rebase this against develop before merging, tests are passing there now :)
Thanks for taking the time to write tests!

OptimusRhine:
@ivan-c Sweet, rebased and it looks like all is well. Thanks Ivan!

@pbugni Moved everything back into the user.fuzzy_match model. Simpler, and now the fuzzy_match test can properly test an incorrect date.

PR #408, opened 2017-01-11
Feature/gil integration

achen2401:
- Integrating GIL html pages to Truenth domain root

- dependence on config variable (i.e. GIL = True) allows pages to be visible in Truenth only

- putting in logic to check whether user is  logged in (e.g. to show login modal )

Paul - let me know if you are okay with my changes especially in portal.py  I'd appreciate any suggestion as to any change(s) that you think are needed.


portal/views/portal.py
```diff
@@ -66,7 +66,29 @@ def landing():
         return next_after_login()
 
     timed_out = request.args.get('timed_out', False)
-    return render_template('landing.html', user=None, no_nav="true", timed_out = timed_out)
+    gil = current_app.config.get('GIL', None)
```
pbugni: The return from a dict.get('key') call where no key is found is None by default.  i.e. this can be simply:
`gil = current_app.config.get('GIL')`

portal/views/portal.py
```diff
@@ -439,16 +461,20 @@ def profile(user_id):
 @portal.route('/legal')
 def legal():
     """ Legal/terms of use page"""
+    gil = current_app.config.get('GIL', None)
     response = requests.get(app_text(LegalATMA.name_key()))
-    return render_template('legal.html', content=response.text)
+    user = current_user()
+    return render_template('legal.html' if not gil else 'gil/legal.html', content=response.text, user = user if user else None)
```
pbugni: current_user() returns None if there isn't one.  therefore you can just call:

`return render_template(..., user=curren_user())`

portal/views/portal.py
```diff
@@ -470,8 +496,9 @@ def contact():
     if request.method == 'GET':
         sendername = user.display_name if user else ''
         email = user.email if user else ''
-        return render_template('contact.html', sendername=sendername,
-                               email=email)
+        gil = current_app.config.get('GIL', None)
```
pbugni: Same as above - no need to pass None.

portal/views/portal.py
```diff
@@ -439,16 +461,20 @@ def profile(user_id):
...
+    return render_template('legal.html' if not gil else 'gil/legal.html', content=response.text, user = user if user else None)
 
 @portal.route('/about')
 def about():
     """main TrueNTH about page"""
     about_tnth = requests.get(app_text(AboutATMA.name_key(subject='TrueNTH')))
     about_mo = requests.get(app_text(AboutATMA.name_key(subject='Movember')))
-    return render_template('about.html', about_tnth=about_tnth.text,
-                           about_mo=about_mo.text)
+    gil = current_app.config.get('GIL', None)
```
pbugni: Same as above, no need to pass None

achen2401:
Paul, thank you very much. I simplified the code as suggested.

PR #414, opened 2017-01-11
Debian Packaging

ivan-c:
These changes provide the groundwork necessary to run Shared Services in Docker by adding tooling and infrastructure to build and distribute Debian packages directly from Github source.

* Uses [dh-virtualenv](dh-virtualenv.readthedocs.io) to build debian package inside docker container (see `docker/Dockerfile.build`)
* Optionally deploys signed package to [Bintray debian repo, if configured](https://bintray.com/ivan-c/true_nth/true_nth_usa_portal)

#### Notes
 * API credentials are now kept as [repo-specific](https://travis-ci.org/uwcirg/true_nth_usa_portal/settings) environmental variables

#### Todo
 * Caching
    * Allow container to reuse host's pip cache
    * Allow current git repo to be passed to container (so it doesn't need to be re-cloned)
 * Investigate TravisCI build options (to improve speed)
    * Build in separate TravisCI "job" or after running tests?
    * Deploy image created by Dockerfile.build to Docker Hub or Bintray
 * Decide whether to build artifacts (.deb or Docker images) for develop or master or both

#### Discussion/Rationale

I chose to build our Docker implementation on top of a Debian package instead of the more typical architecture of adding the application source directly to the image. Although this increases the complexity of building a working Docker image I feel it is worth the cost for the benefit gained in compartmentalizing the different parts of the build process.

If this process is too onerous or slow I'm considering creating an alternative implementation that forgoes the Debian package and builds from the source directly.

Please let me know if you have any comments or suggestions.

WIP


debian/portal.triggers
```diff
@@ -0,0 +1,8 @@
+# Register interest in Python interpreter changes (Python 2 for now); and
+# don't make the Python package dependent on the virtualenv package
+# processing (noawait)
+interest-noawait /usr/bin/python2.6
```
pbugni: Don't bother supporting 2.6 - it's too old and I'm sure we'd run into some syntax errors with the current code base

docker/Dockerfile.build
```diff
@@ -0,0 +1,74 @@
...
+# This Dockerfile will build an image that can be used to create a debian package from latest code in the develop branch. The completed debian package will need to be copied off the image after creation (see below invocation)
+
+# Build the image for creating the package
+# $ docker build --file docker/Dockerfile.build --tag "portal_builddeb:test" .
+
+# Build the image
+# $ docker run "portal_builddeb:test"
+
+# Build the image (override defaults)
+# $ docker run --tty -e REPO_SLUG='ivan-c/true_nth_usa_portal' -e BRANCH='feature/deb-packaging_dockerize-portal_compose' "portal_builddeb:test"
```
pbugni: Do you want REPO_SLUG referring to your user?

requirements.txt
```diff
@@ -97,7 +97,6 @@ WebTest==2.0.24
 Werkzeug==0.11.15
 WTForms==2.1
 xvfbwrapper==0.2.9
--e .
```
pbugni: Ah this again :)

Won't this break our deploy, or at least my practiced way of doing things?

debian/portal.triggers
ivan-c: Good suggestion. This was auto-generated by make-deb and I had overlooked it

docker/Dockerfile.build
ivan-c: I wanted a concrete example of how to override the [default](https://github.com/uwcirg/true_nth_usa_portal/pull/414/files/02331b6bcf3fd199684cda63c5baa6bd3a3daafb#diff-1633c0903adc235424069b03f49ac523R66) of `uwcirg/true_nth_usa_portal` set in the Dockerfile. I'll make it a little less specific

requirements.txt
ivan-c: dh-virtualenv was having issues with modules installed from source and this seemed like a quicker fix. Let me see if I can work around it...

PR #417, opened 2017-01-12
add fix for accessing /home on unauthenticated attempt

achen2401:
- redirecting un-aunthenticated user back to landing page on attempting to access /home

Paul,  this fix is to address this story:
https://www.pivotaltracker.com/story/show/137493505

Let me know if you see any issue with this fix.

### DIFF
--- a/portal/views/portal.py
+++ b/portal/views/portal.py
@@ -62,7 +62,6 @@ def intentional_error():  # pragma: no cover
 def landing():
     """landing page view function - present register / login options"""
     if current_user():
-        print("REALLY")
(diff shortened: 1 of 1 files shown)


portal/views/portal.py
```diff
@@ -62,6 +62,7 @@ def intentional_error():  # pragma: no cover
 def landing():
     """landing page view function - present register / login options"""
     if current_user():
+        print("REALLY")
```
pbugni: Looks like debugging you don't want checked in?

PR #424, opened 2017-01-12
User Doc Download API endpoint

OptimusRhine:
Adding an API endpoint allowing for downloading a specific user document. Returns that user doc's file contents.

Includes tests.

### DIFF
--- a/portal/models/user_document.py
+++ b/portal/models/user_document.py
@@ -72,7 +72,7 @@ class UserDocument(db.Model):
             with open(filepath,"r") as file_in:
                 file_contents = file_in.read()
         except:
-            raise ValueError("count not open file")
+            raise ValueError("could not open file")
 
         return file_contents
 
--- a/portal/models/user_document.py
+++ b/portal/models/user_document.py
@@ -63,3 +63,16 @@ class UserDocument(db.Model):
         return cls(user_id=data['user_id'],document_type=data['document_type'],filename=filename,
                     filetype=filetype,uuid=file_uuid,uploaded_at=datetime.utcnow())
 
+
+    def get_file_contents(self):
+        filepath = os.path.join(current_app.root_path,current_app.config.get("FILE_UPLOAD_DIR"),self.uuid)
+        if not os.path.exists(filepath):
+            raise ValueError("could not find file")
+        try:
+            with open(filepath,"r") as file_in:
+                file_contents = file_in.read()
+        except:
+            raise ValueError("count not open file")
+
+        return file_contents
+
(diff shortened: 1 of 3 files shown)


portal/models/user_document.py
```diff
@@ -63,3 +63,16 @@ def from_post(cls, upload_file, data):
...
+
+    def get_file_contents(self):
+        filepath = os.path.join(current_app.root_path,current_app.config.get("FILE_UPLOAD_DIR"),self.uuid)
+        if not os.path.exists(filepath):
+            raise ValueError("could not find file")
+        try:
+            with open(filepath,"r") as file_in:
+                file_contents = file_in.read()
+        except:
+            raise ValueError("count not open file")
```
pbugni: looks like a typo in the error message?

portal/views/user.py
```diff
@@ -1359,6 +1359,77 @@ def user_documents(user_id):
     return jsonify(user_documents=[d.as_json() for d in
                                        user.documents])
 
+
+@user_api.route('/user/<int:user_id>/user_documents/<int:doc_id>/download')
```
pbugni: is there a need for the /download portion of the request path?  I'm not a RESTful design expert, but it seems like you typically just ask for the resource.  Is something else provided at /user/<user_id>/user_documents/<doc_id> ?

OptimusRhine: @pbugni In the last web app I worked on, a GET endpoint like '/user_documents/' would return the xml/json of the full list of objects, while '/user_documents/<doc_id>' would return the xml/json of just the specific object in question. For a specific action on that file (eg downloading the full filedata), we would use a more specified path.
However, I don't know if that's standard API design or just was the preference of that design/dev team. If it's more in line with the rest of our API approach, I'm fine with modifying this to remove the /download portion.

OptimusRhine:
@pbugni modified per your comments

tests/test_user_document.py
```diff
@@ -54,3 +55,27 @@ def test_post_patient_report(self):
         with open(fpath, 'r') as udoc_file:
             self.assertEqual(udoc_file.read(),test_contents)
         os.remove(fpath)
+
+
+    def test_download_user_document(self):
+        service_user = self.add_service_user()
```
pbugni: This will work - but there's nothing about your API that requires a "service user", and I think it'll just introduce complexity.

Do you mind altering the test to just use the built in test_user?  i.e. self.login() w/ the default user_id of TEST_USER_ID, which can also be used in the format calls below?

OptimusRhine: No prob, switched to the built in test_user.

PR #425, opened 2017-01-13
add custom error pages for errors 404 and 500

achen2401:
add custom error pages for 404 and 500 server errors respectively


portal/views/portal.py
```diff
@@ -31,12 +31,12 @@
 
 
 def page_not_found(e):
-    return render_template('error.html', no_nav="true"), 404
```
pbugni: can we delete error.html?  I believe these two calls were its only use?

achen2401: Paul, yup, I agree.  I removed it.

PR #429, opened 2017-01-14
remove treatment question from profile page

achen2401:
- remove cancer treatment question from Clinical Questions section in profile

this is to address this story: https://www.pivotaltracker.com/story/show/137643955

### DIFF
--- a/portal/templates/profile_macros.html
+++ b/portal/templates/profile_macros.html
@@ -836,9 +836,11 @@
         };
         $(function () {
             $("#patientQ").show();
+            $("#patTx").remove();
+
             $("#patientQ hr").hide();
             {% if person and current_user and person.id != current_user.id %}
-                $("#patientQ input[type='radio'], #treatmentDate, #treatmentMonth, #treatmentYear").each(function() {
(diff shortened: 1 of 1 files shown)


portal/templates/profile_macros.html
```diff
@@ -836,9 +836,11 @@ <h4 class="text-muted">{{ _("Questions inquired of the patient at registration")
         };
         $(function () {
             $("#patientQ").show();
+            $("#patTx").remove();
```
pbugni: Not sure I totally grok the flow, but why are we removing #patTx here?  Shouldn't it just be removed from the profile macros source?

achen2401: Paul, I can't remove it from the source (which is coming from initia_queries_macros.html).  If I do that, the question won't display on the initial queries page (i.e. but we still need the question there).  So to work around that, I basically just remove the question from the rendering DOM on the profile page, therefore effectively making treatment question unavailable only on this page.

pbugni: Makes sense.  Thanks @achen2401

PR #440, opened 2017-01-17
I18n Module

OptimusRhine:
* Created a new i18n module, which contains:
  * the babel.localeselector method (moved from app.py)
  * upsert method to add translatable strings from db to pot file if they don't already exist (and add new ref comments if they do, and said comments don't already exist)
* Added 'translations' option to manage.py, which runs i18n upsert method
* Fixed user locale setter function
* Created test suite for i18n module

### DIFF
--- a/portal/models/i18n.py
+++ b/portal/models/i18n.py
@@ -13,12 +13,12 @@ def get_db_strings():
     elements = defaultdict(set)
     for entry in AppText.query:
         if entry.custom_text:
-            elements["\"" + entry.custom_text + "\""].add("apptext: " + entry.name)
+            elements['"{}"'.format(entry.custom_text)].add("apptext: " + entry.name)
     for entry in Intervention.query:
         if entry.description:
-            elements["\"" + entry.description + "\""].add("interventions: " + entry.name)
+            elements['"{}"'.format(entry.description)].add("interventions: " + entry.name)
         if entry.card_html:
-            elements["\"" + entry.card_html + "\""].add("interventions: " + entry.name)
+            elements['"{}"'.format(entry.card_html)].add("interventions: " + entry.name)
     return elements
 
 
--- a/portal/models/i18n.py
+++ b/portal/models/i18n.py
@@ -1,0 +1,90 @@
+"""Module for i18n methods and functionality"""
+import psycopg2
+import sys
+import os
+from flask import current_app
+from ..extensions import babel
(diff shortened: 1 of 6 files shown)


portal/models/i18n.py
```diff
@@ -0,0 +1,90 @@
...
+        if key not in d:
+            d[key] = set()
+        d[key].add(o)
+
+
+def get_db_strings():
+    cur = connect_to_db()
+    if cur:
+        elements = {}
+        cur.execute("""SELECT custom_text, name FROM apptext""")
```
pbugni: As you've demonstrated, you can run raw SQL statements through the SQLAlchemy engine, but that should only be necessary in the exceptional case.

Catch me online for more help if you'd like.  Query each object for what you need directly, i.e.
```
query = AppText.query.with_entities(AppText.custom_text)
for entry in query:
```

OptimusRhine:
@pbugni @ivan-c yea, querying the object directly makes a lot more sense (for both efficiency and simplicity) than doing a separate db connection. I actually started this as a completely separate script, which in retrospect was probably not the best approach.

In any case, updated upsert methods to query directly instead.

portal/models/i18n.py
```diff
@@ -0,0 +1,65 @@
...
+import sys
+import os
+from flask import current_app
+from ..extensions import babel
+from .user import current_user
+from .app_text import AppText
+from .intervention import Intervention
+
+
+def add_to_dict(d,k,o):
```
pbugni: You can simplify method with the use of defaultdict.  i.e.
```
from collections import defaultdict
d = defaultdict(set)
d[k].add(o)

File: portal/models/i18n.py [hunk 3]
```diff
@@ -0,0 +1,65 @@
...
+    if k:
+        key = "\"" + k + "\""
+        if key not in d:
+            d[key] = set()
+        d[key].add(o)
+
+
+def get_db_strings():
+    elements = {}
+    query = AppText.query.with_entities(AppText.custom_text, AppText.name)
```
pbugni: sorry to flop around on you - but I'd misread that you only wanted a single column in your query.  I only bother using `with_entities` when it would significantly improve readability or speed.  you might just do `AppText.query()` and then reference the named attribute for more readable code, i.e `entry.custom_text` and `entry.name` in the following line.  Make sense?

PR #450, opened 2017-01-19
Add subject_id to audit

OptimusRhine:
* Added subject_id field to audit
  * Migration will populate with affected user from audit.comment (if it can be found), otherwise will copy over user_id
* Modified all existing audit creations and auditable_event() calls to include subject_id
* Added tests for subject_id to audit tests

### DIFF
--- a/portal/views/auth.py
+++ b/portal/views/auth.py
@@ -73,10 +73,10 @@ def deauthorized():
 
 
 def flask_user_login_event(app, user, **extra):
-    auditable_event("local user login", user_id=user.id)
+    auditable_event("local user login", user_id=user.id, subject_id=user.id)
 
 def flask_user_registered_event(app, user, **extra):
-    auditable_event("local user registered", user_id=user.id)
+    auditable_event("local user registered", user_id=user.id, subject_id=user.id)
 
 
 # Register functions to receive signals from flask_user
@@ -144,7 +144,7 @@ def next_after_login():
         invited_user_id = session['invited_verified_user_id']
         assert user.id != invited_user_id
         auditable_event("merging invited user {} into account {}".format(
-            invited_user_id, user.id), user_id=user.id)
+            invited_user_id, user.id), user_id=user.id, subject_id=user.id)
         user.merge_with(invited_user_id)
         invited_user = User.query.get(invited_user_id)
         invited_user.delete_user(acting_user=user)
@@ -267,7 +267,7 @@ def login(provider_name):
                     provider_id=result.user.id).first()
             if ap:
                 auditable_event("login via {0}".format(provider_name),
-                                user_id=ap.user_id)
+                                user_id=ap.user_id, subject_id=result.user.id)
                 user = User.query.filter_by(id=ap.user_id).first()
                 user.image_url=image_url
                 db.session.commit()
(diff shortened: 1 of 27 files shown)


OptimusRhine:
Please hold off on review - per the failing tests, it looks like there are a bunch more audit calls that still need to be updated to use subject_id. I'll update when I'm back in the office on Monday.

OptimusRhine:
@pbugni I think I'm going to add in topic (which I'm tentatively calling 'context') in a separate PR. I have this one to help me re-identify all the audit calls that need to be updated.

Would you mind reviewing this PR? The most important file changes for review are:
portal/audit
portal/models/audit
portal/migrations/versions/4b1e5b7b69eb_.py
tests/test_audit.py

The rest of the file changes should just be adding ', subject_id=<xyz>' to the existing audit creates.

portal/views/auth.py
```diff
@@ -521,14 +523,15 @@ def client():
     db.session.add(client)
     db.session.commit()
     auditable_event("added intervention/client {}".format(
-        client), user_id=user.id)
+        client), user_id=user.id, subject_id=client.user_id)
```
pbugni: i'd suggest `subject_id=user.id` - the subject here is really the "client" - and the client's user.id isn't unique - any user can have any number of clients, and to tie this change to the client.user_id is somewhat misleading.  when you add `context` it'll help.  i don't think it's worth the complexity of making subject a dynamic object type.  (i.e. subject might be a user, or a client, or a ...)

Same for rest of calls in this file.

OptimusRhine: @pbugni I agree that here (line 524), it makes sense just to set `subject_id=user.id` (especially since that's actually the same value as `client.user.id`). But there are some spots where I think it may make more sense to use `client.user_id`? For example, check out line 631. The editing user is not the same one as the user being edited (we even check to see if `current_user()` has permission to edit `client.user_id` above it). Would it not make more sense that the subject of the audit is the `client.user_id`, rather than the `current_user()` doing the editing?

pbugni: Okay - I'm sure you get the point.  Sure, do what you think is best.

OptimusRhine:
Actually, I will just add 'context' to this branch, so we can get both incorporated at once. Please hold off on merging for now.

OptimusRhine:
Audit context now added, including migration (uses regex to extract potential existing contexts) and tests. @pbugni please review whenever you get a chance.

pbugni:
@OptimusRhine sorry, overlapping PRs - several changes in working on the check_role fix included audit calls in the old style.

Would you mind rebasing and resolving these conflicts?

pbugni:
Really nice work, Adam.  The upgrade script is clean - looks to have worked on a larger db (tested on one from stg pulled over weeks ago).

Can we find a way to store the context in its string value?  Of course an int is more efficient, but legibility counts, and we don't even store the mapping in the db - so working with the SQL directly requires an additional lookup.

pbugni:
@OptimusRhine one more change request (sorry to keep 'em coming!)

After running the conversion, several of the "other" context items look to me like they should have their own.  Namely "observation" and "questionnaire_response".

Few examples of observations (I think it's all rows with `set CodeableConcept .* on user` in the comment):

```
 other   | set CodeableConcept  [Coding 131, treatment begun, http://us.truenth.org/clinical-codes] ValueQuantity false,boolean on user 179
 other   | set CodeableConcept  [Coding 141, PCa localized diagnosis, http://us.truenth.org/clinical-codes] ValueQuantity false,boolean on user 195
 other   | set CodeableConcept  [Coding 141, PCa localized diagnosis, http://us.truenth.org/clinical-codes] ValueQuantity true,boolean on user 40
 other   | set CodeableConcept  [Coding 131, treatment begun, http://us.truenth.org/clinical-codes] ValueQuantity true,boolean on user 52
```

and for questionnaire_reponse:
```
 other   | added QuestionnaireResponse 88 for user 12 completed 2016-11-29 00:42:05
 other   | added QuestionnaireResponse 123 for user 413 completed 2016-12-07 18:49:21
 other   | added QuestionnaireResponse 32 for user 194 completed 2016-10-13 12:19:16
```

OptimusRhine:
@pbugni ok, should be rebased with the latest develop, the context field should now be stored as a string rather than an int, and those new migration scenarios should now be taken into account (there's no 
'questionnaire_response' context, but I added that string to the 'assessment' context instead).

pbugni:
Looks great!

PR #481, opened 2017-01-26
Containerize Portal

ivan-c:
Allows creation of docker images and containers from portal source code hosted on Github. Still needs more work as build process is slow (~9 minutes on TravisCI) and does not allow source code to be injected into finished artifacts (which would facilitate interactive development).

#### Notes
* Allows creation of images based on [earlier Debian packaging work](https://github.com/uwcirg/true_nth_usa_portal/pull/414)
  * Packages can now be created locally or use Bintray Debian repo
* Build process moved to separate,[ fail-able job](https://docs.travis-ci.com/user/customizing-the-build#rows-that-are-allowed-to-fail)
  * Will only build if configured with appropriate env vars

##### Operational
- [x] Add documentation, examples
- [x] Create CIRG repos, configure credentials
- [x] Reconfigure to build, but not push artifacts unless configured
- [ ] Add available metadata to artifacts
  - [x] Tags
  - [ ] Labels
##### Code
- [ ] Add volume to persist user-submitted documents/images
- [ ] Log to stdout if unconfigured
- [ ] Implement/configure container [healthchecks](https://docs.docker.com/engine/reference/builder/#/healthcheck)
##### Wishlist
- [ ] Switch to [`flask` CLI](http://flask.pocoo.org/docs/0.12/cli/)
- [ ] Idempotent DB initialization/migration
- [ ] Caching
- [ ] Inject/overlay git checkout into container


pbugni:
Looking over the changed files - I don't think there's anything I can add w/o digging into it more.  It looks great.

Ready for a merge?

One question, do you expect this will slow down our travis build success, as it's building out a docker image every time (I think)?  Probably no easy way around this, just asking.

ivan-c:
I don't mind keeping this open for review while I add some documentation and try to finish some of the easier todos.

>do you expect this will slow down our travis build success, as it's building out a docker image every time (I think)?

I moved the building of Debian packages and docker images into a separate "job" ([example](https://travis-ci.org/ivan-c/true_nth_usa_portal/builds/195588976)) that corresponds with a [tox env](https://github.com/uwcirg/true_nth_usa_portal/pull/481/files#diff-b91f3d5bd63fcd17221b267e851608e8R27) that does nothing and that's allowed to fail without the build waiting for it or affecting its status.

PR #501, opened 2017-01-31
add registration status in profile

achen2401:
This will apply to both Eproms and Truenth

### DIFF
--- a/portal/templates/profile_macros.html
+++ b/portal/templates/profile_macros.html
@@ -237,6 +237,15 @@
     {% if person %}
     <form id="sendEmailForm" action="{{ url_for('portal.invite') }}" method="POST">
         <input type="hidden" name="_userId" id="_userId" value="{{person.id}}" />
+         {% if person %}
+            <div class="registration-status-container">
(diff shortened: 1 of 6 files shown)


portal/templates/profile_macros.html
```diff
@@ -237,6 +237,15 @@
     {% if person %}
     <form id="sendEmailForm" action="{{ url_for('portal.invite') }}" method="POST">
         <input type="hidden" name="_userId" id="_userId" value="{{person.id}}" />
+         {% if person %}
```
pbugni: it appears that you're already in an `{% if person %}` block?

portal/templates/profile_macros.html
```diff
@@ -237,6 +237,15 @@
...
+            <div class="registration-status-container">
+                <span class="text-muted">Registration Status: </span>
+                {% if person and person.has_role("write_only") %}<h4 class="text-warning registration-label">not yet registered</h4>
```
pbugni: same - shouldn't need multiple checks for person being in scope.

PR #518, opened 2017-02-02
account creation fix

achen2401:
not sending in email value instead of the value, "__no_email__" for now to void backend duplicate key error


pbugni:
Not merging.  PR #519 fixes the problem on the back end.

PR #524, opened 2017-02-03
frontend portal wrapper fixes

achen2401:
- retry loading portal wrapper if ajax call failed the first time
- delayed showing wrapper to prevent appearance of unstyled content


ivan-c:
Looks good, going to merge

PR #527, opened 2017-02-03
consent UI restrict viewing of columns for eproms, bug fixes

achen2401:
address this story:https://www.pivotaltracker.com/story/show/138527235
- not showing agreement column based on whether the consent was with parent org

- organization column should show parent organization

- bug fixes while testing - prevent session storage addition in private mode, fix 500 server error in reference.py when trying to cast int of NoneType, remove unnecessary if statement in showWrapper function in main.js

### DIFF
--- a/portal/models/reference.py
+++ b/portal/models/reference.py
@@ -23,7 +23,7 @@ class Reference(object):
     def patient(cls, patient_id):
         """Create a reference object from a known patient id"""
         instance = cls()
-        instance.patient_id = int(patient_id) if patient_id else None
+        instance.patient_id = int(patient_id)
         return instance
 
     @classmethod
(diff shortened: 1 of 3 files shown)


portal/models/reference.py
```diff
@@ -23,7 +23,7 @@ def organization(cls, organization_id):
     def patient(cls, patient_id):
         """Create a reference object from a known patient id"""
         instance = cls()
-        instance.patient_id = int(patient_id) if patient_id else None
+        instance.patient_id = int(patient_id)
```
ivan-c: @pbugni I think one of the migrations Adam wrote for audits failed to account for something in procedures (I'm also seeing an error in running the migrations on my own instance)

PR #533, opened 2017-02-05
Frontend Study Id fixes, timezone error checking fixes

achen2401:
- remove Study Id field from patient's view of profile

- hide Study Id field in Truenth

- display error message when timezone conversion failed (and do not reload the page)


achen2401:
This might create a conflict with pull request #532.  I will resolve it once #532 is reviewed and merged.

PR #537, opened 2017-02-06
Confirm user exists before setting subject_id

OptimusRhine:
In the audit subject_id/context migration, after extracting user id from comments, query users by id to confirm said user exists before using it for subject_id.


pbugni:
Please rebase - https://github.com/uwcirg/true_nth_usa_portal/pull/539 (now merged with develop) should fix the failing tests.

PR #548, opened 2017-02-08
add frontend feature to allow patients list be filtered by org(s)

achen2401:
address this story:https://www.pivotaltracker.com/story/show/139249273

- add widget on patients list page to allow patients list be filtered by orgs based on those of provider's

- refactor code in patients_by_org.html by moving utility js code to standalone admin.js file

- css fixes

### DIFF
--- a/portal/views/patients.py
+++ b/portal/views/patients.py
@@ -51,13 +51,11 @@ def patients_root():
         ).join(UserOrganization).filter(
             and_(UserOrganization.user_id==User.id,
                  UserOrganization.organization_id.in_(org_list)))
-    leaf_organizations = None
-    try:         
-        leaf_organizations = user.leaf_organizations()
-    except ValueError:
-         current_app.logger.debug("In patients list: error retrieving leaf organizations for user %s", str(user.id))
 
-    return render_template('patients_by_org.html', patients_list=patients.all(), user=user, request_org_list=org_list if request_org_list else None, leaf_organizations=leaf_organizations, wide_container="true")
+    leaf_organizations = user.leaf_organizations()
+
+    return render_template('patients_by_org.html', patients_list=patients.all(), user=user, \
+           request_org_list=org_list if request_org_list else None, leaf_organizations=leaf_organizations, wide_container="true")
 
 @patients.route('/profile_create')
 @roles_required(ROLE.PROVIDER)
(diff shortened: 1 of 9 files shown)


portal/views/patients.py
```diff
@@ -46,23 +51,33 @@ def patients_root():
         ).join(UserOrganization).filter(
             and_(UserOrganization.user_id==User.id,
                  UserOrganization.organization_id.in_(org_list)))
+    leaf_organizations = None
+    try:         
+        leaf_organizations = user.leaf_organizations()
+    except ValueError:
+         current_app.logger.debug("In patients list: error retrieving leaf organizations for user %s", str(user.id))
```
pbugni: Is this happening or just debugging?  What error is being raised?

portal/views/patients.py
```diff
@@ -46,23 +51,33 @@ def patients_root():
...
 
-    return render_template('patients_by_org.html', patients_list=patients.all(), wide_container="true")
+    return render_template('patients_by_org.html', patients_list=patients.all(), user=user, request_org_list=org_list if request_org_list else None, leaf_organizations=leaf_organizations, wide_container="true")
```
pbugni: As per [pep 8](https://www.python.org/dev/peps/pep-0008/#maximum-line-length), avoid lines longer than 79 chars, please.

portal/views/patients.py
achen2401: Paul, the ValueError is raised for a provider who doesn't have an org affiliation yet (or "None of the above" is selected" for this provider)

pbugni: Okay, I'd rather fix the source - it should return None w/o orgs in the tree, not raise an error.  Care to make that change, or should I?

achen2401: I can see if I can fix it and add that change to this PR.

portal/views/patients.py
```diff
@@ -47,22 +52,29 @@ def patients_root():
...
 
 @patients.route('/profile_create')
 @roles_required(ROLE.PROVIDER)
 @oauth.require_oauth()
 def profile_create():
     consent_agreements = get_orgs_consent_agreements()
     user = current_user()
-    return render_template("profile_create.html", user = user, consent_agreements=consent_agreements)
+    leaf_organizations = user.leaf_organizations()
+    return render_template("profile_create.html", user = user, consent_agreements=consent_agreements, leaf_organizations=leaf_organizations)
```
pbugni: Please format line < 79 chars per.

portal/views/patients.py
```diff
@@ -47,22 +52,29 @@ def patients_root():
...
 
 
 @patients.route('/sessionReport/<int:user_id>/<instrument_id>/<authored_date>')
 @oauth.require_oauth()
 def sessionReport(user_id, instrument_id, authored_date):
     user = get_user(user_id)
-    return render_template("sessionReport.html",user=user, current_user = current_user(), instrument_id=instrument_id, authored_date=authored_date)
+    return render_template("sessionReport.html",user=user, current_user=current_user(), instrument_id=instrument_id, authored_date=authored_date)
```
pbugni: Please format line < 79 chars per.

PR #551, opened 2017-02-09
front end trim text field fixes

achen2401:
- In GIL, trim text field(s) before submitting responses

- trim text fields on paste event

### DIFF
--- a/portal/templates/flask_user/flask_user_base.html
+++ b/portal/templates/flask_user/flask_user_base.html
@@ -9,7 +9,7 @@
 {% block document_ready %}
 	{{super()}}
 	$(document).on("ready", function() {
-		$("input[type='text']").on("blur", function() {
+		$("input[type='text']").on("blur paste", function() {
 			$(this).val($.trim($(this).val()));
 		});
 	});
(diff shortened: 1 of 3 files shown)


portal/templates/flask_user/flask_user_base.html
```diff
@@ -9,7 +9,7 @@
 {% block document_ready %}
 	{{super()}}
 	$(document).on("ready", function() {
-		$("input[type='text']").on("blur", function() {
+		$("input[type='text']").on("blur paste", function() {
 			$(this).val($.trim($(this).val()));
```
ivan-c: What fields are being trimmed here? If it's user-entered fields it's probably fine but I worry that this is a behavior people will not expect when filling out a form.

ivan-c:
Looks good, did have one question but nothing major

portal/templates/flask_user/flask_user_base.html
achen2401: Ivan, I believe it will just be the username/email and password fields in this case, since flask_user_base.html is only extended by templates within the flask_user folder.  That is why I thought it safe to do that.

PR #563, opened 2017-02-13
Feature/consent index error fixes

achen2401:
Paul and Ivan, I received an "IndexError: list index out of range" error while attempting to go to the GIL home page of user that does not have any organization affiliation.  I traced back the code and fixed the source of the issue.  Let me know if you have issue with this change and/or any feedback on this change is appreciated.

### DIFF
--- a/portal/models/fhir.py
+++ b/portal/models/fhir.py
@@ -617,7 +617,8 @@ class AssessmentStatus(object):
         # If we aren't given a consent, use the first valid consent found
         # for the user.
         if not self._consent:
-            self._consent = self.user.valid_consents[0]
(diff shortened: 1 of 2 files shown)


portal/models/fhir.py
```diff
@@ -617,7 +617,8 @@ def consent_date(self):
         # If we aren't given a consent, use the first valid consent found
         # for the user.
         if not self._consent:
-            self._consent = self.user.valid_consents[0]
+            if self.user.valid_consents and len(list(self.user.valid_consents)) > 0:
```
pbugni: you don't need to check the len, it's considered false if it's an empty list:

```
if self.user.valid_consents:
    self._consent = self.user.valid_constenst[0]
```

PR #564, opened 2017-02-13
Feature/codings by org

OptimusRhine:
https://www.pivotaltracker.com/story/show/139052395
* Added 'coding_options' int to Organization model, to capture whether or not to display specific race/ethnicity/indigenous options for users in that org.
  * Works similarly to user_consent.options (stored as a singular int value, uses masks to determine individual options from that value, has individual getter/setter functions for those options)
    * If the 'use_specific_codings' option isn't set for that org, then the race/ethnicity/indigenous options will be inherited from the parent organization (unless the org has no parent). This continues until an org with specific options is reached, or the root node is hit.
  * Options are available in from_fhir and as_fhir functions
* Added tests for new organization.coding_options
* Added migration for new organization.coding_options
* Added new user property, which grabs the race/ethnicity/indigenous display options from all that user's orgs (recursively inheriting from parents as needed - see above), and returns a dict for each display option, with a True value if any of the user's orgs allow that display option, or a False value otherwise
* Modified profile macros template, so that race/ethnicity/indigenous objects are only displayed if the user's org-based display option function allows for them to be (see above).


pbugni:
Looks good @OptimusRhine - unfortunately, while we were waiting for me to review this, a conflict came along.  Could you please rebase and resolve?

OptimusRhine:
@pbugni Ok, the merge conflicts should be fixed. Looks like there's a failing test, but I think it's unrelated?

pbugni:
@OptimusRhine the comments you added to this PR are great - and need a home outside of the pull request :)

I believe the best home is in docs/source/organization.txt which ultimately ends up at http://truenth-shared-services.readthedocs.io/en/latest/organizations.html

PR #574, opened 2017-02-15
fix orgs UI label for when editable by patient

achen2401:
this should apply to Truenth only as patients can edit orgs in Truenth

### DIFF
--- a/portal/templates/profile_macros.html
+++ b/portal/templates/profile_macros.html
@@ -371,7 +371,8 @@
 <!--this has an ajax call to fill in the content-->
 {% macro profileOrg(person, removeNone, consent_agreements, current_user) -%}
     <div class="form-group profile-section" id="userOrgs">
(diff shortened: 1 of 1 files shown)


portal/templates/profile_macros.html
```diff
@@ -371,7 +371,8 @@ <h4 class="text-success registration-label">registered</h4>
 <!--this has an ajax call to fill in the content-->
 {% macro profileOrg(person, removeNone, consent_agreements, current_user) -%}
     <div class="form-group profile-section" id="userOrgs">
-        <label>{{_("Clinics ")}}<span class="text-muted smaller-text">({{ _("Optional") }})</span></label>
+        {% if 'patient' in config.CONSENT_EDIT_PERMISSIBLE_ROLES %}<label>Main clinic for prostate cancer care</label>{%else%}
```
pbugni: Probably want to lookup translations for this case like the else?

PR #587, opened 2017-02-16
Change provider to staff, add staff_admin

OptimusRhine:
https://www.pivotaltracker.com/story/show/137496767
https://www.pivotaltracker.com/story/show/131323461

* Add `ROLE` to jinja env, so it is callable from templates
* Change `provider` role to `staff` role
* Change all references of (`provider` role or `ROLE.PROVIDER`) to `ROLE.STAFF`
  * Also changed some nearby hardcoded string role references in the templates to their `ROLE` counterparts. Eventually we'll want to change all these I think.
* Changed `provider_html` function+field to `staff_html`
* Migration to change name of role in db from `provider` to `staff`, and to change 'provider_html' column name to 'staff_html'
* Added `staff_admin` role - will be used later (see https://www.pivotaltracker.com/story/show/131323461 )


pbugni:
@OptimusRhine and @achen2401 - thanks - looks good so far.
I see we have a few occurrences of 'provider' in the site config files.  I think we should change these as well, just to avoid confusion (the third example even presents a problem):
```
site_persistence_file.json:        "PATIENTS_BY_PROVIDER_ADDL_FIELDS = ['status']\n", 
site_persistence_file.json:        "PROVIDER_BULK_DATA_ACCESS = True\n", 
site_persistence_file.json:        "CONSENT_EDIT_PERMISSIBLE_ROLES = ['provider', 'admin']\n",
```
I'd suggest a like named feature branch in both site persistence repos.

achen2401:
@pbugni @OptimusRhine 
Was making the same comment on that in pivotal tracker, Paul.  Yeah, a named feature branch sounds like a good idea.

OptimusRhine:
@pbugni So we not only want to change the role values on the list (e.g. ['provider', 'admin'] --> ['staff', 'admin']), but also the variable names themselves (e.g. PROVIDER_BULK_DATA_ACCESS --> STAFF_BULK_DATA_ACCESS)? Just want to confirm, as this second change involves modifying a lot of template code.

pbugni:
@OptimusRhine yes - the change should be consistent and thorough - the term "provider" has been replaced with "staff" - it should be replaced in the comments, templates, etc.

OptimusRhine:
@pbugni got it. I don't normally do CIRG work on Fridays, but I'll aim to make those changes (and fix the inevitable rebase issues) when we get back next week (unless you think there's an urgent need to get this merged this weekend).

OptimusRhine:
-Related site config change PRs-
truenth:
https://github.com/uwcirg/TrueNTH-USA-site-config/pull/14
eproms:
https://github.com/uwcirg/ePROMs-site-config/pull/13

pbugni:
Getting very close, @OptimusRhine 
I see some comment strings needing attention, such as those in portal/views/patients

OptimusRhine:
Should now include
- comments
- tests
- `provider_html` -> `staff_html` (including migration to update the column name on the `user_interventions` table)

PR #608, opened 2017-02-23
GIL content fixes

achen2401:
addressing this story: 
https://www.pivotaltracker.com/story/show/140455791

- fix GIL legal and terms-of-use pages to get content from LR

- add copy right information in GIL header and footer of SS pages

- change display of legal to privacy

- css fixes


achen2401:
This is not quite ready yet.  I need to point GIL's terms of condition page to the correct new LR item.  I will inform when this is ready.

PR #614, opened 2017-02-24
temporary fix for showing profile menu for login-as user

achen2401:
address this story: https://www.pivotaltracker.com/story/show/140106033
- temporary fix to allow user logging as patient to see profile menu item in banner
- made use of session variable to keep track of login-as scenario
- add css fixes for mobile view of patients list

The fix for login-as is needed for today, hence temporary in nature, will have a longer term fix for this

### DIFF
--- a/portal/views/truenth.py
+++ b/portal/views/truenth.py
@@ -177,6 +177,7 @@ def portal_wrapper_html():
     cookie_timeout = request.cookies.get('SS_TIMEOUT')
 
     disable_links = True if 'disable_links' in request.args else False
+    login_as = True if 'login-as' in session else False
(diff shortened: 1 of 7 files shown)


portal/views/truenth.py
```diff
@@ -177,6 +177,7 @@ def expires_in():
     cookie_timeout = request.cookies.get('SS_TIMEOUT')
 
     disable_links = True if 'disable_links' in request.args else False
+    login_as = True if 'login-as' in session else False
```
ivan-c: This could be moved into `render_template()` and simplified as `login='login-as' in session`

PR #617, opened 2017-02-25
Add Ironman demographics

ivan-c:
* Add Ironman demographics instrument to sequence, last, after PREMS

### DIFF
--- a/portal/models/fhir.py
+++ b/portal/models/fhir.py
@@ -702,7 +702,7 @@ class AssessmentStatus(object):
                 (1, "Due"),
                 (30, "Overdue")
             )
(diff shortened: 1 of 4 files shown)


pbugni:
@ivan-c , @mcjustin 
I realize this is merged - but it seems as if this list changes a lot - any reason not to move to a pair of config values, one for localized instruments, one for metastatic?

ivan-c:
@pbugni I think it would be a good idea to formalize the list of instruments per study but I'm not sure the best way to do that without creating configuration specific to eproms/truenth . Talking with @mcjustin it seems difficult to model in a relationship that would work well with both ePROMs/TrueNTH since the instruments are driven by org and interventions respectively.

PR #623, opened 2017-02-27
add banner displaying system version in portal wrapper

achen2401:
addressing this story:  https://www.pivotaltracker.com/story/show/140188597

- add banner in wrapper displaying system version

- banner is absolutely positioned, overlaying the wrapper, so does not change the layout of wrapper

- banner is displayed based on system config variable, "SYSTEM_TYPE".  This is done similar to how CPRO has done:  
Possible system type:
Staging | Development | Demonstration | Production
I welcome suggestion(s) of alternative way of doing this.

### DIFF
--- a/portal/templates/portal_wrapper.html
+++ b/portal/templates/portal_wrapper.html
@@ -59,7 +59,12 @@
                   .on("click", "#log-out", sessMon.logout)
                   .find("#remaining-time").text(sessMon.timeBeforeWarning / 1000);
        {% endif %}
-     });
+
+       (function() {
+        var __env = {% if config.SYSTEM_TYPE %}"{{config.SYSTEM_TYPE}}"{% else %}""{% endif %};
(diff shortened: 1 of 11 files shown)


portal/templates/portal_wrapper.html
```diff
@@ -59,7 +59,12 @@
                   .on("click", "#log-out", sessMon.logout)
                   .find("#remaining-time").text(sessMon.timeBeforeWarning / 1000);
        {% endif %}
-     });
+
+       (function() {
+        var __env = {% if config.SYSTEM_TYPE %}"{{config.SYSTEM_TYPE}}"{% else %}""{% endif %};
+        if ((__env != "" && __env.toLowerCase() != "production") && ($(".watermark").length == 0)) $("<div class='watermark'>TRUE<sup>NTH</sup> - " + __env + " version - Not for study or clinical use</div>").appendTo("body");
```
ivan-c: Could you add a default of "development" to [config.py](https://github.com/uwcirg/true_nth_usa_portal/blob/develop/portal/config.py#L89-L92) as well? We'll need to override it on every production system (just us.truenth.org for now) in application.cfg (I'll take care of that as soon as this is merged)

achen2401: Ivan, should be added.  Let me know if you see anymore issue.

PR #631, opened 2017-02-28
Adding Encounter model

OptimusRhine:
https://www.pivotaltracker.com/story/show/140792077

FHIR documentation:
http://www.hl7.org/FHIR/encounter.html

* Creating Encounter model
  * Fields: id, status, user_id, start_time, end_time, auth_method
  * Includes as_fhir and from_fhir , which use the standard FHIR format ('patient', 'period', etc)
* Created migration for Encounter table
* Created tests for encounter.as_fhir and encounter.from_fhir


tests/test_encounter.py
```diff
@@ -0,0 +1,37 @@
+"""Unit test module for Procedure API and model"""
```
pbugni: Looks like some overlooked cut & paste errors - s/procedure/encounter

OptimusRhine: Ah yea, missed that one. Should be fixed now

PR #637, opened 2017-03-01
fix backend error for terms-and-conditions page

achen2401:
- correct assignment of value returned from tuple to get terms-and-conditions content

### DIFF
--- a/portal/views/portal.py
+++ b/portal/views/portal.py
@@ -542,9 +542,10 @@ def legal():
 def terms_and_conditions():
     """ terms-and-conditions of use page"""
     gil = current_app.config.get('GIL')
-    response_text = VersionedResource.fetch_elements(app_text(Terms_ATMA.name_key()))[0]
+    content, _ = VersionedResource.fetch_elements(
+            app_text(Terms_ATMA.name_key()))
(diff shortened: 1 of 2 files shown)


portal/views/portal.py
```diff
@@ -542,9 +542,10 @@ def legal():
 def terms_and_conditions():
     """ terms-and-conditions of use page"""
     gil = current_app.config.get('GIL')
-    response_text = VersionedResource.fetch_elements(app_text(Terms_ATMA.name_key()))[0]
+    content, _ = VersionedResource.fetch_elements(
+            app_text(Terms_ATMA.name_key()))
```
pbugni: This is still relying on Terms_ATMA for ePROMs - that expands to "Terms and Conditions URL", which I only see in the TrueNTH persistence.  I think we're missing that app_text from [ePROMs persistence](https://github.com/uwcirg/ePROMs-site-config/blob/develop/site_persistence_file.json)

achen2401: Paul, I added the item in this PR: https://github.com/uwcirg/ePROMs-site-config/pull/16   Let me know if you see any issue.

PR #658, opened 2017-03-07
Cleanly recreate test database on schema change

ivan-c:
* Re-analyze test database schema and attempt to drop again if first attempt fails because of schema change between test runs


ivan-c:
Prematurely celebrated, will likely need to traverse and delete all tables/constraints

See:
[DropEverything](https://bitbucket.org/zzzeek/sqlalchemy/wiki/UsageRecipes/DropEverything)

PR #674, opened 2017-03-09
Redirect invited pre-existing users to login page

OptimusRhine:
When users are invited that are already fully registered in the system, email link (access token url) now redirects to the login page, with email field pre-populated (unless the user uses an alternative auth method).

### DIFF
--- a/portal/views/portal.py
+++ b/portal/views/portal.py
@@ -275,7 +275,6 @@ def access_via_token(token):
         auditable_event("invited user entered using token, pending "
                         "registration", user_id=user.id, subject_id=user.id,
                         context='account')
-        existing_user = User.query.filter_by(email=user.email).first()
         session['challenge.user_id'] = user.id
         session['challenge.next_url'] = url_for('user.register', email=user.email)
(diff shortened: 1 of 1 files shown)


portal/views/portal.py
```diff
@@ -268,20 +268,24 @@ def verify_token(valid_seconds):
...
-    session['challenge.merging_accounts'] = True
-    return redirect(url_for('portal.challenge_identity'))
+        # If WRITE_ONLY user does not have PROMOTE_WITHOUT_IDENTITY_CHALLENGE,
+        # challenge the user identity, followed by a redirect to the
+        # registration page. Preserve the invited user id, should we need to
+        # merge associated details after user proves themselves and logs in
+        auditable_event("invited user entered using token, pending "
+                        "registration", user_id=user.id, subject_id=user.id,
+                        context='account')
+        existing_user = User.query.filter_by(email=user.email).first()
```
OptimusRhine: @pbugni Ah yea, whoops, that line was leftover from an earlier attempt at this logic (it's not even used or referenced anywhere at this point). Should be removed now.

PR #699, opened 2017-03-15
Adding mechanism to share environment file between docker and virtualenvwrapper

pbugni:


ivan-c:
Looks great! Thanks for giving this a try, I'll probably need to do some fine-tuning on the packaging but that's easier done in develop and we're not relying on our docker setup being foolproof at the momment

pbugni:
This isn't ready for prime time.  Closing for now, reopen when ready.

PR #700, opened 2017-03-15
Reset password functionality

OptimusRhine:
https://www.pivotaltracker.com/story/show/141435785

* Added API endpoint for resetting passwords
  * Triggering user must have editing permissions over target user
  * Sends password reset email to target user (using target user's email)
  * Creates audit entry for password reset trigger
* Added div+button on profile page to call password reset endpoint
  * div only visible if current user is not the profile user
* Added test for API endpoint method

### DIFF
--- a/portal/views/user.py
+++ b/portal/views/user.py
@@ -1604,3 +1604,63 @@ def upload_user_document(user_id):
         doc.uuid, user_id), user_id=current_user().id, subject_id=user_id,
         context='assessment')
     return jsonify(message="ok")
+
+@user_api.route('/user/<int:user_id>/password_reset', methods=('POST',))
+@oauth.require_oauth()
+def trigger_password_reset_email(user_id):
+    """Trigger a password reset email for the specified user
+
+    Allows admins, staff, etc to manually trigger password reset emails to
+    patients, allowing them to easily change their passwords and login without
+    knowing their current password or email.
+    ---
+    tags:
+      - User
+      - Password
+      - Access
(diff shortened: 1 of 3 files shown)


portal/views/user.py
```diff
@@ -1604,3 +1604,63 @@ def posted_filename(request):
...
+def trigger_password_reset_email(user_id):
+    """Trigger a password reset email for the specified user
+
+    Allows admins, staff, etc to manually trigger password reset emails to
+    patients, allowing them to easily change their passwords and login without
+    knowing their current password or email.
+    ---
+    tags:
+      - User
+      - Password
```
pbugni: These tags build additional categories in the swagger docs.  I think we should avoid lots of tags until the tree is deep and unruly.  For now, lets just use 'User' for this.

portal/views/user.py
```diff
@@ -1604,3 +1604,63 @@ def posted_filename(request):
         doc.uuid, user_id), user_id=current_user().id, subject_id=user_id,
         context='assessment')
     return jsonify(message="ok")
+
+@user_api.route('/user/<int:user_id>/password_reset', methods=('POST',))
+@oauth.require_oauth()
```
pbugni: I believe we want to restrict this to a set of roles.  Probably (and note you want it to be a list not a tuple, as a list get's any one, whereas tuple means all) [ROLE.INTERVENTION_STAFF, ROLE.STAFF, ROLE.ADMIN]

PR #729, opened 2017-03-24
fix flash of unstyled content issue in portal wrapper

achen2401:
This is **not** to be included in the current release to Production

- fix the flash of un-styled content issue in portal wrapper by using combo of css and javascript fix - hiding content (via javascript) until it is rendered (via moving script code after html)
- remove white space in portal wrapper code to aid loading speed

### DIFF
--- a/portal/templates/portal_wrapper.html
+++ b/portal/templates/portal_wrapper.html
@@ -58,7 +58,7 @@
 {% if not enable_links %}
 <style>#tnthNavXs .tnth-nav-btn{ display: none; }</style>
 {% endif %}
-{% if config.SYSTEM_TYPE and config.SYSTEM_TYPE != "Production" %}<div class='watermark'>TRUE<sup>NTH</sup> - {{config.SYSTEM_TYPE}} version - Not for study or clinical use</div>{% endif %}
+{% if config.SYSTEM_TYPE and config.SYSTEM_TYPE != "Production" and config.SYSTEM_TYPE != "production" %}<div class='watermark'>TRUE<sup>NTH</sup> - {{config.SYSTEM_TYPE}} version - Not for study or clinical use</div>{% endif %}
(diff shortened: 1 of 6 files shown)


achen2401:
Reference used for these changes:

http://www.techrepublic.com/blog/web-designer/how-to-prevent-flash-of-unstyled-content-on-your-websites/

portal/templates/portal_wrapper.html
```diff
@@ -92,10 +56,9 @@
 <noscript><p><img src="//piwik.cirg.washington.edu/piwik.php?idsite={{ config.PIWIK_SITEID }}" style="border:0;" alt="" /></p></noscript>
 {% endif %}
 {% if not enable_links %}
-<style>
-    #tnthNavXs .tnth-nav-btn{ display: none; }
-</style>
+<style>#tnthNavXs .tnth-nav-btn{ display: none; }</style>
 {% endif %}
+{% if config.SYSTEM_TYPE and config.SYSTEM_TYPE != "Production" and config.SYSTEM_TYPE != "production" %}<div class='watermark'>TRUE<sup>NTH</sup> - {{config.SYSTEM_TYPE}} version - Not for study or clinical use</div>{% endif %}
```
ivan-c: I might change this a little- `SYSTEM_TYPE` will always be a defined config value since it's in [`config.py`](https://github.com/uwcirg/true_nth_usa_portal/blob/develop/portal/config.py#L105) (no need to check `if config.SYSTEM_TYPE`) and this check might miss someone who configures `SYSTEM_TYPE` to be "PRODUCTION". 

Something a bit more robust (and terse):
```
{% if config["SYSTEM_TYPE"].lower() == "production" %}
```

ivan-c:
@achen2401 Thanks for the article, didn't know this was such a historic problem! :)

PR #739, opened 2017-03-29
Update password reset email

OptimusRhine:
https://www.pivotaltracker.com/story/show/118768435

* Change instances of 'Forgot your password?' to 'Having trouble logging in?'
* For TrueNTH only (based on app_name), forgot password email is changed
  * Slightly different wording (as constructed by Justin in the above PT story)
  * List of used auth methods, based on the user.auth_providers
  * Either link to reset TrueNTH user password, or link to establish a TrueNTH login, based on whether or not user has username (i.e. whether they've used TrueNTH login as an auth method)

Site persistence PR: https://github.com/uwcirg/TrueNTH-USA-site-config/pull/33


OptimusRhine:
@ivan-c This should be ready to merge, whenever you have a chance to review it

portal/templates/flask_user/emails/forgot_password_message.html
```diff
@@ -0,0 +1,36 @@
+{% extends 'flask_user/emails/base_message.html' %}
+
+{% block message %}
+
+{% if app_name == "TrueNTH" %}
+
+<p>It sounds like you're having trouble logging in to TrueNTH.</p>
```
ivan-c: You may want to use [`USER_APP_NAME`](https://github.com/uwcirg/true_nth_usa_portal/blob/develop/portal/config.py#L85) instead of instances of "TrueNTH" to keep consistent

OptimusRhine:
@ivan-c You're thinking the template should just have the TrueNTH version, and should be renamed to something new, and then in the TrueNTH config file we point the USER_FORGOT_PASSWORD_EMAIL_TEMPLATE key at the new TrueNTH template?

That makes a lot of sense, I'll make that change shortly

ivan-c:
@OptimusRhine Yup, sounds right. 

It doesn't particularly matter if the default template is TrueNTH or ePROMs (I think TrueNTH is typically, but whatever is the most generic/adaptable template) but if something needs special language/formatting etc it could have a separate template that is specified by `USER_FORGOT_PASSWORD_EMAIL_TEMPLATE` and included in either the [TrueNTH-USA-site-config](https://github.com/uwcirg/TrueNTH-USA-site-config) or [ePROMs-site-config](https://github.com/uwcirg/ePROMs-site-config)

Hope this is clear as mud now ;)

OptimusRhine:
@ivan-c do you mind reviewing again? It should be more straightforward now, using the config setting to select the desired template (note, I had to also add _message.txt and _subject.txt files for the new custom template).

Sometime after this is merged, we should also merge https://github.com/uwcirg/TrueNTH-USA-site-config/pull/33 , so that we can actually test the new template in TrueNTH.

portal/templates/flask_user/emails/truenth_forgot_password_message.html
```diff
@@ -0,0 +1,23 @@
+{% extends 'flask_user/emails/base_message.html' %}
+
+{% block message %}
+
+<p>It sounds like you're having trouble logging in to {{ app_name }}.</p>
```
ivan-c: Could you use the config `USER_APP_NAME` here instead of `app_name`?

It looks like `app_name` is set by the config `PROJECT`, and defaults to "portal", which makes me think `PROJECT` was a hacky way to get the [name of the project](https://github.com/uwcirg/true_nth_usa_portal/blob/develop/setup.py#L13), [a long time ago...](https://github.com/uwcirg/true_nth_usa_portal/commit/ae377588fc79d94ce54481e68d1e7d07995177fc#diff-b9ff4c6d3e8ce65648f66d4f5cef9fd6L9)

portal/templates/flask_user/emails/forgot_password_message.html
ivan-c: My mistake- looks like [Flask-User maps USER_APP_NAME to app_name in templates](https://github.com/lingthio/Flask-User/blob/1e31e1b4c6a2766d9888a97b30398cdc96de1705/flask_user/settings.py#L21)

portal/templates/flask_user/emails/truenth_forgot_password_message.html
ivan-c: Whoops, my mistake, looks like [Flask-User maps those two together](https://github.com/lingthio/Flask-User/blob/1e31e1b4c6a2766d9888a97b30398cdc96de1705/flask_user/settings.py#L21)

PR #742, opened 2017-03-30
Fixes for email hints popover

achen2401:
- add feedback for unmatched criteria
- prevent user from submitting form without re-typing the password
- add one fix for IE for birthday date match (in profile)

### DIFF
--- a/portal/templates/flask_user/register.html
+++ b/portal/templates/flask_user/register.html
@@ -171,7 +171,7 @@ $(document).ready(function(){
     $("#retype_password").on("blur", function() {
         if (!hasValue($(this).val())) {
             $("#errorretype_password").addClass("has-error").text("Please re-type your password.");
-        };
+        } else $("#errorretype_password").text("");
     });
 
 });
--- a/portal/templates/flask_user/register.html
+++ b/portal/templates/flask_user/register.html
@@ -168,7 +168,7 @@ $(document).ready(function(){
     });
     $("#retype_password").on("blur", function() {
         if (!hasValue($(this).val())) {
-            $("#errorretype_password").addClass("has-error").text("Please re-type your password.");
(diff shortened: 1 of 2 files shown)


portal/templates/flask_user/register.html
```diff
@@ -123,36 +128,52 @@ <h3 class="modal-title" id="myModalLabel">About Using Google or Facebook for Tru
...
-        if (!$(".password-popover").is(":visible")) $(this).popover("show");
+
+    $("#password").popover({"content": "<b>Password must have at least:</b><ul id='pwdHintList'><li class='text-info eight-characters'>Eight characters</li><li class='text-info lowercase-letter'>One lowercase letter</li><li class='text-info uppercase-letter'>One uppercase letter</li><li class='text-info one-number'>One number</li></ul>", "placement": "top", "html": true, "title": "Requirements", "template": '<div class="password-popover popover" role="tooltip"><div class="arrow"></div><h3 class="popover-title"></h3><div class="popover-content"></div></div>', "container": ".password-wrapper", "trigger": "manual"}).click(function(e) {
```
ivan-c: Could you wrap any visible text with translation tags (ie `{%trans%}`)? We'll have to come up for a strategy for translating content in static javascript files at some point, but I think we'll still need to tag strings like this.

portal/templates/flask_user/register.html
```diff
@@ -123,36 +126,52 @@ <h3 class="modal-title" id="myModalLabel">About Using Google or Facebook for Tru
...
+    }).on("shown.bs.popover", function () {
       $(this).trigger("keyup");
     });
 
+    $("html").click(function(e) {
+        $("#password").popover("hide");
+    });
+    $("#retype_password").on("blur", function() {
+        if (!hasValue($(this).val())) {
+            $("#errorretype_password").addClass("has-error").text("Please re-type your password.");
```
ivan-c: I think you forgot to tag a string here

PR #744, opened 2017-03-30
add loading indicator for submit buttons on UI

achen2401:
address this story:  https://www.pivotaltracker.com/story/show/142739533

- display loading indicator after user clicking on submit button on UI.  This applies to initial queries and home pages.

### DIFF
--- a/portal/templates/portal.html
+++ b/portal/templates/portal.html
@@ -1,6 +1,5 @@
 {% extends "layout.html" %}
 {% block main %}
-
 {#
 {% if not user.has_role('write_only') %}
 <div class="portal-item">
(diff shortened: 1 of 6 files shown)


portal/templates/portal.html
```diff
@@ -18,55 +17,63 @@ <h2 class="tnth-subhead">{{ _("Welcome to TrueNTH") }}</h2>
...
-                    </a>
+            {% if display.card_html %}
+                <div class="portal-description">
+                    {{ display.card_html | safe }}
                 </div>
-                {% endif %}
-            </div>
+            {% endif %}
+            
+            {% if display.link_url is none %}
```
ivan-c: I think the following check might be more appropriate, in case `link_url` is an empty string and not just None/null. I may be wrong though...

```
{% if not display.link_url %}
```

PR #747, opened 2017-03-31
GIL header fixes

achen2401:
address a number of stories here:
https://www.pivotaltracker.com/story/show/142862071
https://www.pivotaltracker.com/story/show/142861769
https://www.pivotaltracker.com/story/show/142861737

Fix GIL header location in logged-in state
add missing header info in PC Facts and  About pages


ivan-c:
Looks good, we'll have to go through and tag all of the GIL content for translation at some point

PR #772, opened 2017-04-06
Profile UI changes 

achen2401:
address profile UI and registration related changes in the stories:
https://www.pivotaltracker.com/story/show/142862373
as documented here: https://promowiki.movember.com/display/TU/User+Profile+%28Patient+or+Caregiver%29+Review+6+March+2017#UserProfile(PatientorCaregiver)Review6March2017-UserProfile
https://www.pivotaltracker.com/story/show/143212197
https://www.pivotaltracker.com/story/show/142862625

- numerous changes in profile page and registration related changes
- initial queries bug fix

### DIFF
--- a/portal/templates/initial_queries.html
+++ b/portal/templates/initial_queries.html
@@ -11,18 +11,33 @@
       {{tou(terms)}}
   {% endif %}
   <div id="aboutForm" class="tnth-hide">
+    {% if user and user._email %}
+          <input type="hidden" name="email" id="email" value="{{user._email}}" />
+    {% endif %}
      <form class="form tnth-form to-validate" id="queriesForm" action='initial-queries' method='POST'>
         {% if 'name' in still_needed or 'dob' in still_needed or 'role' in still_needed %}
             <div id="demographicsContainer" class="iq-container" parent-section="true">
-                {% if user and user._email %}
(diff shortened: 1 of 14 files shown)


achen2401:
looking at one issue with the account creation.  Will inform when this is ready for reviewing.

achen2401:
@pbugni @ivan-c 
Paul and/or Ivan -  This PR should be ready.  If either one of you could help me review this PR, I'd really appreciate it.  Quite a few changes here, just want to make sure I didn't mess up anything.

portal/templates/initial_queries.html
```diff
@@ -11,12 +11,12 @@
       {{tou(terms)}}
   {% endif %}
   <div id="aboutForm" class="tnth-hide">
+    {% if user and user._email %}
```
pbugni: @achen2401 anything with an underscore in front of it, by convention, is considered a hidden (similar to private) variable in python.  There are exceptions, but you don't typically want to access these outside of the class.

Is there a reason we're looking at `user._email` here rather than `user.email`?  The hidden value holds the INVITE and NO_EMAIL prefixes, which I don't think should ever be seen or used at the UI level.

achen2401:
@pbugni 
Paul, thank you very much for catching that.  It is older code. I can't believe that I just copied it over without giving it a thought.  I made the change to use user's email instead.

PR #774, opened 2017-04-07
SERVER_NAME env var

ivan-c:
* Allow Flask SERVER_NAME config to be set by env var
  * Required to correctly render portal wrapper (URL, including domain needed)
* Use SERVER_NAME env var to set Apache ServerName in Docker container


ivan-c:
@pbugni Gonna merge this for now; let me know if I've overlooked something in the configuration

PR #781, opened 2017-04-12
bug fixes for staff creation page

achen2401:
fix issues raised in this story:  https://www.pivotaltracker.com/story/show/143217817

- fix staff creation page issue to allow staff admin to create staff account for org(s) below the current user org(s), not just the leave orgs on the org tree
- on staff list page, get list of staff users that is below the current user(org) on the org tree (excluding current user orgs)

@pbugni Paul, if you have a minute, do you mind reviewing the changes?  I am mostly concerned with the query to gather the staff list for the staff admin.
The criteria should be a list of users that belong to org(s) below current user's orgs in the org tree.

### DIFF
--- a/portal/static/css/eproms.css
+++ b/portal/static/css/eproms.css
@@ -1355,8 +1355,16 @@ div.noOrg-container span {
 #createProfileForm .title {
   margin-top: 5%;
 }
-#createProfileForm #phoneGroup {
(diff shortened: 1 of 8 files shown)


achen2401:
closing this for now, as I don't know if this is the best way to proceed.  The best option may be to revert the merged PR for staff admin page.

PR #785, opened 2017-04-12
UI buttons fixes

achen2401:
address stories:
https://www.pivotaltracker.com/story/show/143599725
https://www.pivotaltracker.com/story/show/143602641
https://www.pivotaltracker.com/story/show/143603341
- center complete registration and continue button in initial queries page
- make sure disabled class is removed from email button when its attribute, disabled, is removed
- verbiage fix from "user" to "patient" in patient creation page

### DIFF
--- a/portal/templates/profile_macros.html
+++ b/portal/templates/profile_macros.html
@@ -226,15 +226,24 @@
         <input type="text" class="form-control" id="email" name="email" data-save-container-id="emailGroup" placeholder="{{ _('Email Address') }}" value="{{ person.email if person and person.email and person.email != '__no_email__' }}" data-customemail="true" data-error-field="erroremail" required />
         <div class="help-block with-errors" id="erroremail"></div>
     </div>
-    <script>$(function() {
-        $("#email").on("keyup", function() {
-            if (hasValue($(this).val())) {
-                $("#btnPasswordResetEmail").attr("disabled", false);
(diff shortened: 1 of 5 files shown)


achen2401:
**Note**

These changes are needed for the eproms demo release.  @pbugni @ivan-c 

Paul and/or Ivan, if you have a minute, can you review this?
Thank you very much in advance,
Amy

portal/templates/profile_macros.html
```diff
@@ -226,15 +226,24 @@
...
-                $("#btnProfileSendEmail").attr("disabled", false);
-            } else {
+    <script>
+    function checkEmail(email) {
+        if (hasValue(email)) {
+            $("#btnPasswordResetEmail").attr("disabled", false);
+            $("#btnProfileSendEmail").attr("disabled", false);
+            $('#btnProfileSendEmail').removeClass('disabled');
+        } else {
+            if ($("#email").get(0).validity.valid) {
```
pbugni: not sure I follow the js logic here, but what's the point of the nested if?  don't you just want to disable the control if there's no value for the user's email?

achen2401: The function is linked to the keyup event of the email field.  So, it will disable the button dynamically if the user clear the text of email but re-enable it if there is text in the email again.

pbugni: Ah - thanks.

PR #789, opened 2017-04-12
orgs UI fix in profile 

achen2401:
address this story:
https://www.pivotaltracker.com/story/show/143212197

made the fix in the orgs UI in profile so that - 
In EPROMs where staff is allowed to edit patient's clinic, staff should only be able to select/edit one affiliated clinic for the patient.

### DIFF
--- a/portal/templates/profile_macros.html
+++ b/portal/templates/profile_macros.html
@@ -563,7 +563,9 @@
             if (leafOrgs) OT.filterOrgs(leafOrgs);
             if (!orgsEditable) disableOrgs();
 
-            {% if (person.has_role('patient') or current_user.has_role('patient')) and 'patient' in config.CONSENT_EDIT_PERMISSIBLE_ROLES %}
+            {%- if ((person.has_role('patient') or current_user.has_role(ROLE.PATIENT)) and 'patient' in config.CONSENT_EDIT_PERMISSIBLE_ROLES) or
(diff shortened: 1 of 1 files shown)


portal/templates/profile_macros.html
```diff
@@ -563,7 +563,9 @@ <h4 class="text-success registration-label">registered</h4>
             if (leafOrgs) OT.filterOrgs(leafOrgs);
             if (!orgsEditable) disableOrgs();
 
-            {% if (person.has_role('patient') or current_user.has_role('patient')) and 'patient' in config.CONSENT_EDIT_PERMISSIBLE_ROLES %}
+            {%- if ((person.has_role('patient') or current_user.has_role(ROLE.PATIENT)) and 'patient' in config.CONSENT_EDIT_PERMISSIBLE_ROLES) or
```
pbugni: Can we please use ROLE.name rather than strings :)

portal/templates/profile_macros.html
```diff
@@ -576,7 +578,7 @@ <h4 class="text-success registration-label">registered</h4>
                         $(this).prop("checked", true);
                     };
                 });
-            {% endif %}
+            {%- endif -%}
```
pbugni: Is this a comment?

achen2401: Paul, the '-' is to strip the unnecessary whitespace.  I see the if statement rendered as a blank line so decide to strip the whitespace it created.

PR #790, opened 2017-04-13
fixes issues with staff admin page

achen2401:
address issues raised in this story:
https://www.pivotaltracker.com/story/show/143217817

- add birthday field in the staff creation page\
- allow user to select all orgs (not just leaves) here/below his/her current org for new staff in staff account page
- add checks for users that have admin and/or staff admin role in the same org(s) as the user so they can be excluded from user's staff list
- css fixes

@pbugni Paul, do you mind reviewing this PR when you have a minute?  as this contains some changes in the portal.py.  Thanks a lot in advance.

### DIFF
--- a/portal/static/css/eproms.css
+++ b/portal/static/css/eproms.css
@@ -1363,8 +1363,14 @@ div.noOrg-container span {
 #createProfileForm .title {
   margin-top: 5%;
 }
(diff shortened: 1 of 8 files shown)


achen2401:
ah..per standup, need to add email invite functionality to this.  Will inform when this PR is ready.

PR #792, opened 2017-04-13
fix inconsistent display in UI

achen2401:
fix inconsistent display of text (e.g. date string, no data), button in profile

### DIFF
--- a/portal/templates/profile_macros.html
+++ b/portal/templates/profile_macros.html
@@ -267,7 +267,7 @@
             {%- if person.has_role('write_only') %}<option value="invite">{{ app_text('profileSendEmail option invite') }}</option>{%- endif %}
             <option value="reminder">{{ app_text('profileSendEmail option reminder') }}</option>
         </select>
-        <button id="btnProfileSendEmail" class="btn btn-tnth-primary disabled">{{ _("Send email") }}</button>
+        <button id="btnProfileSendEmail" class="btn btn-tnth-primary disabled" type="button">{{ _("Send email") }}</button>
         <div id="profileEmailMessage" class="text-info"></div>
     </div>
     <script>$(function () {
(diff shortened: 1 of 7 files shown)


portal/templates/profile_macros.html
```diff
@@ -1687,7 +1688,21 @@ <h4 id="communicationsLoc" class="profile-item-title index-item">{{ _("Communica
                 </div>
             </div>
         </div>
-    {%- endif %}
+    {%- elif person and person.has_role(ROLE.STAFF) and current_user and current_user.has_role(ROLE.STAFF_ADMIN)-%}
```
ivan-c: Is the check for `ROLE.STAFF` redundant with the check for `ROLE.STAFF_ADMIN`? ie are all `STAFF_ADMIN` also `STAFF`? Not entirely familiar with story details so disregard if I'm missing something.

portal/templates/profile_macros.html
```diff
@@ -1687,7 +1688,21 @@ <h4 id="communicationsLoc" class="profile-item-title index-item">{{ _("Communica
...
+        <div class="row">
+            <div class="col-md-12 col-xs-12">
+                <div class="profile-item-container">
+                    <h4 id="communicationsLoc" class="profile-item-title index-item">{{ _("Communications") }}</h4>
+                    <div class="flex">
+                        <div id="resetPasswordContainer" class="flex-item">
+                            <p class="text-muted communication-prompt">Send reset password email to staff</p>
```
ivan-c: Aha! Found the untagged string you've been adding to make sure I actually read your code ;)

portal/templates/profile_macros.html
achen2401: Ah, Ivan, I see this is confusing because the person here refers to the person whose profile is being viewed while the current_user is the person viewing the profile.  So what I want to accomplish here is to allow the staff_admin to view the communications section of the person that has the role of staff.

ivan-c: Ah my bad, didn't read that closely enough :)

PR #798, opened 2017-04-17
orgs ui fixes

achen2401:
Fixes to address this story:
https://www.pivotaltracker.com/story/show/143463375

- make consent associated with a fake agreement_url to be not editable (no link url is rendered for fake agreement url)
- allow state from a parent org that has no children to be selectable
- render consent modal only if agreement asset is present


achen2401:
This will fix the display issues of orgs in UI.  minor fixes here so will merge.

PR #801, opened 2017-04-18
Allow UI tests to fail without failing build

ivan-c:
* Let build pass if UI/selenium testing fails (`TOXENV=ui`)


pbugni:
Clean change.  I think we're getting too many false positives lately, and the UI tests aren't flushed out much at this time.  Merging for now - we should in the future see if sauce becomes more stable and revert this.

PR #802, opened 2017-04-18
Restore running tests on test db on systems with SQL environment variables

pbugni:
Fix accidental consequence from commit 3ea3586271006dd690bf98585d458faf0025a8b0


pbugni:
I believe we finally have a combo that will allow travis tests to use the built in test database, and for developers with environment vars set to not purge their dbs on every test run!

PR #809, opened 2017-04-20
initdb is now safe to call on existing dbs, and stamps the migration

pbugni:
Fixes as per https://www.pivotaltracker.com/story/show/134463297

### DIFF
--- a/manage.py
+++ b/manage.py
@@ -31,6 +31,18 @@ manager.add_command('db', MigrateCommand)
 manager.add_command('runserver', ConfigServer(host='0.0.0.0', threaded=True))
 
 
+def _run_alembic_command(args):
+    """Helper to manage working directory and run given alembic commands"""
+    # Alembic looks for the alembic.ini file in CWD
+    # hop over there and then return to CWD
+    cwd = os.getcwd()
+    os.chdir(MIGRATIONS_DIR)
+
+    alembic.config.main(argv=args)
+
+    os.chdir(cwd)  # restore cwd
+
+
 def stamp_db():
     """Run the alembic command to stamp the db with the current head"""
     # if the alembic_version table exists, this db has been stamped,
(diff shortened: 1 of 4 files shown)


pbugni:
@ivan-c You'll note I removed the dropdb call - we shouldn't be using this path for the extremely rare need of reseting the db - just use the RDBMS directly.  Turns out you can safely call db.create_all() w/o repercussions on an existing db.

ivan-c:
@pbugni Does `python manage.py db migrate` still needs to be called after initializing a completely new database?

pbugni:
@ivan-c I can see the desire to have it run the mirgrations as well:
https://github.com/uwcirg/true_nth_usa_portal/blob/develop/portal/models/fhir.py#L863-L878

I'll modify it to do so - then you can just always call initdb

pbugni:
@ivan-c tells me he can just call migrate right after initdb - works for me, reopening PR.

manage.py
```diff
@@ -24,20 +25,52 @@
...
 
 
+def _run_alembic_command(args):
+    """Helper to manage working directory and run given alembic commands"""
+    # Alembic looks for the alembic.ini file in CWD
+    # hop over there and then return to CWD
+    cwd = os.getcwd()
+    os.chdir(MIGRATIONS_DIR)
+
+    alembic.config.main(argv=args)
```
ivan-c: I think you can pass the path to alembic.ini as an arg (or kwarg `file_`) to the[ `Config` class](https://github.com/briandailey/alembic/blob/master/alembic/config.py#L23) instead of `cd`ing... but I don't think there's a functional difference

pbugni: Yes, I saw that, but it's a different path from being able to call with argv options - you'd need to do the equivalent of what's done in the main() therein, and then maintain it.  This hack looked like the path of least resistance.

ivan-c:
Should be ready to merge unless there's more to discuss

PR #818, opened 2017-04-22
Log to stdout by default

ivan-c:
* No longer require `LOG_FOLDER` config variable
* Write all logs to stdout if unconfigured

**NB: Will required `LOG_FOLDER` to be explicitly set in `application.cfg` for existing deployments**

### DIFF
--- a/portal/app.py
+++ b/portal/app.py
@@ -162,7 +162,7 @@ def configure_logging(app):  # pragma: no cover
         sql_log.addHandler(sql_file_handler)
 
     if app.testing or not app.config.get('LOG_FOLDER', None):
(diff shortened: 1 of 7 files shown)


portal/app.py
```diff
@@ -161,7 +161,7 @@ def configure_logging(app):  # pragma: no cover
         sql_log.setLevel(logging.INFO)
         sql_log.addHandler(sql_file_handler)
 
-    if app.testing:
+    if app.testing or not app.config.get('LOG_FOLDER', None):
         # Skip test mode. Just check standard output.
```
pbugni: Please update the comment - it's no longer just a test mode switch.

PR #821, opened 2017-04-23
Run migrations on Datica post build

ivan-c:
* Run database migrations on Datica after build


ivan-c:
Accidentally picked to merge into develop instead of feature/initdb-improvements

PR #822, opened 2017-04-23
Run database migrations after build on Datica

ivan-c:
* Run database migrations on Datica after build


ivan-c:
Don't merge until feature/initdb-improvements is merged- interested to see what happens to a branch of a branch when the original is merged before it's branch

PR #823, opened 2017-04-24
Feature/preselectedclinic access code fixes

achen2401:
found this bug while testing - it seems access code only worked for child org
- made the fix so any access code that is associated with a parent org without any child org is deemed valid.
- consolidate code in main.js by creating common methods (e.g. get consent modal) for accessing 
  orgs and consent information - for easier manageability
- remove unused and unnecessary code

### DIFF
--- a/portal/static/js/main.js
+++ b/portal/static/js/main.js
@@ -608,6 +608,7 @@ var fillContent = {
                         ckOrg.prop("checked", true);
                         $("#stateSelector").find("option[value='" + state + "']").prop("selected", true).val(state);
                     });
+                    $(".noOrg-container").show();
                 } else {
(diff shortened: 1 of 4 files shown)


ivan-c:
@achen2401 Does this need to be in this week's release ([release/v17.4.22](https://github.com/uwcirg/true_nth_usa_portal/tree/release/v17.4.22))?

achen2401:
@ivan-c 
Ivan, I believe so.  This is a fix related to access code only worked for child org. It should work for parent org without any child org.

ivan-c:
I recreated your branch and PR (#824) so it's based off of the release branch ("release/v17.4.22") instead of develop. that way we can merge your branch into the release branch without any of the changes in develop

PR #826, opened 2017-04-24
On identity verification, don't throw error if year < 1900

OptimusRhine:
https://www.pivotaltracker.com/story/show/143695513

Bug fix, to no longer throw an error when a user inputs a birthyear < 1900 during identity verification. Instead, verification just fails as normal (since no users can have birthyears < 1900 in the first place).

### DIFF
--- a/portal/models/user.py
+++ b/portal/models/user.py
@@ -1171,7 +1171,8 @@ class User(db.Model, UserMixin):
         # birthdate is trickier - raw delta doesn't make sense.  treat
         # it like a string, mismatch always results in a 0 score
         dob = self.birthdate or datetime.utcnow()
-        if (dob.strftime('%d%m%Y') != birthdate.strftime('%d%m%Y')):
+        if (birthdate.date().year < 1900 or
+            dob.strftime('%d%m%Y') != birthdate.strftime('%d%m%Y')):
             return 0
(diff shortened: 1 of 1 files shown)


portal/models/user.py
```diff
@@ -1171,7 +1171,8 @@ def fuzzy_match(self, first_name, last_name, birthdate):
         # birthdate is trickier - raw delta doesn't make sense.  treat
         # it like a string, mismatch always results in a 0 score
         dob = self.birthdate or datetime.utcnow()
-        if (dob.strftime('%d%m%Y') != birthdate.strftime('%d%m%Y')):
+        if (birthdate.year < 1900 or
+            dob.strftime('%d%m%Y') != birthdate.strftime('%d%m%Y')):
```
pbugni: I don't believe we should ever have a member with a birthday < 1900.  The problem is the string given to strftime will raise an exception if the year within is < 1900.

OptimusRhine: Right, with this change, it no longer raises an exception. Now the conditional is false before it even reaches the `.strftime()` call, so it just moves on to the `return 0` line (same as if the dates didn't match in the first place)

pbugni: ah, my bad, I hadn't expanded the code to see the context of 'birthdate'.

PR #829, opened 2017-04-24
filter patients list by consented users

achen2401:
made the fix to filter patients list by consented users.


achen2401:
@pbugni 
Paul, if you have a second, do you mind reviewing this PR?  The fix is to allow patients list to be filtered by consented users - users that have given consent to org(s) belonging to the staff user.   The fix is prompted by 401 errors we are now seeing in demo.us.   The PR failed the tests - related to unmatched staff_html() property of user object - it could be caused by the changed query in this PR.  Please advice.  Thank you very much in advance.

pbugni:
I can't reproduce the test failure locally - calling it a travis testing anomaly for now.  We'll address if it shows up after the merge.

PR #833, opened 2017-04-25
GIL new lived experience page

achen2401:
address this story:
https://www.pivotaltracker.com/story/show/143547263

create new lived experience for GIL

not for this upcoming release but can be merged into develop upon reviewed


portal/views/portal.py
```diff
@@ -144,6 +144,18 @@ def exercise_and_diet():
 def lived_experience():
     return render_template('gil/lived-experience.html', user=current_user())
 
+@portal.route('/david-andrew-story')
```
pbugni: Can I push back on having all these links in the root context?  Perhaps that was requested?  I'd rather see something like a nested /stories/... directory for easier growth down the road.

PR #835, opened 2017-04-26
patients list child orgs fixes

achen2401:
made these fixes after rounds of testing against eproms and then truenth persistence settings

- refactor code to correctly use existing org list for querying for consented users as the org list also contains child org(s) of the current orgs

@ivan-c Ivan, FYI, this should be included in the release branch tomorrow


achen2401:
changes here are flawed.  Need to think about it.

PR #838, opened 2017-04-26
Replace Flask-Script with builtin Flask.cli

ivan-c:
* Replace Flask-Script commands in manage.py with built-in [Flask CLI](http://flask.pocoo.org/docs/0.12/cli/) (based on [click module](/pallets/click))
* Allows running commands independent of current path (eg `flask initdb`), after setting `FLASK_APP`
  * `FLASK_APP` can be set in venv activate script ie `FLASK_APP='/path/to/manage.py'`

#### Todo:
- [ ] Refactor `purge_user()` acting-user prompt to use click

### DIFF
--- a/manage.py
+++ b/manage.py
@@ -81,7 +81,11 @@ def sync():
 @click.option('--keep_unmentioned', '-k', default=False, help='Keep orgs and interventions not mentioned in persistence file')
 def seed(include_interventions=False, keep_unmentioned=False):
     """Seed database with required data"""
-    add_static_concepts()
+
+    # Request context necessary for generating data from own HTTP APIs
+    with app.test_request_context():
+        add_static_concepts()
+
     add_static_interventions()
     add_static_organization()
     add_static_relationships()
--- a/manage.py
+++ b/manage.py
@@ -27,6 +27,11 @@ app = create_app()
 MIGRATIONS_DIR = os.path.join(app.root_path, 'migrations')
 migrate = Migrate(app, db, directory=MIGRATIONS_DIR)
 
+@app.cli.command()
+def runserver():
+    # Todo: figure out how to override default host in `flask run`
+    # http://click.pocoo.org/5/commands/#overriding-defaults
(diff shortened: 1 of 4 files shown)


manage.py
```diff
@@ -93,7 +101,7 @@ def seed(include_interventions=False, keep_unmentioned=False):
     SitePersistence().import_(include_interventions, keep_unmentioned)
 
 
-@manager.command
+@app.cli.command()
 def export_site():
```
ivan-c: @pbugni I noticed `export_site()` writes to the same file we read from (set by `PERSISTENCE_FILE`). Would it make more sense to write to a file specified by an argument instead?

pbugni: It might - but we never use this anymore, as too many hand edits have since taken place on the site persistence file, and the generated order of JSON attributes isn't consistent.  That was by design initially, and will only function if `PERSISTENCE_FILE` is pointing to a filesystem (not URL) location.

manage.py
```diff
@@ -74,10 +76,16 @@ def sync():
...
+@click.option('--include_interventions', '-i', default=False, help='Include (overwrite) intervention data')
+@click.option('--keep_unmentioned', '-k', default=False, help='Keep orgs and interventions not mentioned in persistence file')
 def seed(include_interventions=False, keep_unmentioned=False):
     """Seed database with required data"""
-    add_static_concepts()
+
+    # Request context necessary for generating data from own HTTP APIs
+    with app.test_request_context():
+        add_static_concepts()
+
```
pbugni: Why aren't all of the add_static_*() calls within the same `with` block?

ivan-c: As far as I can tell only `add_static_concepts()` makes HTTP requests against the portal (when adding [valuesets](https://github.com/uwcirg/true_nth_usa_portal/tree/develop/portal/views/fhir.py#L10))

ivan-c:
Needs more work - doesn't fit with deploy.sh yet

PR #840, opened 2017-04-26
Organization locale extension

OptimusRhine:
https://www.pivotaltracker.com/story/show/141258425
(nb: this does not contain _all_ the requested functionality (e.g. the actual UI filtering of user locale options based on org locales); it's just the baseline functionality for storing locales on the org level)

* created OrganizationLocale model, which connects Organizations to locale Codings
  * includes association in Organization model
  * includes db migration for new organization_locales table
* broke Extension class out from the User module into its own module
  * made slightly more generic to work for non-User objects
* added locale extension to Organization
  * created LocaleExtension class (extends generic Extension class)
  * created extension mapping function
  * added 'extension' element to Organization fhir (both inbound and outbound)
    * added test code for extension element to test_organization module


portal/models/user.py
```diff
@@ -44,60 +44,23 @@
...
 
 
-class Extension:
-    """Abstract base class for common user extension FHIR objects"""
-    __metaclass__ = ABCMeta
-
+class UserIndigenousStatusExtension(CCExtension):
+    # Used in place of us-core-race and us-core-ethnicity for
+    # Australian configurations.
     def __init__(self, user, extension):
```
pbugni: Was it an error to leave the `Extension.__init__()` method behind?  I believe you want that on the base class.

portal/models/user.py
```diff
@@ -44,60 +44,23 @@
...
     extension_url = TRUENTH_EXTENSTION_NHHD_291036
 
     @property
     def children(self):
         return self.user.indigenous
 
 
-class UserEthnicityExtension(Extension):
+class UserEthnicityExtension(CCExtension):
+    def __init__(self, user, extension):
```
pbugni: Once you put the init method back in the base, you shouldn't need it elsewhere, of course.

OptimusRhine: I wasn't sure the best way to approach the `init` method. When Extensions are used in the user model, the method takes a `user` parameter, but for orgs we want it to take an `organization` instead. So I just put it in the extended classes, instead of the base class.

I'm definitely open to suggestions on that one, if there's a good way to work it into the base class.

pbugni:
Adam and I discussed and decided to leave as is - other options just introduce more inheritance complexity given the existence of the middle level CCExtension.

PR #846, opened 2017-04-27
Only display locale options valid for user's orgs

OptimusRhine:
https://www.pivotaltracker.com/story/show/141258425

* added a new property to user model, which returns a set of that user's orgs' locales
* modified the locale section in profile_macros, to only display each locale if it (a) is in that user's org locale list, or (b) if it is the system default

### DIFF
--- a/portal/models/user.py
+++ b/portal/models/user.py
@@ -483,6 +483,17 @@ class User(db.Model, UserMixin):
         return options
 
 
+    @property
+    def locale_display_options(self):
+        """Collates all the locale options from the user's orgs
+        to establish which should be visible to the user"""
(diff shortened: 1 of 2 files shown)


OptimusRhine:
Based on @mcjustin 's comments, it seems we just need en_AU available for eproms and en_US available for truenth for now.

Since those are those systems' defaults, and this code change will always allow/display the system default (regardless of what's available in the org locales), we should be set for now.

Just note that when locales are added to the system in the future, we'll need to also add those locales to the relevant orgs in each system.

In other words, @pbugni , this should also be ready to merge, once you've had a chance to review.

PR #855, opened 2017-05-03
Demographics org reference can use external identifier

OptimusRhine:
https://www.pivotaltracker.com/story/show/143697363

* modified Reference model, so that Organizations can be referenced from within other fhir api calls (e.g. `User.update_from_fhir`) via an external identifier
  * format: instead of `{'reference': 'api/organization/<id>'}` (or variants), expects` {'reference': 'api/organization/<identifier_system>/<identifier_value>'}` (or variants)
  * as with normal org reference, throws an error if no org is found; however, _also_ throws an error if more than one org is found (reference can only be used with unique identifiers)
* added test for org external identifier reference to test_demographics
* added GET api endpoint at `/api/organization/<identifier_system>/<identifier_value>`
* added test for GET endpoint

### DIFF
--- a/portal/models/reference.py
+++ b/portal/models/reference.py
@@ -1,7 +1,10 @@
 """Reference module - encapsulate FHIR Reference type"""
 import re
+from sqlalchemy import and_
 
 from ..database import db
+from .identifier import Identifier
+from .organization import Organization, OrganizationIdentifier
 
 class MissingReference(Exception):
     """Raised when FHIR references cannot be found"""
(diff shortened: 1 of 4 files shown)


portal/models/reference.py
```diff
@@ -72,6 +75,29 @@ def parse(cls, reference_dict):
...
+                result = Organization.query.join(
+                      OrganizationIdentifier).join(Identifier).filter(and_(
+                          Organization.id==OrganizationIdentifier.organization_id,
+                          OrganizationIdentifier.identifier_id==Identifier.id,
+                          Identifier.system==id_system,
+                          Identifier._value==id_value))
+            if not result.count():
+                raise MissingReference("Reference not found: {}".format(
+                    reference_text))
+            elif result.count() > 1:
```
pbugni: I think for consistency, as well as the potential need to catch it in view code somewhere, we should define our own, say MultipleReference exception.

OptimusRhine:
@pbugni gotcha. now includes the api get endpoint (+ tests), and uses a new MultipleReference exception for multiple results returned during the reference search.

PR #856, opened 2017-05-03
Significant refactor to support QuestionnaireBanks

pbugni:
Extension of issue https://www.pivotaltracker.com/story/show/143842221
"instruments" are for the most part now referred to as their proper
FHIR Resource Type, Questionnaire.  A QuestionnaireBank is associated
with an organization, and defines the set of Questionnaires and their
associated meta-data, such as days_till_overdue.


pbugni:
Be sure to merge https://github.com/uwcirg/ePROMs-site-config/pull/22 shortly after this PR is merged, for proper funtion.

pbugni:
Ivan - you'll note this isn't yet directly tying QuestionnaireResponses to Questionnaires, but it soon should.  Perhaps you'd like to add that to this PR, or probably better, we merge this in and do that subsequently.

Comments welcome.

portal/models/questionnaire.py
```diff
@@ -0,0 +1,47 @@
+"""Questionnaire module"""
+from sqlalchemy.dialects.postgresql import JSONB
+
+from ..database import db
+
+
+class Questionnaire(db.Model):
+    __tablename__ = 'questionnaires'
+    id = db.Column(db.Integer, primary_key=True)
+    title = db.Column(db.Text, nullable=False, unique=True)
```
ivan-c: Why not make the title (AKA `instrument_id`) the primary key? 

No strong feelings either way, will probably continue to use `instrument_id` in the developer-facing parts of the API

ivan-c: Also, would prefer if this attribute had some name to indicate that it's a unique identifier or code. I feel like title implies human-readable

portal/models/questionnaire_bank.py
```diff
@@ -0,0 +1,176 @@
...
+
+class QuestionnaireBankQuestionnaire(db.Model):
+    """link table for n:n association between Questionnaires and Banks"""
+    __tablename__ = 'questionnaire_bank_questionnaires'
+    id = db.Column(db.Integer(), primary_key=True)
+    questionnaire_bank_id = db.Column(db.Integer(), db.ForeignKey(
+        'questionnaire_banks.id', ondelete='CASCADE'), nullable=False)
+    questionnaire_id = db.Column(db.Integer(), db.ForeignKey(
+        'questionnaires.id', ondelete='CASCADE'), nullable=False)
+    days_till_due = db.Column(db.Integer, nullable=False)
```
ivan-c: Shouldn't these allow null (`days_till_due`, `days_till_overdue`)?

portal/models/questionnaire_bank.py
```diff
@@ -0,0 +1,176 @@
...
+            db.session.add(self)
+            if commit_immediately:
+                db.session.commit()
+        else:
+            self.id = existing.id
+        self = db.session.merge(self)
+        return self
+
+    @staticmethod
+    def q_for_user(user):
```
ivan-c: Would this method make more sense in the User model?

portal/models/questionnaire.py
pbugni: I chose the name 'title' to match the [FHIR spec](https://www.hl7.org/fhir/questionnaire.html).  The db field is unique.

Within the spec, you'll note we have the standard identifiers that can be added.  I see 'name' is also there.  (i'm taking this discussion to skype...)

portal/models/questionnaire_bank.py
pbugni: I did it this way, as user is a HUGE model already.

portal/models/questionnaire_bank.py
pbugni: I don't believe so, at least not at this time.  Adding the constraint on the db means the comparison code doesn't have to handle nulls.  But perhaps I'm wrong...

PR #859, opened 2017-05-04
Feature/ui datetime format fixes

achen2401:
address this story: https://www.pivotaltracker.com/story/show/144718977
- display date/time in the format as specified in the story
- create common date validation and formatting functions to be used for date field(s) from various sources of UI
- remove/clean up date conversion code using locale and timezone as it yields inconsistent results across browsers (will use library of choice later to do this perhaps)

review and comment(s) on these changes are greatly appreciated.

### DIFF
--- a/portal/static/js/main.js
+++ b/portal/static/js/main.js
@@ -106,7 +106,12 @@ function convertToLocalTime(dateString) {
         var offset = d.getTimezoneOffset() / 60;
         var hours = d.getHours();
         newDate.setHours(hours - offset);
-        convertedDate = newDate.toLocaleString();
+        var options = {
+            year: 'numeric', day: 'numeric', month: 'short',
+            hour: 'numeric', minute: 'numeric', second: 'numeric',
+            hour12: false
+        };
+        convertedDate = newDate.toLocaleString(options);
     };
     return convertedDate;
 };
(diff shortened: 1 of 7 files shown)


portal/static/js/main.js
```diff
@@ -2896,8 +2920,130 @@ var tnthDates = {
...
+        var convertedDate = ""+date.getFullYear() + (date.getMonth()+1) + date.getDate();
+        var givenDate = "" + y + m + d;
+        return ( givenDate == convertedDate);
+    },
+    /*
+     * NB
+     * For dateString in ISO-8601 format date as returned from server
+     * e.g. '2011-06-29T16:52:48'*/
+
+    "formatDateString": function(dateString, format) {
```
ivan-c: It might be better to split this up into smaller functions that act on Date objects (instead of strings), separating the parsing and formatting steps and calling those smaller functions as necessary.

Passing around Date objects would only require you to parse the date string once, then they could be formatted in any number of ways. As far as I understand the Intl library acts on Date objects so using them as an intermediate format would be a step in the right direction.

portal/static/js/main.js
```diff
@@ -2896,8 +2920,130 @@ var tnthDates = {
         }
         return toReturn
     },
-    "isDate": function(d) {
-        return Object.prototype.toString.call(d) === "[object Date]" && !isNaN( d.getTime() );
+    "isValidDefaultDateFormat": function(date, errorField) {
```
achen2401: Ivan, as discussed, I am hesitant to use Date.parse, as from what I read and experienced, it is returning inconsistent results across browsers.

portal/static/js/main.js
achen2401: Ivan, I agree, I plan to do this when we incorporate locale and timezone functionality later on.  Right now, I like to keep the code in this one function instead of separate smaller functions as it is easier to maintain this way, at least that is my opinion.

PR #860, opened 2017-05-04
Making sure localeselector override function is used

OptimusRhine:
https://www.pivotaltracker.com/story/show/144709337

Looks like there's some strange issue where, when the app is running via apache, babel doesn't notice our override of `get_locale` in the i18n.py module. Adding this import line to the portal.py view seems to resolve the problem.


ivan-c:
@OptimusRhine Did you end up trying gunicorn and seeing if it had the same issue?

I'm guessing the way the Flask app is [created under Apache](https://github.com/uwcirg/true_nth_usa_portal/blob/develop/portal.wsgi#L15) is different than on the [flask dev server](https://github.com/uwcirg/true_nth_usa_portal/blob/develop/manage.py#L24).

PR #862, opened 2017-05-04
Correct error text for demographics PUT when missing json header.

pbugni:
Found when debugging a bad request call with Martin - now clearly returning the details when a user forgets the `Content-Type: application/json` header.


ivan-c:
Seems hard to generalize to every API endpoint (maybe as a decorator?), but makes sense on some of the more public ones.

PR #864, opened 2017-05-05
Feature/org tree patients for staff

pbugni:
Fix for https://www.pivotaltracker.com/story/show/142538519 Now restricting exports (via /api/patient/assessment) to the list of patients the staff user has permission to view.

### DIFF
--- a/portal/date_tools.py
+++ b/portal/date_tools.py
@@ -1,0 +1,69 @@
+"""Module for additional datetime tools/utilities"""
+from datetime import date, datetime
+from dateutil import parser
+from flask import abort, current_app
+import pytz
+
+
+def as_fhir(obj):
+    """For builtin types needing FHIR formatting help
+
+    Returns obj as JSON FHIR formatted string
+
+    """
+    if hasattr(obj, 'as_fhir'):
+        return obj.as_fhir()
+    if isinstance(obj, datetime):
+        # Make SURE we only communicate unaware or UTC timezones
+        tz = getattr(obj, 'tzinfo', None)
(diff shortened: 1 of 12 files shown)


portal/date_tools.py
```diff
@@ -0,0 +1,69 @@
...
+        """
+        # As we use datetime.strftime for display, and it can't handle dates
+        # older than 1900, treat all such dates as an error
+        epoch = datetime.strptime('1900-01-01', '%Y-%m-%d')
+        try:
+            dt = parser.parse(data)
+        except ValueError:
+            msg = "Unable to parse {}: {}".format(error_subject, data)
+            current_app.logger.warn(msg)
+            abort(400, msg)
```
ivan-c: Should HTTP logic be outside the view? 
I understand there aren't any other use cases outside of HTTP but I thought it was more typical to raise an exception, catch it in the view, and returning the relevant HTTP code (depending on the exception)?

portal/models/organization.py
```diff
@@ -530,6 +535,40 @@ def at_or_below_ids(self, organization_id, other_organizations):
...
+        for org in staff_user.organizations:
+            staff_user_orgs.update(self.here_and_below_id(org.id))
+
+        if not staff_user_orgs:
+            return []
+
+        patient_role_id = Role.query.filter_by(name=ROLE.PATIENT).one().id
+        now = datetime.utcnow()
+        query = db.session.query(User.id).join(
+            UserRoles).join(UserConsent).join(UserOrganization).filter(
```
ivan-c: Ah, so that's how you do all those joins ;)

PR #868, opened 2017-05-05
Feature/eproms patient home page new ui changes

achen2401:
Add following changes to eproms home page:

add due date fix
add date format fix
add styling fixes

### DIFF
--- a/portal/models/intervention_strategies.py
+++ b/portal/models/intervention_strategies.py
@@ -15,7 +15,7 @@ the parameters given to the closures.
 """
 from flask import current_app, url_for
 import json
-from datetime import timedelta
+from datetime import datetime, timedelta
 from sqlalchemy import and_, UniqueConstraint
 from sqlalchemy.dialects.postgresql import JSONB
 from sqlalchemy.orm.exc import NoResultFound, MultipleResultsFound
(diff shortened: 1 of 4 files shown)


portal/models/intervention_strategies.py
```diff
@@ -228,21 +227,20 @@ def get_top_level_org_name():
...
-            # )
-            # for days_allowed, org in thresholds:
-            #     if org == parent_org:
-            #         return assessment_status.consent_date + timedelta(days=days_allowed)
             instrument_due_date = None
-            for instrument, details in assessment_status.instrument_status.items():
-                if not instrument_due_date:
-                    instrument_due_date = details.get('by_date')
+            # prefer due_date for first instrument needing full assessment
+            needing = assessment_status.instruments_needing_full_assessment()
```
ivan-c: You might be able to simplify this using `or` in the following way (`or` returns the first value that evaluates to `True`). Not required, just a little easier to read :)
```
instruments = (
    assessment_status.instruments_needing_full_assessment() or
    assessment_status.instruments_in_process()
)
if instruments:
    return assessment_status.instrument_status[instruments[0]].get('by_date')
return datetime.utcnow()
```

PR #883, opened 2017-05-09
Alt phone field

OptimusRhine:
* added new alt_phone field to User model (+ migration)
* added functionality to handle alt_phone field to the User API (via Telecom, following the FHIR structures)
  * New phone field has `use: 'home'` attribute
  * Old phone field now has `use: 'mobile'` attribute (note that if a phone field is sent without a `use` value, we will assume it's this old phone field, and so we should be backwards compatible)
* added 'Phone (Other)' field to profile + patients list
* extended test_demographics to test phone functionality (including removing a phone value)


OptimusRhine:
Please don't merge yet - needs input from Amy

OptimusRhine:
@pbugni ready for review

ivan-c:
@OptimusRhine Looks great but the frontend doesn't seem to indicate that the phone number will be stored as a "home" phone number. Is that intentional?

OptimusRhine:
@ivan-c Somewhat intentional. The only non-'mobile' options for use (per the FHIR doc here) were 'home', 'work', 'temp', and 'old'.

Eproms specifically asked for that 'Phone (Other)' label, and I wasn't sure which use to map it to on the back-end, so just went with the most innocuous one ('temp' and 'old' seem like they have specific implications that we may not want to include here).

But I'm open to suggestions, if you think one of the other use values is more applicable.

ivan-c:
@OptimusRhine No, that seems like a good decision ([Typically used with urgent cases, *or if no other contacts are available.*](https://www.hl7.org/fhir/valueset-contact-point-use.html#expansion))

If we need to further expand this and support any number of phones/contact points [the rank attribute ](https://www.hl7.org/fhir/datatypes-definitions.html#ContactPoint.rank)might be a good way to differentiate primary/optional phones.

portal/models/telecom.py
```diff
@@ -12,37 +12,49 @@ class Telecom(object):
...
-        return "Telecom: {0.phone} {0.email}".format(self)
+        return "Telecom: {0.phone} {0.alt_phone} {0.email}".format(self)
 
     @classmethod
     def from_fhir(cls, data):
         telecom = cls()
         for item in data:
             attr = item.get('system')
             value = item.get('value')
+            use = item.get('use')
```
pbugni: Seems like you're really only supporting two use types - I'd suggest you raise an exception if they give you an unexpected one, and just set it to the default if they don't pass one.  I believe it'll simplify this code block, and catch errors that might get through (i.e. what would happen if they sent in use of 'other'?

OptimusRhine: It should already be setting to default if they don't pass one. But I agree, we should throw a warning if they send an invalid one, I'll add that in

OptimusRhine:
After talking with Paul, we're approaching this from another direction, by adding in a ContactPoint class. A fair amount of work's now needed here, so I'm going to go ahead and close this PR.

PR #889, opened 2017-05-09
Invert logic - user now must take action to exclude_interventions on

pbugni:
import.


pbugni:
It'd be good to include this - makes updates (like we have for Truenth interventions) happen by default.  It's a far more rare event that we need to capture 3rd party intervention changes.

PR #896, opened 2017-05-10
Feature/flask cli cleanup

ivan-c:
* Allow acting user (actor) to be passed as argument
* Move warning prompt out of model to click decorator

### DIFF
--- a/manage.py
+++ b/manage.py
@@ -132,7 +132,7 @@ def export_site():
     help='Email of user to act as.',
     prompt= \
         "\n\nWARNING!!!\n\n"
-        " This will permanently the target user and all their related data.\n"
+        " This will permanently delete the target user and all their related data.\n"
         " If you want to contiue,"
         " enter a valid user email as the acting party for our records"
 )
--- a/manage.py
+++ b/manage.py
@@ -127,10 +127,20 @@ def export_site():
 
 
 @click.option('--email', '-e', help='Email of user to purge.')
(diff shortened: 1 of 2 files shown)


manage.py
```diff
@@ -126,11 +126,21 @@ def export_site():
...
 
 
-@click.option('--username', '-u', help='Username of user to purge.')
+@click.option('--email', '-e', help='Email of user to purge.')
+@click.option(
+    '--actor', '-a',
+    help='Email of user to act as.',
+    prompt= \
+        "\n\nWARNING!!!\n\n"
+        " This will permanently the target user and all their related data.\n"
```
pbugni: typo - 'This will permanently...'

portal/models/user.py
```diff
@@ -113,7 +113,7 @@ def children(self):
         raise NotImplementedError
 
 
-def permanently_delete_user(username, user_id=None, acting_user=None):
+def permanently_delete_user(username, user_id=None, acting_user=None, actor=None):
```
pbugni: Why do you need both `acting_user` and `actor`?

ivan-c: `acting_user` is a DB object, `actor` is a string

looks like we need to support both- string for Flask CLI, DB object for [creating accounts](https://github.com/uwcirg/true_nth_usa_portal/blob/develop/portal/views/user.py#L240)

PR #901, opened 2017-05-11
fix display of study Id in patients list and profile

achen2401:
Found while testing in eproms:

Fix display of study Id in patients list and profile
remove inline styling from patients list

not for this release. can be merged into develop upon review.


achen2401:
merging this minor styling fix

PR #904, opened 2017-05-12
remove duplicate code for footer in UI, extend common footer template

achen2401:
- remove duplicate code for rendering footer and extend common footer template for flexibility for different layout and different logo image

### DIFF
--- a/portal/templates/flask_user/_macros.html
+++ b/portal/templates/flask_user/_macros.html
@@ -89,21 +89,26 @@
 
 </div>
 {%- endmacro %}
-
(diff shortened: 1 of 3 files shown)


pbugni:
@achen2401  I realize this branch is already merged, but I'm wondering why we have all this footer code in those three files and not in layout.html?

achen2401:
Paul, they should all be in one single file, flask/_macro.html now, instead of all these three different places before.  I removed the duplicate code.  Layout.html call the footer macro from flask/_macro.html.

PR #905, opened 2017-05-12
Adding fhir-compliant default locale to orgs

OptimusRhine:
https://www.pivotaltracker.com/story/show/145362585

* new default_locale property in org model
  * stored as id relation to coding table
  * getter and setter functions that just return/take the code (e.g. 'en_US'), but verify that code is in the codings table
  * appears in API as 'language' (FHIR compliant: https://www.hl7.org/fhir/organization.html )
* migration for new column
* additional test for new field


portal/migrations/versions/426c876ed746_.py
```diff
@@ -0,0 +1,28 @@
...
+revision = '426c876ed746'
+down_revision = 'b92dc277c384'
+
+from alembic import op
+import sqlalchemy as sa
+
+
+def upgrade():
+    ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('organizations', sa.Column('default_locale_id', sa.Integer(), nullable=True))
```
pbugni: Curious about the name "default_locale_id".  Not sure it has to change, but why not just call it "locale_id"?  If the value gets changed, the column name will be misleading, no?

OptimusRhine: @pbugni we already have the 'locales' relation (via organization_locales table). The 'default_locale_id' is that organization's default (or "highest priority") locale.

Basically, per Justin's story, when a user is first created and assigned to an org, we want to use that org's locale(s) to set the user's initial locale. But because there may be cases where an org has multiple locales assigned to it, the default_locale_id should dictate which specific locale should be assigned to that user.

OptimusRhine: (just realized I forgot the story link in the original description - added now)

PR #909, opened 2017-05-15
add locale based on org UI changes, refactor org tool for common use,css fixes

achen2401:
address this story:
https://www.pivotaltracker.com/story/show/145316847

- save locale for new user/staff upon creation (locale is based on the assigned clinic's (or its parent's) default locale.  If no default locale is set, will search the extension of the organization re
- refactor organization client-size code (i.e. OrgTool) for common use 
- remove un-used code from profile_macro.html
- css fixes


achen2401:
close this now, as API will be changed so won't need to traverse org tree to get at default locale

PR #910, opened 2017-05-15
Raise 500 status code instead of 400 on file save error

OptimusRhine:
* when POSTing a user doc via the API, if the file cannot be saved, the response code should be 500 (to indicate a server-side error)


pbugni:
@OptimusRhine I'd like to push this out to demo, do you mind branching from release/v17.5.10 and creating another PR?

pbugni:
I believe Ivan effectively duplicated this PR as a hotfix.  Please reopen if this assumption is false.

PR #911, opened 2017-05-15
raise 500 status code instead of 400 on file save error

OptimusRhine:
* creating an additional PR for https://github.com/uwcirg/true_nth_usa_portal/pull/910 , branched from release/v17.5.10 (so it can be pushed to demo-eproms)


OptimusRhine:
Don't think this'll work for hotfixing v17.5.10, so just closing for now (https://github.com/uwcirg/true_nth_usa_portal/pull/910 will put the actual change into develop)

PR #916, opened 2017-05-16
Feature/user locale ui changes

achen2401:
address this story: https://www.pivotaltracker.com/story/show/145316847
- save default locale for new user and new staff
- remove unused code and consolidate shared code
- css fixes


pbugni:
As discussed with @achen2401 , we'd like to move most of the OrgTool needs back into the server side API.  This will require we figure out what's missing, and refactor the front end to do so.

Merging this issue for now, that will be a pending refactor for future consideration.

PR #917, opened 2017-05-16
Org name in email template + other minor fixes

OptimusRhine:
* modified eproms forgot_password and password_changed email templates per eproms request
  * in the password_changed email template, if the user has just one org, instead of `contact your representative at {{ app_name }}`  we now display `contact your representative at {{ org name }}`
  * removed `Sincerely, {{ app_name}}` from end of emails
  * fixed email titles to not repeat app_name


portal/templates/flask_user/emails/eproms_password_changed_message.html
```diff
@@ -3,8 +3,20 @@
...
 <p>Your password has been changed.</p>
 {% if user_manager.enable_forgot_password %}
+
+{% if user.organizations.count() == 1 %}
+<p>If you did not initiate this password change, <a href="{{ url_for('user.forgot_password', _external=True) }}">please click here to reset it</a>, or contact your representative at {{ user.organizations.first().name }}.</p>
+{% else %}
 <p>If you did not initiate this password change, <a href="{{ url_for('user.forgot_password', _external=True) }}">please click here to reset it</a>, or contact your representative at {{ app_name }}.</p>
+{% endif %}
+
+{% else %}
```
pbugni: what's with this dangling `else`?  is there reason to avoid indentation?

OptimusRhine: @pbugni the else is a part of another if statement (this new statement is embedded within that statement)

as for indentation, I just wasn't sure what the standard was for these html templates

OptimusRhine: @pbugni added indentations to html template to improve readability (couldn't add them to the txt template, as those would have shown up in the actual email)

PR #919, opened 2017-05-17
Feature/portal footer html api changes

achen2401:
address this story:  https://www.pivotaltracker.com/story/show/145451931
- add an API call for portal footer HTML
- add in-line style for footer to eliminate the need for external stylesheet
- add images for icons in the footer, instead of drawing icons using Awesome Font font face, which requires including font-face css that may affect styling for pages that include this footer


pbugni:
Looks clean, @achen2401 .  Did you test on both authenticated and not?

achen2401:
@pbugni 
Paul, I neglected to test when not authenticated.  I added a check for that.  Let me know if you see any issue with that fix.

PR #922, opened 2017-05-17
Smartling file upload + translation file cleanup

OptimusRhine:
https://www.pivotaltracker.com/story/show/144245779

* manage/flask command to upload .pot language file to smartling
  * authenticates in smartling (generates an auth key)
  * generates a .pot file
  * adds translatable strings from the db (interventions, app_text) to the .pot file
  * uploads .pot files to smartling using auth key
* new config settings for smartling
  * SMARTLING_PROJECT_ID included in defaults (config.py)
  * SMARTLING_USER_ID and SMARTLING_USER_SECRET need to be manually set in application.cfg (for security reasons)
* remove 'en_ES' translation files (not a currently supported language)
* add 'en_MX' and 'fr_CA' translation files
* updated .pot file


OptimusRhine:
@ivan-c here's what I've got for the smartling upload. it's functioning great, but let me know if you see any areas where my approach can be changed or improved

portal/config.py
```diff
@@ -105,6 +105,9 @@ class BaseConfig(object):
 
     SYSTEM_TYPE = 'development'
 
+    SMARTLING_LANGUAGES = ['en_AU', 'es_MX', 'fr_CA']
+    SMARTLING_PROJECT_ID = '13f8e2dcf'
```
pbugni: Shouldn't the project be unique for TrueNTH versus ePROMs?  If so, please move this value to the respective site_persistence files.  Probably true for the languages as well?

OptimusRhine: I'm not sure we want to have separate projects for TrueNTH vs ePROMS. The files are uploaded to single projects in Smartling, so if we had multiple projects, it means we would be uploading message.po files for the same language to different projects, and doubling the amount of translation work needed (even if it's just copying/pasting all the translations across projects)

As for the languages, this is just the 'master list' of languages we have PO files for. The actual languages available to users are already restricted per site, based on the site orgs and their locales. So I would think the SMARTLING_LANGUAGES shouldn't be site-specific (especially if, as mentioned above, we only have one project for both sites anyway)

OptimusRhine:
Per discussion with @ivan-c , modified to just upload the .pot file (seems like Smartling will generate the separate .po files for download). Makes the upload process a fair bit simpler (and also means we don't need a language list in the config, for now)

portal/models/i18n.py
```diff
@@ -52,6 +55,49 @@ def upsert_to_template_file():
...
+    pot_fpath = os.path.join(translation_fpath, 'messages.pot')
+    config_fpath = os.path.join(current_app.root_path, "../instance/babel.cfg")
+    # create new .pot file from code
+    check_call(['pybabel', 'extract', '-F', config_fpath, '-o', pot_fpath,
+                current_app.root_path])
+    # update .pot file with db values
+    upsert_to_template_file()
+    # upload .pot file to smartling
+    headers = {'Authorization': 'Bearer {}'.format(auth)}
+    files = {'file': ('messages.pot', open(pot_fpath, 'rb'))}
```
ivan-c: Could you `open()` in a `with` or add a `pot_fpath.close()` line?

portal/models/i18n.py
```diff
@@ -52,6 +55,49 @@ def upsert_to_template_file():
...
+    data = {
+        'fileUri': 'portal/translations/messages.pot',
+        'fileType': 'gettext'
+    }
+    resp = post('https://api.smartling.com/files-api/v2/projects/{}/' \
+            'file'.format(current_app.config.get("SMARTLING_PROJECT_ID")),
+            data=data, files=files, headers=headers)
+    resp.raise_for_status()
+
```
ivan-c: Could you add a logging line indicating success? Very helpful in a CI environment where it's difficult to debug remotely

OptimusRhine:
Logging and proper file handling (aka using `with open(...`) added

ivan-c:
Talked with @mcjustin - sounds like we'll need a way to segregate ePROMs and TrueNTH content at some point to prevent translators from translating content for a language that's not available on their project. Most of the content will probably be in features exclusive to each project- GIL content for TrueNTH and assessment content for ePROMs.
It may make sense to have a Smartling "project" for each to make separating their content easier. Smartling has a feature called "translational memory" that allows reuse of old translations, across projects. Want to get an idea of how easy it is to reuse translations in that way before picking an approach.

PR #923, opened 2017-05-17
Feature/multiple qbs

pbugni:
Fair amount of rework in this PR - questionnaire banks now have a classification and any number of them can be associated with a single organization.

Reworked internals for AssessmentStatus, to make it easier to understand and work with.

UI may need a little attention, as we're still trying to generate multiple cards from a single intervention_strategy.


pbugni:
Of note, we need https://github.com/uwcirg/ePROMs-site-config/pull/29 which breaks out the indefinite QuestionnaireBank from the baselines.

PR #925, opened 2017-05-18
fix 401 error when accessing intervention json items for GIL menu

achen2401:
Fix the issue of encountering 401 error when making an ajax call from portal footer when testing in intervention environment.  
@pbugni 
Paul, let me know if you see any issue with this fix.  thanks a lot in advance.


achen2401:
this appears to be more complicated, closing this now.  will investigate further

PR #929, opened 2017-05-18
Smartling file download

OptimusRhine:
* manage/flask command to download .po language files from smartling
  * authenticates in smartling (generates an auth key)
  * either downloads one translation file (if specific language code is specified with `-l xx_XX`), or all available translation files, for the stored messages.pot file
  * saves po files to relevant directories (e.g. `portal/translations/fr_CA/LC_MESSAGES/messages.po`)
  * compiles new mo files based on downloaded po files
* cleanup debug messages, etc from smartling upload/auth methods


OptimusRhine:
@ivan-c Here's the other side of the smartling coin, the po file download. Seems pretty solid with my testing, but let me know if you notice anywhere it could be improved.

And as a side note, luckily the smartling api has a GET endpoint that automatically gets a zip file that contains po files for _all_ translations from the specified pot file - so we still don't need to store a list of languages in our config anywhere (we'll basically just track them in smartling)

ivan-c:
Sorry for the delay, wanted to make sure we didn't need anything else to get it to work on TravisCI :)

PR #932, opened 2017-05-19
Coredata API includes still_needed and required endpoints.

pbugni:
Refactored coredata to separate 'required' data from what is
'still_needed'.  Previously these were poorly coupled in the hasdata
method for each data point class.  More logical grouping, enables front
end developers to determine which controls should be available to a user
(list of required options) versus the ones for which we have yet to
gather any data (still_needed).

Backend changes needed for issue https://www.pivotaltracker.com/story/show/144417951

### DIFF
--- a/portal/models/coredata.py
+++ b/portal/models/coredata.py
@@ -175,7 +175,7 @@ class RaceData(CoredataPoint):
         return False
 
     def hasdata(self, user):
-        return len(user.races) > 0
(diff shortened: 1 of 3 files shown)


pbugni:
@mcjustin Take a gander at the resulting list of classes inheriting from `CoredataPoint`.  The `required` method in particular for each.

PR #933, opened 2017-05-19
Smartling CI upload

ivan-c:
* Upload strings for translation in new `translations` TravisCI job/tox env
  * Only upload from `develop` branch


ivan-c:
> if we need to do separate projects for each site, we can override the SMARTLING_PROJECT_ID via site persistence file

Yeah, that's seeming more likely, thanks!

PR #934, opened 2017-05-19
make consent date field in patients table sortable

achen2401:
address this story: https://www.pivotaltracker.com/story/show/142739321
allowing consent date field in the patients list table sortable
sorting of consent date is based on raw consent (e.g. 11-11-1980T00:00:00) but will display using the format, d M yyyy, as requested

### DIFF
--- a/portal/static/css/eproms.css
+++ b/portal/static/css/eproms.css
@@ -861,7 +861,7 @@ div.right-panel {
 }
 #patientList .email-field,
 #staffList .email-field {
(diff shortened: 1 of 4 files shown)


pbugni:
@achen2401 is this still in process, given yesterday's discussion?

achen2401:
@pbugni 
Paul, I am not making the assessment status column sortable just yet, will do another PR for that.  this should be k to merge upon review.

PR #936, opened 2017-05-22
Translating strings within strategy html

OptimusRhine:
* broke out translatable strings from `intervention_strategy` html blocks, so that they could be wrapped with gettext ( `_( )` )
  * note that translatable strings still contain some variable format brackets (e.g. "Hello {},"), so those will need to be translated around

### DIFF
--- a/portal/models/intervention_strategies.py
+++ b/portal/models/intervention_strategies.py
@@ -14,6 +14,7 @@ the parameters given to the closures.
 
 """
 from flask import current_app, url_for
+from flask_babel import gettext as _
 import json
 from datetime import timedelta
 from sqlalchemy import and_, UniqueConstraint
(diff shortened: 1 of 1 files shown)


portal/models/intervention_strategies.py
```diff
@@ -230,89 +231,91 @@ def intro_html(assessment_status):
                 due_date = assessment_status.next_available_due_date(
                     classification='baseline')
                 assert due_date
+                h2 = _("Hi {}").format(user.display_name)
```
pbugni: I don't like the `h2` and `h4` variable names - too likely to age poorly, like when the format changes.

how about `welcome_text` and `reminder` or whatever suits you?

OptimusRhine: Sure, I'll rename to something more descriptive

PR #938, opened 2017-05-22
enforce UI check for existence of organization for staff/staff admin user

achen2401:
address server errors for staff having no organization affiliation

- add UI check for existence of organization affiliation if the user has the role of STAFF or STAFF_ADMIN

### DIFF
--- a/portal/static/js/main.js
+++ b/portal/static/js/main.js
@@ -1683,9 +1683,21 @@ OrgTool.prototype.handleEvent = function() {
             } else {
                 var isChecked = $("#userOrgs input[name='organization']:checked").length > 0;
                 if (!isChecked) {
+                    //do not attempt to update if all orgs are unchecked for staff/staff admin
+                    var isStaff = false;
+                     $("#rolesGroup input[name='user_type']").each(function() {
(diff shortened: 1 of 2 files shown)


portal/static/js/main.js
```diff
@@ -1683,10 +1683,22 @@ OrgTool.prototype.handleEvent = function() {
             } else {
                 var isChecked = $("#userOrgs input[name='organization']:checked").length > 0;
                 if (!isChecked) {
+                    //do not attempt to update if all orgs are unchecked for staff/staff admin
+                    var isStaff = false;
+                     $("#rolesGroup input[name='user_type']").each(function() {
+                        if (!isStaff && ($(this).is(":checked") && ($(this).val() == "staff" || $(this).val() == "staff_admin"))) {
+                            $("#userOrgs .help-block").addClass("error-message").text("Cannot ununcheck.  A staff member must be associated with an organization");
```
pbugni: Does the .text() call need i18n?

achen2401: Paul, yes, and other similar places in main.js (and other applicable js files).   When we agree on which third-party library to use (perhaps one of the ones suggested by Ivan), then we can do this en mass.  The translation tags we are using currently only work for JS code embedded in a flask template (or within python file).

PR #941, opened 2017-05-23
Race/Ethnicity sections UI display fixes

achen2401:
address this story:  https://www.pivotaltracker.com/story/show/144417951

hiding Race & Ethnicity sections based on return of API from this PR #932.

**NOTE** this PR is dependent on changes in PR #932 

@pbugni 
Paul, thank you very much for the API.  It works for me.  If I understand correctly how this API works, I will need to add race, ethnicity to REQUIRED_CORE_DATA config list in EPROMs persistence file for my changes to work correctly, right?


pbugni:
@achen2401 Yes, please change to site_persistence if you want race&ethnicity for CRV users.

PR #944, opened 2017-05-23
Add UserDocument PatientReports to reports column

OptimusRhine:
* add UserDocuments of doctype 'PatientReport' to the Reports column on the /patients list
  * includes helper method on User model
  * docs are displayed by filename, and are clickable download links for that doc

### DIFF
--- a/portal/models/user.py
+++ b/portal/models/user.py
@@ -1192,7 +1192,6 @@ class User(db.Model, UserMixin):
             return '<div>' + '</div><div>'.join(
                 [ui.staff_html for ui in uis]) + '</div>'
 
-
     def fuzzy_match(self, first_name, last_name, birthdate):
         """Returns probability score [0-100] of it being the same user"""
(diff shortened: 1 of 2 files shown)


portal/models/user.py
```diff
@@ -1192,6 +1192,21 @@ def staff_html(self):
...
 
+    def report_html(self):
+        """Helper used from templates to display all patient report information
+
+        Combines the intervention links from staff_html() with the patient's
+        list of UserDocuments of type PatientReport (including download links)
+
+        """
+        html = self.staff_html()
+        for doc in self.documents.filter_by(document_type='PatientReport'):
```
pbugni: Not crazy about mixing html and backend code.

Makes sense to provide a filter accessor, but I'd rather it called from the template and marked up there.

Also, use `url_for` rather than hardcoding request paths.

pbugni:
Now that's tidy!

PR #947, opened 2017-05-24
patients list column sort fix

achen2401:
address this story:
https://www.pivotaltracker.com/story/show/142739321

- makes assessment status field sortable AND searchable
  It seems the most clean way to do this is by making the assessment status a property of the user
  otherwise, it seems error-prone trying to acquire assessment status conditionally
- fix consent date field sorting issue AND also making it searchable
- add loading indicator during url redirect

@pbugni  Paul, your feedback on my changes is greatly appreciated.  Thank you very much in advance.


achen2401:
trying a different approach, will have ready shortly

PR #949, opened 2017-05-24
QuestionnaireResponse Entry method capture and reporting

ivan-c:
See [PT#142101133](https://www.pivotaltracker.com/n/projects/1225464/stories/142101133)
* Extend Encounter model to allow [`type`](http://www.hl7.org/FHIR/encounter-definitions.html#Encounter.type) to be captured
* Set session var and infer Encounter.type from it
* Add JSON and CSV reporting of Encounter.type
  * CSV includes human-readable types
    * `entry_method`
    * `author_role`


portal/models/fhir.py
```diff
@@ -199,6 +199,43 @@ def FALSE_VALUE(self):
...
+            display='Information collected over telephone system',
+        ).add_if_not_found(True)
+        cc = CodeableConcept(codings=[coding,]).add_if_not_found(True)
+        assert coding in cc.codings
+        return cc
+
+EC = EncounterConstants()
+
+
+
```
pbugni: PEP8 whitespace rules - 2 empty lines here.

portal/models/fhir.py
```diff
@@ -526,11 +567,31 @@ def consolidate_answer_pairs(answers):
 
         return filtered_answers
 
+    def entry_method(row_data, qnr_data):
+        # Todo: replace with EC.PAPER CodeableConcept
```
pbugni: isn't this TODO done?

portal/models/fhir.py
```diff
@@ -526,11 +567,31 @@ def consolidate_answer_pairs(answers):
...
+        if 'paper' in (c.get('code') for c in qnr_data['encounter']['type']):
+            return 'enter manually - paper'
+        if row_data.get('subject_id') == row_data.get('author_id'):
+            return 'online'
+        else:
```
pbugni: Not sure of the scenario, but why not enforce you recognize the 'code'?
As structured, this is order dependent.  Could have both subject_id == author_id and code == paper..

portal/models/fhir.py
```diff
@@ -526,11 +567,31 @@ def consolidate_answer_pairs(answers):
...
+            return 'enter manually - interview assisted'
+
+    def author_role(row_data, qnr_data):
```
pbugni: I don't believe we need `author_role` - this should be covered by the encounter.

portal/models/fhir.py
ivan-c: I was referring to replacing the string `'paper'` with the coded value `EC.PAPER.code` (or however the correct syntax is)

portal/models/fhir.py
ivan-c: As the UI is implemented, it wouldn't be possible for a patient to enter their own paper version (subject_id == author_id). It is a little fragile but I hope to lean more on Encounter(.type) to improve it.

portal/models/fhir.py
ivan-c: I agree, the same info is available in the Encounter. This just surfaces it as a column in the CSV and gives it a more human-understandable mapping

PR #952, opened 2017-05-24
adding necessary migration script to correct constraints.

pbugni:


pbugni:
This was intended to be part of #951 - overlooked the migration script.

PR #958, opened 2017-05-25
Foreign key null constraints

OptimusRhine:
* added `nullable=false` constraints to the following objects:
  * Client.user_id
  * Grant.user_id
  * Token.user_id
  * QuestionnaireResponse.subject_id
  * QuestionnaireResponse.encounter_id
  * UserIntervention.user_id
  * UserIntervention.intervention_id
  * Procedure.encounter_id
  * UserObservation.encounter_id
* created migration for new `nullable=false` constraint
  * creates Encounters for existing objects that lack them
  * removes these generated Encounters on downgrade
* modified `User.save_constrained_observation()` to add Encounter to new UserObservation (and create UserObservation explicitly, rather than by appending to user.observations, to allow for this Encounter logic)
* modified `Procedure.from_fhir()` to create/set an Encounter as part of the Procedure creation process
* fixed tests to match new nullable=false constraints
  * mostly involves adding `self.login()`, so that in `User.save_constrained_observation()`, the `encounter = get_user(audit.user_id).current_encounter` line returns a value as expected


ivan-c:
@OptimusRhine If you rebase develop pep8speaks should pick up a config change that will make it comment only on lines changed (not the entire file)

OptimusRhine:
@ivan-c ah awesome, I was going to ask about that during the call :)

portal/migrations/versions/823de677d394_.py
```diff
@@ -28,101 +28,107 @@ def create_missing_encounters(table_name):
...
 def remove_generated_encounters(table_name):
     bind = op.get_bind()
     session = Session(bind=bind)
 
-    for obj_id, enc_id in session.execute("SELECT {table_name}.id, encounters.id FROM {table_name} " \
-                    "JOIN encounters ON {table_name}.encounter_id = encounters.id " \
-                    "WHERE encounters.status = 'finished' AND encounters.auth_method = 'staff_authenticated' " \
-                    "AND encounters.start_time = '2000-01-01 01:01:01'".format(table_name=table_name)):
-        session.execute('UPDATE {} SET encounter_id = NULL WHERE id = {}'.format(table_name, obj_id))
+    for obj_id, enc_id in session.execute("SELECT {table_name}.id, " \
```
pbugni: Inside parens, you don't need the line continuation character '\\'

OptimusRhine: Yea, learned that from the pep8 bot :)

ivan-c:
@OptimusRhine FYI I saw a feature mentioned in the docs that will auto-fix all PEP8 issues if you want to give it a try.
>Write @pep8speaks pep8ify in a review summary and it will create a Pull Request with changes suggested by autopep8 against the branch of the author of the PR. autopep8 fixes most of the errors reported by pycodestyle

PR #961, opened 2017-05-25
Smartling ePROMs upload

ivan-c:
* Move Smartling project_id to each persistence file
* Upload translations for each project

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,6 +23,7 @@ env:
 matrix:
   include:
     - env: TOXENV=translations SQLALCHEMY_DATABASE_URI="$SQLALCHEMY_DATABASE_TEST_URI" PERSISTENCE_FILE="https://raw.githubusercontent.com/uwcirg/TrueNTH-USA-site-config/develop/site_persistence_file.json"
+    - env: TOXENV=translations SQLALCHEMY_DATABASE_URI="$SQLALCHEMY_DATABASE_TEST_URI" PERSISTENCE_FILE="https://raw.githubusercontent.com/uwcirg/ePROMs-site-config/develop/site_persistence_file.json"
     - env: TOXENV=docs
       services: []
       addons: {}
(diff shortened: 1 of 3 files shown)


ivan-c:
Looks like we may need an API token per project in Smartling- waiting to confirm with Tyler

.travis.yml
```diff
@@ -22,7 +23,8 @@ env:
 
 matrix:
   include:
-    - env: TOXENV=translations SQLALCHEMY_DATABASE_URI="$SQLALCHEMY_DATABASE_TEST_URI" PERSISTENCE_FILE="https://raw.githubusercontent.com/uwcirg/TrueNTH-USA-site-config/develop/site_persistence_file.json"
+    - env: TOXENV=translations PERSISTENCE_FILE="https://raw.githubusercontent.com/uwcirg/TrueNTH-USA-site-config/develop/site_persistence_file.json"
+    - env: TOXENV=translations PERSISTENCE_FILE="https://raw.githubusercontent.com/uwcirg/ePROMs-site-config/develop/site_persistence_file.json"
```
pbugni: Aren't you simply masking the TrueNTH value of PERSISTENCE_FILE when resetting it for ePROMs?

ivan-c: I don't think PERSISTENCE_FILE is set by default anywhere so those two rows should be the only time it's being set in TravisCI. Each row should run as a separate TravisCI "job" with those env vars set

pbugni: Ah, I didn't understand how that worked.  Makes sense.  What do you make of the failed tests from the PR below?

ivan-c: After emailing with Tyler it seems I was using a project-specific (TrueNTH) access token I had generated instead of an account-specific token that should work across projects.

It seems to be now working after plugging in the new token Tyler gave us (still waiting on tests to finish :) )

PR #963, opened 2017-05-26
Improve error details for strange multi-active encounter situation.

pbugni:
Also, don't generate a stack, but log the error and march on.

### DIFF
--- a/portal/models/user.py
+++ b/portal/models/user.py
@@ -362,8 +362,15 @@ class User(db.Model, UserMixin):
             Encounter.status=='in-progress')
         if query.count() == 0:
             return None
-        assert (query.count() == 1)  # shouldn't have more than one active
-        return query.one()
+        if query.count() != 1:
+            # Not good - we should only have one `active` encounter for
+            # the current user.  Log details for debugging and return the
+            # first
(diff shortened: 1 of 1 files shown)


portal/models/user.py
```diff
@@ -362,8 +362,15 @@ def current_encounter(self):
...
-        return query.one()
+        if query.count() != 1:
+            # Not good - we should only have one `active` encounter for
+            # the current user.  Log details for debugging and return the
+            # first
+            msg = "Multiple active encounters found for {}: {}".format(
+                self.user,
+                [(e.status, str(e.start_time), str(e.end_time)) for e in query])
+            current_app.logger.error(msg)
+        return query.first()
```
ivan-c: Would it make sense to sort by creation time of encounter? Not sure if `first()` would get oldest or newest encounter...

pbugni: No idea - this is an unexpected error situation.  Which would be the correct one to return :)

ivan-c: Good point. I wanted the Encounter returned to be consistent (either first or last) but that doesn't really matter...

ivan-c:
Looks good, not sure what else we could log...

ivan-c:
Merged into develop too FYI

PR #975, opened 2017-05-30
Adding new phone fields to user.merge_with

OptimusRhine:
* noticed that `user.merge_with()` was still looking for old `phone` field; updating to look for the new `phone_id` & `alt_phone_id` fields


pbugni:
Thanks for noticing that - perhaps we should omit the few exceptional attributes rather than naming?

PR #976, opened 2017-05-30
Change user.merge_with attr selection logic

OptimusRhine:
Based on @pbugni 's comment in https://github.com/uwcirg/true_nth_usa_portal/pull/975

* modified db column attribute selection logic in `user.merge_with()` to be exclusionary (just ignore `['id', 'email', 'reset_password_token']`) rather than inclusionary (which requires manually adding fields to the attr list in `merge_with()` every time a new column is added to the model)

### DIFF
--- a/portal/models/user.py
+++ b/portal/models/user.py
@@ -5,7 +5,7 @@ from flask import abort, current_app
 from flask_user import UserMixin, _call_or_get
 import pytz
 from sqlalchemy import text
-from sqlalchemy.orm import synonym
+from sqlalchemy.orm import synonym, class_mapper, ColumnProperty
 from sqlalchemy import and_, or_, UniqueConstraint
 from sqlalchemy.ext.hybrid import hybrid_property
 from sqlalchemy.dialects.postgresql import ENUM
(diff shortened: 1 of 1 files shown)


portal/models/user.py
```diff
@@ -1025,14 +1030,13 @@ def merge_with(self, other_id):
...
-        # intentionally skip {email, reset_password_token}
-        for attr in (
-            'password', 'first_name', 'last_name', 'birthdate', 'phone_id',
-            'alt_phone_id', 'gender', 'locale_id', 'timezone', 'confirmed_at',
-            'registered', 'image_url', 'active', 'deleted_id', 'deceased_id'):
-            if not getattr(other, attr):
-                continue
-            setattr(self, attr, getattr(other, attr))
+        # intentionally skip {id, email, reset_password_token}
+        exclude = ['id', 'email', 'reset_password_token']
```
pbugni: Well done - I like the column_names method.

This will work, but there's a generator pattern you may like, that I think improves legibility:
`for attr in (col for col in self.column_names() if col not in exclude):`

OptimusRhine: Yea, that's cleaner. I'll update

PR #978, opened 2017-05-30
Adding endpoint for front end error reports.

pbugni:
@achen2401 will this work for sending in errors from front end code?


pbugni:
@achen2401 if the logger level is configured to send an email, it'll do so, as is the case on stg & production.

https://github.com/uwcirg/true_nth_usa_portal/blob/develop/portal/app.py#L210-L220

PR #979, opened 2017-05-31
add client side generated server error reporting via ajax

achen2401:
client side error reporting via ajax call to API, see PR #978


pbugni:
Looks like a good start :)

@achen2401 I see a number of `console.log` calls - are any of these worth review for inclusion?

achen2401:
@pbugni Paul, thank you, the console.log calls should now be removed.

PR #993, opened 2017-06-02
remove left navigation panel in profile, layout and bug fixes

achen2401:
address these stories:
https://www.pivotaltracker.com/story/show/146494723
https://www.pivotaltracker.com/story/show/146494631

remove left navigation panel from profile page, fix layout in profile, minor text change in treatments section and slight refactoring of code in profile page
Note, keeping code for left navigation panel commented out so can put it back later if needed again
fix styling issue in initial queries page - progress bar is now drawn in proportion, and fix footer layout issue

### DIFF
--- a/portal/static/css/eproms.css
+++ b/portal/static/css/eproms.css
@@ -229,9 +229,9 @@ select:not([multiple]) {
   z-index: 20;
   position: relative;
 }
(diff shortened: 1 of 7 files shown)


achen2401:
found issue with some of the changes, will investigate

PR #997, opened 2017-06-05
made privacy policy page unavailable for EPROMs

achen2401:
address this story: https://www.pivotaltracker.com/story/show/146635611

For Eproms only, prevents privacy policy page from publicly viewable

### DIFF
--- a/portal/views/portal.py
+++ b/portal/views/portal.py
@@ -665,13 +665,15 @@ def profile(user_id):
 
 @portal.route('/privacy')
 def privacy():
-    """ privacy use page"""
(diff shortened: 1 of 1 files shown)


pbugni:
@achen2401 this change will clearly prevent non-GIL users from being able to access /privacy - can I assume there was a change elsewhere to hide the link?

achen2401:
Paul, I believe this change only applies to Eproms (The GIL privacy page is still available only in Truenth.  Note the difference in templates).  Currently, there is no direct link available to the privacy page in Eproms but technically, a user can access it by typing the url, i.e., /privacy, in the browser.

PR #998, opened 2017-06-05
Add missing delete orphan cascades

OptimusRhine:
* added `cascade='delete'` to:
  * UserObservation.encounter
  * QuestionnaireResponse.encounter
  * Procedure.audit
  * Procedure.encounter
  * User.user_audits (new relationship)
  * User.subject_audits (new relationship)
  * User.auth_providers
  * User._consents
  * User.encounters
  * User.questionnaire_responses
  * User.procedures
  * User.documents
  * User._phone
  * User._alt_phone
(note that any relationship with a `secondary` table, automatically handles cascading deletes for that secondary table, and so needs no explicit `cascade='delete'` addition)
* removed most of `purge_user()` logic, as that's mostly all handled by the cascades now (besides a few unique cases)
* created fairly extensive test for `permanently_delete_user()`

### DIFF
--- a/portal/models/fhir.py
+++ b/portal/models/fhir.py
@@ -390,7 +390,7 @@ class UserObservation(db.Model):
         db.ForeignKey('encounters.id', name='user_observation_encounter_id_fk'),
         nullable=False)
 
-    encounter = db.relationship('Encounter')
+    encounter = db.relationship('Encounter', cascade='delete')
 
     __table_args__ = (UniqueConstraint('user_id', 'observation_id',
         name='_user_observation'),)
@@ -462,7 +462,7 @@ class QuestionnaireResponse(db.Model):
     encounter_id = db.Column(
         db.ForeignKey('encounters.id', name='qr_encounter_id_fk'),
         nullable=False)
-    encounter = db.relationship("Encounter")
+    encounter = db.relationship("Encounter", cascade='delete')
 
     # Fields derived from document content
     status = db.Column(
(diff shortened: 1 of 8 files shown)


OptimusRhine:
Adding `cascade='delete'` to procedures seems to have caused a bunch of procedure creations in different tests to fail. Investigating...

OptimusRhine:
@pep8speaks pep8ify

pbugni:
@OptimusRhine looks great - thanks.  I'm moving forward with the merge.  For testing, we should grab a database copy from something like stage.us.truenth.org and try deletion of several types of long standing users.

OptimusRhine:
@pbugni ah right, I had originally planned to do that. I tested fairly extensively with users in my local db, but I'll grab some db dumps from the eproms and truenth stage instances, load them locally, and try deleting some different users (long standing users, like you mentioned, and/or users with a lot of different related child objects attached to them)

ivan-c:
@OptimusRhine What happened when you did this? Just curious :)
>@pep8speaks pep8ify

OptimusRhine:
@ivan-c strangely, nothing. I waited for like 10 minutes, then just made the changes myself, and deleted the comment.

Not sure if I just need to wait longer (I figured it'd be near-instantaneous, based on how quickly it comments), or if there's some other detail to the call that I'm missing...

Have you managed to use it successfully?

ivan-c:
@OptimusRhine Haven't tried it out but assumed it worked the way you described. I'll try adding a comment like that and going to bed next time it complains about something :)

PR #1002, opened 2017-06-06
TOU 'type' field

OptimusRhine:
https://www.pivotaltracker.com/story/show/146719817

* added new 'type' field to Tou object
  * value must be one of:
    * website terms of use
    * subject website consent
    * stored website consent form
    * privacy policy
* added migration for 'type' field
  * existing Tous are assigned default value of 'website terms of use'
* added new field to Tou creations in tests
* modified /tou/accepted POST logic to use 'type' response value if present, or else default to 'website terms of use'


OptimusRhine:
Please don't merge yet - waiting to hear if current type defaulting logic in /tou/accepted/ POST is correct, or if we should change to always expect a 'type' value in the response json

portal/views/tou.py
```diff
@@ -66,7 +89,63 @@ def get_tou(user_id):
...
+
+    return jsonify(tous=[d.as_json() for d in tous])
+
+
+@tou_api.route('/user/<int:user_id>/tou/<string:tou_type>')
+@oauth.require_oauth()
+def get_tou_by_type(user_id, tou_type):
+    """Access Terms Of Use info for given user & ToU type
+
+    Returns ToU{'accepted': true|false} for requested user given the
```
pbugni: I think we should make these two APIs more consistent - lets return the same JSON for the one, if accepted.  Not sure how to best handle the false case - ideas?

portal/models/tou.py
```diff
@@ -10,9 +15,18 @@ class ToU(db.Model):
...
     """tracks when and by whom the terms were agreed to"""
 
     def __str__(self):
         return "ToU ({0.audit}) {0.agreement_url}".format(self)
+
+    def as_json(self):
+        d = {}
+        d['id'] = self.id
+        d['agreement_url'] = self.agreement_url
+        d['type'] = self.type
```
pbugni: Can we include a field for when it was `accepted`? use the value from the audit.timestamp

portal/views/tou.py
OptimusRhine: @pbugni I'm fine with changing the true case to return the full ToU JSON.
I would say that in the false case, we should either return the `jsonify(accepted=False)` as we do now, or we just return an empty set, which should be taken to mean that no such ToU agreement was found. @achen2401 which makes more sense from your end?

achen2401: @OptimusRhine Adam, I think jsonify(accepted=False) makes more sense to me -  I can check for that first to see if user has not accepted any terms yet

OptimusRhine: OK, cool. So now if a result is found, it'll return something like this:
```
{
  "accepted": "2017-05-22T17:34:24", 
  "agreement_url": "predates agreement_url", 
  "id": 5, 
  "type": "privacy policy"
}
```
whereas if nothing is found, it'll just return
```
{
  "accepted": false
}
```

PR #1008, opened 2017-06-08
Move audit refs from Observation to UserObservation

OptimusRhine:
https://www.pivotaltracker.com/story/show/146721583

* moved `audit_id` db column and `audit` relationship from Observation to UserObservation
  * adding `cascade='delete'` param to `UserObservation.audit` relationship
* migration for column move
  * on upgrade, sets each `UserObservation.audit_id` to the removed `Observation.audit_id` value of the `UserObservation.observation`
  * on downgrade, sets each `Observation.audit_id` to the removed `audit_id` value of the last UserObservation to reference that Observation; then, creates a new audit entry for any Observation still lacking an `audit_id` value
* moved any `audit=audit` param refs in Observation initializations to their corresponding UserObservation initializations
* removed special Observation audit logic from `purge_user()` (now properly handled via delete cascades)
* updated tests


portal/migrations/versions/38cfbd643c73_.py
```diff
@@ -0,0 +1,88 @@
...
+    for obs_id, audit_id in session.execute('SELECT id, audit_id FROM observations'):
+        audit_ids[obs_id] = audit_id
+
+    op.drop_constraint(u'observations_audit_id_fkey', 'observations', type_='foreignkey')
+    op.drop_column('observations', 'audit_id')
+    op.add_column('user_observations', sa.Column('audit_id', sa.Integer(), nullable=True))
+    op.create_foreign_key(u'user_observations_audit_id_fkey', 'user_observations',
+                          'audit', ['audit_id'], ['id'])
+
+    # store existing audit_id values in UserObservations, based on observation_id
```
pbugni: @OptimusRhine thanks for cleaning up the mess!  Won't this create a future problem where the user behind the user_observation isn't necessarily the one who owns the audit?  i.e. when the observation was created, the inital user creating said observation owned the audit - then we added any number of user observations to that.

Truth is, we lack an audit trail here for existing data (except for the first user_observation for each observation added) unless we scrub logs.  It'd be reasonable to generate a filler audit owned by the correct user with a comment, I think - you?

OptimusRhine: @pbugni ah, fair point. How bout I just update that audit's user_id/subject_id with the UserObservation's user_id? Might be better than creating a bunch of new audits.

pbugni: but it's not a 1:1 mapping, is it?  I'm assuming you'll find n > 1 user_observations for each observation.

portal/migrations/versions/38cfbd643c73_.py
```diff
@@ -26,49 +26,41 @@ def upgrade():
...
     op.add_column('user_observations', sa.Column('audit_id', sa.Integer(), nullable=True))
     op.create_foreign_key(u'user_observations_audit_id_fkey', 'user_observations',
                           'audit', ['audit_id'], ['id'])
 
-    # store existing audit_id values in UserObservations, based on observation_id
-    for obs_id in audit_ids:
+    # create new audits for UserObservations
+    for uo_id, user_id in session.execute('SELECT id, user_id FROM user_observations'):
+        aud = Audit(user_id=user_id, subject_id=user_id,
+                    comment="generated as part of migration 38cfbd643c73 upgrade")
```
pbugni: These may well show up in the UI - for a user's history.

How about "entry predates audit records for user observations" or the like?

pbugni: I think you may also want to purge the existing audits set on observations.

PR #1028, opened 2017-06-15
Toggle-able public terms link in footer

OptimusRhine:
* 'Terms' link in footer should only appear when logged in, or when `SHOW_PUBLIC_TERMS` config is `True` (which is the default)
* fixed typo in terms render method


pbugni:
@OptimusRhine thanks for catching that typo!

PR #1031, opened 2017-06-15
Gettext-ify privacy string

OptimusRhine:
*enclosing 'Privacy' str in `{{ _('') }}`


ivan-c:
I think I got this in #1030, should be in develop too now

PR #1034, opened 2017-06-16
Fix Proc DELETE issue

OptimusRhine:
* removed `'delete'` cascade from `Procedure.encounter` relationship, which should fix the error when the UI attempts to delete a Procedure

### DIFF
--- a/portal/models/procedure.py
+++ b/portal/models/procedure.py
@@ -50,7 +50,7 @@ class Procedure(db.Model):
     coding.system is required to be `http://snomed.info/sct`
     """
 
-    encounter = db.relationship('Encounter', cascade='delete',
+    encounter = db.relationship('Encounter',
                                 foreign_keys=[encounter_id])
 
     def as_fhir(self):


pbugni:
@OptimusRhine have you tried `delete-orphan`?  Seems like the appropriate cascade for situations like this.

OptimusRhine:
@pbugni I'm not so sure. For one, it looks like `delete-orphan` assumes `delete` (http://docs.sqlalchemy.org/en/latest/orm/cascades.html#delete-orphan , or if you try adding just the `delete-orphan` cascade for an object and running the server, it'll throw a warning saying the same (`SAWarning: The 'delete-orphan' cascade option requires 'delete'.`)).

Also, the issue here is that the same encounter is referenced by multiple objects (Procedures and UserObservations). So even if we could add the `delete-orphan` cascade to either of those relationships, I think we would still get the same problem (if an Encounter disconnects from a Procedure for whatever reason, and we have the `delete-orphan` cascade, the Encounter would be deleted, which would still cause issues if it's being referenced by a UserObservation).

Definitely open to ideas here though. Right now I've still got the `delete` cascade on the `UserObservation.encounter` relationship, so while we shouldn't have any completely orphaned Encounters, we still might get issues if we attempt to delete a UserObservation that references the same Encounter as an existing Procedure (if that's even possible right now). Maybe it's best if we just remove the `delete` cascade from both, and accept that we may sometimes get orphaned Encounters?

PR #1042, opened 2017-06-19
Merge incoming po files, instead of overwriting

OptimusRhine:
* changed `i18n.smartling_download()` logic - now, when we get PO files from Smartling, instead of overwriting the existing files, we merge any changes from the new files into the existing ones
  * important to merge the changes, because we'll have PO files for the same language coming from (at least) two different projects, each with their own subset of unique strings; just overwriting the files means deleting those unique string translations
* added `polib` library to support PO file handling
* fixed db string extraction to properly escape double-quotes (double backslashes become a single backslash when uploaded to Smartling), to avoid issues with polib during file parsing
* minor fixes and cleanup with general Smartling upload/download process
* updated POT, fr_CA PO, and fr_CA MO files


pbugni:
Looks great @OptimusRhine.  Would it be worth adding some tests?

OptimusRhine:
@pbugni looks like Ivan's actually got some `TOKENENV=translations` tests in place - in fact, you can see where they failed (as they should have) in the `breaking into smaller functions` commit, above

pbugni:
good point - my grep was obviously lacking!

ivan-c:
@OptimusRhine I started a review after you requested but you fixed some of my comments as I was writing them!

portal/models/i18n.py
```diff
@@ -19,12 +21,15 @@ def get_db_strings():
     elements = defaultdict(set)
     for entry in AppText.query:
         if entry.custom_text:
-            elements['"{}"'.format(entry.custom_text)].add("apptext: " + entry.name)
+            ct = re.sub('"', r'\\"', entry.custom_text)
```
OptimusRhine: @ivan-c Wouldn't that be (slightly) less efficient though, cycling back through elements again?

Would it be better just to move the `re.sub` and insertion into `elements` into a separate method, just to avoid duplicating code and improve readability?

ivan-c: Sorry, that comment was from before you refactored it into smaller functions. No need to make any changes :)

PR #1043, opened 2017-06-19
initial queries ui code bug fixes and refactoring

achen2401:
address these issues related to initial queries page:
https://www.pivotaltracker.com/story/show/147461653
https://www.pivotaltracker.com/story/show/147371649

- refactor UI code
- hide terms intro terms as needed (for eproms)
- refactor portal.py code for getting terms app text

### DIFF
--- a/portal/views/portal.py
+++ b/portal/views/portal.py
@@ -16,9 +16,9 @@ from ..audit import auditable_event
 from .crossdomain import crossdomain
 from ..database import db
 from ..extensions import oauth, user_manager
-from ..models.app_text import app_text, AppText, VersionedResource
+from ..models.app_text import app_text, AppText, VersionedResource, UndefinedAppText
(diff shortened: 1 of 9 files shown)


portal/views/portal.py
```diff
@@ -493,16 +484,28 @@ def website_consent_script(patient_id):
...
-    else:
-        terms = VersionedResource(app_text(InitialConsent_ATMA.name_key()))
+    terms = get_terms(org)
     return render_template(
         'website_consent_script.html', user=user, terms=terms,
         entry_method=entry_method, redirect_url=redirect_url,
         patient_id=patient_id)
 
+
+def get_terms(org, role):
```
pbugni: It looks like you want to be able to call `get_terms()` with `org` and `role` as optional parameters.  If so, you need to give them default values, or line 487 is going to raise `TypeError: get_terms() takes exactly 2 arguments (1 given)`

simply change the definition to include the defaults:
`def get_terms(org=None, role=None)`

PR #1050, opened 2017-06-21
Log PatientReport POST 500 errors

OptimusRhine:
* added logging for 500 errors on PatientReports, for easier debugging

### DIFF
--- a/portal/views/user.py
+++ b/portal/views/user.py
@@ -1624,6 +1624,7 @@ def upload_user_document(user_id):
     except ValueError as e:
         abort(400, str(e))
     except OSError as e:
+        current_app.logger.error('patient_report post error - {}'.format(e))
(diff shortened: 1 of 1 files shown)


pbugni:
@OptimusRhine can you test (even if it's just a local one-off attempt) that this generates a log statement (when attempting to write to an existing directory for which the process user doesn't have write permissions)?  I'm surprised that we didn't see one given the 500 that was already in place.

OptimusRhine:
@pbugni looks like it does:
2017-06-21 14:26:45,111 ERROR: patient_report post error - could not save file [in /home/adam/truenth_ss/portal/views/user.py:1627]

pbugni:
Very strange we didn't see this on prod.  Anyhow, thanks!

PR #1051, opened 2017-06-21
add missing translation tags in portal templates

achen2401:
add missing translation tags in portal templates


pbugni:
I haven't merged this PR due to the `o` issue.  If you'd like to push it out, and circle back later once TrueNTH weighs in, that's fine with me.

PR #1052, opened 2017-06-22
discrete celery

pbugni:
Config and task changes to support running unique celery daemon per portal instance.

Added endpoint /celery-info to farm a task to celery to report its configured `CELERY_BROKER_URL` and `SERVER_NAME` - to confirm which service we're hitting.

NB - the actual setup to run a unique celery daemon per portal is driven by puppet changes in the `t__celerydaemonpervhost` branch.  This will also require changes on the portal's application.cfg file such as:

```
# Use a unique redis db (the integer value following the last '/' in the REDIS_URL)
# so named celery daemon only processes requests from this application
REDIS_URL = 'redis://localhost:6379/15'
CELERY_BROKER_URL = REDIS_URL
REQUEST_CACHE_URL =  REDIS_URL
SESSION_REDIS_URL =  REDIS_URL
```


pbugni:
@ivan-c You are welcome to merge this when ready.  We do NOT need the puppet branch merged prior to rolling this out - in fact the sooner the better, as we can test the /celery-info link to see which celery daemon is handling the request.

I would hold off on merging the puppet branch - that will likely be a bit bumpy in bringing the newly named daemons up in place of old, and having the config correctly setup beforehand.

portal/config.py
```diff
@@ -37,14 +36,19 @@ class BaseConfig(object):
...
+    REDIS_DB = os.environ.get(
+        'REDIS_DB',
+        0
+    )
+
     ANONYMOUS_USER_ACCOUNT = True
     CELERY_BROKER_URL = os.environ.get(
         'CELERY_BROKER_URL',
-        REDIS_URL + '/0'
+        '{}/{}'.format(REDIS_URL, REDIS_DB)
```
ivan-c: I think this would break if `REDIS_URL` was set as an env var and already included a DB eg:
```
REDIS_URL = 'redis://hostname:password@example.com:1234/9'
```

pbugni:
@ivan-c as per your suggestion, we're no longer defining a REDIS_DB value, but expecting the REDIS_URL (if defined) to include the trailing `/<int db>` value.

ivan-c:
Great, will merge as soon as tests pass. Go ahead if you get impatient :)

PR #1060, opened 2017-06-23
Kiosk mode 5 minute browser timeout

OptimusRhine:
https://www.pivotaltracker.com/story/show/135168829

* modified browser timeout logic:
  * when using the kiosk (aka `login-as`) functionality, browser timeout gets set to 5 minutes (via cookies, just as we do in the /settings page
    * if admin user already had a custom browser timeout set (via /settings), that value is stored in another temporary cookie
  * when logging back in normally (not using `login-as`), the 5 minute timeout either gets replaced with the previous value (via the temporary cookie), or if there was no previous value, just gets removed (and the system will use the app default)
    * at this time, the temporary cookie is removed (if it exists)

### DIFF
--- a/portal/views/auth.py
+++ b/portal/views/auth.py
@@ -216,8 +216,8 @@ def next_after_login():
     if session.get('login_as_id'):
         if request.cookies.get('SS_TIMEOUT'):
             resp.set_cookie('SS_TIMEOUT_REVERT',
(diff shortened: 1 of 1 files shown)


OptimusRhine:
(please don't merge until @pbugni 's had a chance to review)

PR #1066, opened 2017-06-27
add more informative text for items in consent list for EPROMs

achen2401:
noticed this when testing:
add more descriptive text for items listed in Consent History section for EPROMs (i.e. which type of consent terms the subject agreed to)


achen2401:
@pbugni Paul, thus far, I have not translated text in the js file.  I will do that in a near future when I translate all text in js files (including gil.js for GIL, too) in a separate PR.

PR #1067, opened 2017-06-27
Feature/external study id as constant

pbugni:
Using CONSTANT string in system_uri for external-study-id.

### DIFF
--- a/portal/system_uri.py
+++ b/portal/system_uri.py
@@ -13,6 +13,8 @@ TRUENTH_ENCOUNTER_CODE_SYSTEM = '{}/encounter-types'.format(TRUENTH_NAMESPACE)
 # Auth identities - typically used with suffix for provider (i.e. google)
 # or to name identity type (i.e. TrueNTH-username)
 TRUENTH_IDENTITY_SYSTEM = '{}/identity-codes'.format(TRUENTH_NAMESPACE)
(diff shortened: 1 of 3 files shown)


portal/system_uri.py
```diff
@@ -13,7 +13,9 @@
 # Auth identities - typically used with suffix for provider (i.e. google)
 # or to name identity type (i.e. TrueNTH-username)
 TRUENTH_IDENTITY_SYSTEM = '{}/identity-codes'.format(TRUENTH_NAMESPACE)
-TRUENTH_ID ='{system}/{provider}'.format(
+TRUENTH_EXTERNAL_STUDY_SYSTEM = '{system}/external-study-id'.format(
+    system=TRUENTH_IDENTITY_SYSTEM)
```
ivan-c: I assume `TRUENTH_IDENTITY_SYSTEM` will be the same for ePROMs/TrueNTH is that ok for something like external-study-id?

pbugni:
@ivan-c you'll note i opted to squeeze in a related need, for https://www.pivotaltracker.com/story/show/147878139

portal/system_uri.py
pbugni: We're simply using that as a namespace.  See other FHIR systems as examples.  I don't see any need and expect it'd only create bugs to distinguish on ePROMs/TrueNTH.

PR #1071, opened 2017-06-28
Feature/org in tou

pbugni:
ToU now houses an optional organization_id.
Necessary for https://www.pivotaltracker.com/story/show/147655653


achen2401:
@pbugni Paul I made a tiny fix.  Let me know if you see any problem with my change.

pbugni:
@achen2401 of course - you'll want that data in the JSON if defined.  I altered slightly to avoid including null values in the json.

PR #1081, opened 2017-06-29
Swagger requires unique symbols for return ids.

pbugni:
Replaced generic `response` with distinct names so the resulting swagger
reports the desired return description.

Fix for https://www.pivotaltracker.com/story/show/148112793


pbugni:
@ivan-c Bonus points for turning this around soon - I'd like it to be live on stg by morning to fix the swagger for Weldon's use.

PR #1083, opened 2017-06-30
Support for scheduled jobs via celery

OptimusRhine:
* added new ScheduledJob class
  * stores task, args, and schedule (in cron format)
  * added migration for new task
* on celery-worker start, now adds all active ScheduledJobs to Celery as PeriodicTasks
* added `celerybeat-schedule` to gitignore (it's a file that just stores the last runtimes of celery beat tasks
* do not include celery_worker autodoc
  * waiting for `celery.contrib.sphinx` fix to get pushed in next Celery version

Note that we'll need to slightly change how we start Celery, to include beat:
http://docs.celeryproject.org/en/latest/userguide/periodic-tasks.html#starting-the-scheduler


pbugni:
NB - we need to merge the puppet branch `t_celerydaemonpervhost` into develop/master before merging this PR.

OptimusRhine:
@pbugni Good question on tests - not sure how possible it will be to create a generic Celery beat scheduler test that's task-agnostic, but I'll look into it.

Right now we only add `active=True` ScheduledJobs to Celery, so you could hypothetically "cancel" a job by deactivating it (via the db, I guess - creating an admin-only api + ui for the jobs is on my to-do list) and restarting celery_worker. But obviously that's not a great permanent solution. Adding logic to dynamically canceling/pausing/terminating jobs without restarting Celery...initial investigation shows that the logic is fairly complex, but not impossible.

I'll start work on those, but if `t_celerydaemonpervhost` is merged, please go ahead and merge this as well. I'd rather handle at least the pause/terminate job logic in another PR anyway, just to prevent this PR from getting too busy.

pbugni:
@OptimusRhine I agree - another PR is fine.

Consider some form of a view for looking at tasks - perhaps that could be used for deactivating / deleting.  Not sure yet how we'll want to configure / add such tasks to a system, suggestions?

OptimusRhine:
@pbugni Yea, I was thinking a UI that allowed for creating/inactivating/etc ScheduledJobs, which includes a dropdown list of available task methods.

Right now, the only way to add new task methods to the system is to add them to the `tasks.py` file via a PR. We could try other approaches, such as importing site-specific `tasks.py` files (that live in the eproms/truenth config repos).

But I assume we don't want any sort of UI or even API option to add new task methods; that opens the doors for people to insert and run any code into the system without any sort of oversight, which seems very dangerous (even if we restrict it to say, just admins).

OptimusRhine:
**Current Status**
I'm making a few changes here:
* Rebasing code, updating migrations to base off most recent version
* Modifying scheduler code to allow for dynamic runtime periodic task modification (this one's gonna take some time)
* Adding post and put endpoints for above
* Adding lots of tests

We can hold off on merging this until I've completed the above.

OptimusRhine:
@pbugni I think (once checks finish) this should be ready for you to re-review and merge.

pbugni:
The puppet branch `t_celeryheartbeat` is installed and apparently working on docker-dev-1 and promethium.  Plan to merge this and roll out there shortly.

PR #1094, opened 2017-07-06
Recaptcha for contact form

OptimusRhine:
https://www.pivotaltracker.com/story/show/148387559

* adding flask-recaptcha to app
  * added config values for captcha (connects to `bob25mary@gmail.com` admin panel) - should get pulled from `os.environ`
* adding recaptcha to `/contact` page
  * only appears (and subsequently checks) if user is not logged in, and proper configs are set
  * if recaptcha cannot be verified on submit, reloads `/contact` page instead of attempting to send a contact message

### DIFF
--- a/portal/config.py
+++ b/portal/config.py
@@ -110,6 +110,10 @@ class BaseConfig(object):
     SMARTLING_USER_SECRET = os.environ.get('SMARTLING_USER_SECRET', None)
     SMARTLING_PROJECT_ID = os.environ.get('SMARTLING_PROJECT_ID', None)
 
+    RECAPTCHA_ENABLED = True
+    RECAPTCHA_SITE_KEY = "6LcNNigUAAAAAIgA9i1ZUeGzDf2dJg5ghMUHVggR"
+    RECAPTCHA_SECRET_KEY = "6LcNNigUAAAAAFp84MbFhVMj64SmcfhQZpFRk8ai"
+
 
 class DefaultConfig(BaseConfig):
     """Default configuration"""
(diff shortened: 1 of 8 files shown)


portal/config.py
```diff
@@ -110,6 +110,10 @@ class BaseConfig(object):
     SMARTLING_USER_SECRET = os.environ.get('SMARTLING_USER_SECRET', None)
     SMARTLING_PROJECT_ID = os.environ.get('SMARTLING_PROJECT_ID', None)
 
+    RECAPTCHA_ENABLED = True
+    RECAPTCHA_SITE_KEY = "6LcNNigUAAAAAIgA9i1ZUeGzDf2dJg5ghMUHVggR"
+    RECAPTCHA_SECRET_KEY = "6LcNNigUAAAAAFp84MbFhVMj64SmcfhQZpFRk8ai"
```
pbugni: Are these values required to be site specific?  i.e. can we use the same on eproms and truenth and on all the flavors?

ivan-c: I'm hesitant to hardcode config values for a specific service...
Shouldn't the secret key stay secret too? ;)

setup.py
```diff
@@ -54,6 +54,7 @@
         "Flask-Celery-Helper",
         "Flask-Migrate",
         "Flask-OAuthlib",
+        "Flask-Recaptcha",
```
ivan-c: I think you'll need to add it to `requirements.txt` too; `pip freeze` should tell you the currently installed version...

portal/config.py
OptimusRhine: Not required to be site-specific; we just need to add the relevant domains to the whitelist in the [google recaptcha admin panel](https://www.google.com/recaptcha/admin) (which should already be done).

We COULD use different site keys if we wanted to keep track of verification attempts in the admin panel on a per-site basis. But for the initial implementation I don't think that's necessary.

setup.py
OptimusRhine: Ah right, forgot that we also needed it in `requirements.txt`. Should be there now.

PR #1102, opened 2017-07-08
exclude link(s) to home page from being disabled in initial queries, …

achen2401:
Found these while testing:
-  allow link(s) to home remain enabled in the initial queries page
-  remove code (e.g. window.focus) that is not working consistently across browsers


achen2401:
found issue with these changes.  closing this for now.

PR #1105, opened 2017-07-10
move settings/KEY under the common api/ request path

pbugni:
as per https://www.pivotaltracker.com/story/show/147150023


pbugni:
Rolling out for Gifford to work with immediately.

PR #1107, opened 2017-07-10
Check oauth config during oauth login

OptimusRhine:
* on oauth login:
  * explicitly check that the given `provider_name` is valid given our enum `providers_list`
  * explicitly check that we have the correct configurations (`CONSUMER_KEY` and `CONSUMER_SECRET`) for the given provider.

### DIFF
--- a/portal/views/auth.py
+++ b/portal/views/auth.py
@@ -282,7 +282,7 @@ def login(provider_name):
     prv = 'FB' if (provider_name == 'facebook') else provider_name.upper()
 
     if (not current_app.config.get('{}_CONSUMER_KEY'.format(prv)) or
-        not current_app.config.get('{}_CONSUMER_SECRET'.format(prv))):
+            not current_app.config.get('{}_CONSUMER_SECRET'.format(prv))):
         abort(500, "no configuration found for provider "
               "{}".format(provider_name))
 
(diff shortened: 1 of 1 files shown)


OptimusRhine:
More or less a duplicate of https://github.com/uwcirg/true_nth_usa_portal/pull/1106 . Closing.

portal/views/auth.py
```diff
@@ -276,6 +276,15 @@ def picture_url(result):
         from werkzeug.datastructures import ImmutableMultiDict
         request.args = ImmutableMultiDict()
 
+    prv = 'FB' if (provider_name == 'facebook') else provider_name.upper()
+
+    if (not current_app.config.get('{}_CONSUMER_KEY'.format(prv)) or
```
ivan-c: Would this also catch when `*_CONSUMER_KEY` is not defined at all (as opposed to just being an empty string)?

 I think you may need to add a second/default arg to `get()`...

pbugni: the second arg defines a default - which is itself `None` if not provided.  this code looks to be doing what we need, to me.

portal/views/auth.py
```diff
@@ -24,7 +24,7 @@
 from ..date_tools import FHIR_datetime
 from ..extensions import authomatic, oauth
 from ..models.auth import AuthProvider, Client, Token, create_service_token
-from ..models.auth import validate_client_origin
+from ..models.auth import providers_list, validate_client_origin
```
pbugni: Looks like you are no longer using `providers_list`.

portal/views/auth.py
OptimusRhine: @ivan-c shouldn't we be catching cases like that as well?

PR #1118, opened 2017-07-11
add button--teal css class back to login-register button

achen2401:
found this while testing in stg.us:

fix styling of login button in login/register modal for GIL


achen2401:
**note**  not needed for the release today

PR #1122, opened 2017-07-12
Import locale selector method

OptimusRhine:
Looks like this line somehow got removed...

Note that we NEED this line for translations to work properly in the portal. For whatever reason, when instances are deployed through apache, our replacement `locale_selector` method isn't registered (and therefore doesn't get used). This import line assures that our method gets pulled in.

### DIFF
--- a/portal/views/portal.py
+++ b/portal/views/portal.py
@@ -52,6 +52,11 @@ def server_error(e):  # pragma: no cover
 
 
 @portal.before_app_request
+def assert_locale_selector():
+    # Confirm import & use of custom babel localeselector function
+    assert get_locale()
+
+@portal.before_app_request
 def debug_request_dump():
     if current_app.config.get("DEBUG_DUMP_HEADERS"):
         current_app.logger.debug(
--- a/portal/views/portal.py
+++ b/portal/views/portal.py
@@ -53,7 +53,9 @@ def server_error(e):  # pragma: no cover
 
 @portal.before_app_request
 def assert_locale_selector():
-    # Confirm import & use of custom babel localeselector function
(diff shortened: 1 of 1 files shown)


pbugni:
Sorry, probably something I did as my editor complained:

`portal/views/portal.py|25 col 1 error| '..models.i18n.get_locale' imported but unused [F401]`

Can you add a dummy call to silence this warning, and the explanatory comment?

pbugni:
@OptimusRhine Looking up the decorator you implemented with get_locale, the comment suggests why this is a problem:

```
   def localeselector(self, f): 
          """Registers a callback function for locale selection.  The default
          behaves as if a function was registered that returns `None` all the
          time.  If `None` is returned, the locale falls back to the one from
          the configuration.
          
          This has to return the locale as string (eg: ``'de_AT'``, ``'en_US'``)
          """
          assert self.locale_selector_func is None, \
              'a localeselector function is already registered'
          self.locale_selector_func = f 
          return f
```

I wonder if we don't want to put this more in the intentional path?  If nothing else, the dummy call I alluded to above could assert we don't get back null.

portal/views/portal.py
```diff
@@ -52,6 +52,11 @@ def server_error(e):  # pragma: no cover
 
 
 @portal.before_app_request
+def assert_locale_selector():
+    # Confirm import & use of custom babel localeselector function
```
pbugni: might add so this doesn't get removed by some over zealous optimizer:

```
# Necessary to import get_locale to bring into the request scope to
# prevent the default babel locale selector from being used.
```

OptimusRhine: @pbugni Ok, added

PR #1124, opened 2017-07-12
Secure cookies

ivan-c:
Only set cookies over "secure" channels (HTTPS) for non-dev deployments

See [rfc6265](https://tools.ietf.org/html/rfc6265#section-4.1.2.5)


ivan-c:
Whoops, PR created from wrong branch

PR #1129, opened 2017-07-12
account creation retry changes

achen2401:
Follow up on account creation error messages stemming from request timing out/network issues

- remove static functions and change them to methods of an object that also handle retries of ajax calls for account creation
- remove unused code

### DIFF
--- a/portal/static/js/main.js
+++ b/portal/static/js/main.js
@@ -3769,6 +3769,9 @@ function getSaveLoaderDiv(parentID, containerID) {
     };
 };
 
+function __getLoaderHTML() {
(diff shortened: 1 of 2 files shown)


pbugni:
Looks like a healthy refactor!

@achen2401   Lots of code there - I trust you found a way to test a few cycles of timeouts?

achen2401:
@pbugni 
Paul, yes, I tested a few cycles by changing the browser load speed setting, shortening the timeout of ajax call, passing bad parameters to call, etc.  all of which was to force the ajax call for creating account to fail and retry again.  I logged attempts in browser console while testing just to make sure that retries did take place.

PR #1134, opened 2017-07-13
Properly handle edgecases and throw ValueErrors for org_id

OptimusRhine:
https://www.pivotaltracker.com/story/show/148701941

* we're already capturing and aborting on `ValueError`s in `set_valid_consents()` in the user API; however, we weren't properly throwing `ValueErrors` on bad data in `UserConsent.from_json()`. Modified the logic slightly and added some extra error handling, so we can avoid SQL exceptions (see PT story).

### DIFF
--- a/portal/models/user_consent.py
+++ b/portal/models/user_consent.py
@@ -104,9 +104,11 @@ class UserConsent(db.Model):
         user = 'user_id' in data and User.query.get(data['user_id'])
         if not user:
             raise ValueError("required user_id not found")
-        org = Organization.query.get(data.get('organization_id'))
-        if not org:
+        if not 'organization_id' in data:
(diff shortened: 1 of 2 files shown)


portal/models/user_consent.py
```diff
@@ -104,17 +104,19 @@ def from_json(cls, data):
         user = 'user_id' in data and User.query.get(data['user_id'])
         if not user:
             raise ValueError("required user_id not found")
-        org = Organization.query.get(data.get('organization_id'))
-        if not org:
+        if 'organization_id' not in data:
             raise ValueError("required organization_id not found")
+        org_id = int(data.get('organization_id'))
```
pbugni: should this be using the check_int call from the other PR?

PR #1142, opened 2017-07-18
Validate redirect URLs

OptimusRhine:
https://www.pivotaltracker.com/story/show/149043973

* added validation checks on client redirect URLs (using `validate_client_origin()`) for:
  * `/website-consent-script/<int:patient_id>` [GET] - check `redirect_url`
  * `/login/<provider_name>` [GET] - check optional `next` param
* created new `validate_local_origin()` method to check if given URL is "local" (netloc matches server name, or the URL is just a relative path)
* added validation checks on local redirect URLS (using new `validate_local_origin()`) for:
  * `/challenge` [POST] - check `form.next_url`
  * `capture_next_view_function` - check `next`
* created tests for generate origin validation methods + specific endpoint redirect scenarios

### DIFF
--- a/portal/models/auth.py
+++ b/portal/models/auth.py
@@ -353,6 +353,29 @@ def validate_client_origin(origin):
     abort(401, "Failed to validate origin %s" % origin)
 
 
+def validate_local_origin(origin):
+    """Validate that the origin is the local server
+
+    Origin should either match the server name, or be a relative path.
+
+    :raises :py:exc:`werkzeug.exceptions.Unauthorized`: if we don't
+      find a match.
+
+    """
+    if not origin:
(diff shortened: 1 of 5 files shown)


pbugni:
@OptimusRhine if you don't mind - it'd be great to have tests to verify each case errors out correctly.  This will help guard against regression - i.e. should /website-consent-script get modified and forget to check the redirect_url.

OptimusRhine:
@pbugni yea, your 'http://...' suggestion removed the need for the 'localhost' check, so those should both be fixed now.

Also added specific endpoint tests for the redirect scenarios (grouped them all in `test_portal.test_redirect_validation`, as not all these endpoints have their own unit test yet)

PR #1144, opened 2017-07-19
Feature/csrf

pbugni:
Significant changes to confront the CSRF vulnerability as mentioned in issue [#149043773](https://www.pivotaltracker.com/story/show/149043773)


portal/csrf.py
```diff
@@ -0,0 +1,57 @@
...
+        # if it's valid.  That will be handled in time with the OAuth
+        # decorators on the respecive views.  Confirm we don't ALSO have what
+        # looks like a valid local cookie auth, as they shouldn't ever both
+        # be present.
+
+        if request.headers['Authorization'].startswith('Bearer '):
+            # As this function is called before the OAuth decorator has had
+            # a chance, we should never see a current user (unless one is tied
+            # to a local login session).
+            if current_user():
```
ivan-c: Is this effectively checking for the presence of a session/cookie (local login seems a little ambiguous)? Might be good to note that if it's the case...

portal/csrf.py
```diff
@@ -0,0 +1,57 @@
...
+    # Don't get in the way of the initial oauth dance.
+    if request.path.startswith('/oauth/'):
+        return
+
+    # Backdoor for testing
+    if current_app.config.get('TESTING') == True:
+        return
+
+    # Look for legit OAuth requests, and exclude these from csrf protection
+    if request.headers and request.headers.get('Authorization'):
```
ivan-c: Is it necessary to check if `request.headers` has a value? Shouldn't at least be an empty dict if no headers are set?

pbugni: i sure would think it would always be there - just being careful given the frequent use of this method in the call path. fair point - i'll eliminate the redundant check.

PR #1147, opened 2017-07-19
Show same error message for invalid email & invalid password

OptimusRhine:
https://www.pivotaltracker.com/story/show/149172459

After trying a couple different approaches (including extending out a whole new LoginForm class), I found that there's actually a simple config change for this :)

* set `USER_SHOW_USERNAME_EMAIL_DOES_NOT_EXIST = False`, which turns off specific error messages for invalid email vs. invalid password, and just shows one message ('Incorrect Email and/or Password') instead.


pbugni:
Nice find.

PR #1149, opened 2017-07-19
add missing csrf header in UI forms

achen2401:
per issue found by Sie-ce -
add missing csrf token field to forms that need them


pbugni:
Love the fix, but I'm concerned we may have overlooked others.

@achen2401 could you maybe grep through the templates - see if any other forms with PUT/POST/DELETE actions need this change as well?

achen2401:
@pbugni 
Paul, I did grep through the templates to make sure that we didn't miss any form.  These are the ones I found that were missing the token.

PR #1150, opened 2017-07-19
fix dashboard content for expired patient that completed the question…

achen2401:
address this issue:
https://www.pivotaltracker.com/story/show/148875877

add logout button for dashboard display for expired patient that completed questionnaire

### DIFF
--- a/portal/models/intervention_strategies.py
+++ b/portal/models/intervention_strategies.py
@@ -241,7 +241,7 @@ def update_card_html_on_completion():
                   </div>
                 </div>""".format(
                     greeting=greeting, confirm=confirm, reminder=reminder,
-                    logout = logout)
+                    logout=logout)
 
         def intro_html(assessment_status):
             """Generates appropriate HTML for the intro paragraph"""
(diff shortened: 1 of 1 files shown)


portal/models/intervention_strategies.py
```diff
@@ -439,9 +439,17 @@ def completed_card_html(assessment_status):
...
-                    name=user.display_name,
-                    registry=assessment_status.organization.name)
+                logout_label = _("Log Out")
+                card_html = """
+                            {thank_you}
+                            <div class="button-container portal-header-logout-container">
+                                <a class="btn-lg btn-tnth-primary" href="/logout">{logout}</a>
+                            </div>
+                            """.format(
+                                      thank_you=thank_you_block(name=user.display_name,
```
pbugni: I'm unclear about why this case of "thank you" would be different from the other.  Shouldn't we simply update the `thank_you_block` function to include the logout button?

achen2401: make sense, Paul.  will make the change

PR #1155, opened 2017-07-20
Handle ValueErrors in when comparing org trees

OptimusRhine:
https://www.pivotaltracker.com/story/show/148436359

* when comparing org levels (e.g. comparing staff and patient orgs during `check_role()`), properly handle `ValueError`s, so that, among other things, we properly handle patients who just belong to org id=0 ('None of the above').


pbugni:
Clean fix.  thanks.

PR #1157, opened 2017-07-20
hide continue button from being printed

achen2401:
found this while testing -
minor fix to hide continue button when printing the consent form


achen2401:
will combine this with other fixes

PR #1158, opened 2017-07-20
terms display fix

achen2401:
per conversation with Justin:
hide terms only when staff agreed to terms for subject via manual entry mode
still display it in initial queries page


achen2401:
will combine this with other fixes.

PR #1159, opened 2017-07-20
put in temporary fix for kiosk logout (will be replaced with backend changes)

achen2401:
address this story https://www.pivotaltracker.com/story/show/149024499

as per conversation with Justin, putting in temporary fix for kiosk logout, will be replaced with changes with backend later on.


pbugni:
@achen2401 if you rebase with develop, the tests should pass.

achen2401:
Paul, I will rebase.  Need to close it for now to make an adjustment.

PR #1160, opened 2017-07-20
Update tests for changes to website-consent-script API

OptimusRhine:
* updating `test_portal.test_redirect_validation()` to reflect changes in https://github.com/uwcirg/true_nth_usa_portal/pull/1156

### DIFF
--- a/tests/test_portal.py
+++ b/tests/test_portal.py
@@ -248,7 +248,7 @@ class TestPortal(TestCase):
 
         # validate redirect of /website-consent-script GET
         rv = self.client.get('/website-consent-script/{}?redirect_url='
-                             '{}'.format(TEST_USER_ID, client_url))
+                             '{}'.format(TEST_USER_ID, local_url))
         self.assert200(rv)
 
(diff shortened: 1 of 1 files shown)


pbugni:
@OptimusRhine I see you and I fixed the same error, but in different ways.  I've been waiting on my local tests to pass - but what I did was to load the test client with the configured value of `SERVER_NAME` so it'll include the port.  This seems like a better solution in general.  Mind if I close this one?

OptimusRhine:
@pbugni Ah, I see. Well since #1156 changed the validation from `validate_client_origin()` to `validate_local_origin()`, I figured it made sense just to update the test to check against the `local_url` instead of the `client_url`.

My only worry with loading the `SERVER_NAME` into the `client._redirect_uris`, is that now that `client_url` will now pass both the different validation checks. So if someone unintentionally changes an instance of `validate_client_origin()` to `validate_local_origin()` (or vice versa) in the future, we won't catch it.

OptimusRhine:
(although I guess that would be an odd thing to get changed _unintentionally_ ...)

PR #1174, opened 2017-07-24
Validate Client callback_url against Client redirect_uris

OptimusRhine:
https://www.pivotaltracker.com/story/show/149047283

* added new `Client.callback_url` validation on `/client/<client_id> [POST]`
  * `callback_url`'s `{scheme}://{netloc}` must match that of any of the provided `redirect_urls`
  * only triggers if `callback_url` is provided (i.e. the field is still optional)
* modified `test_auth.test_client_edit`, so that client's new `callback_url` is also present in its list of `redirect_urls`

### DIFF
--- a/portal/views/auth.py
+++ b/portal/views/auth.py
@@ -575,7 +575,7 @@ class ClientEditForm(FlaskForm):
     def validate_callback_url(form, field):
         origins = form.application_origins.data.split()
         og_uris = ['{uri.scheme}://{uri.netloc}'.format(uri=urlparse(url))
-                       for url in origins]
+                   for url in origins]
         if field.data:
             cb_uri = urlparse(field.data)
             if '{uri.scheme}://{uri.netloc}'.format(uri=cb_uri) not in og_uris:
--- a/portal/views/auth.py
+++ b/portal/views/auth.py
@@ -17,6 +17,7 @@ from wtforms import BooleanField, FormField, HiddenField, SelectField
 from wtforms import validators, TextField
 from werkzeug.exceptions import Unauthorized
 from werkzeug.security import gen_salt
+from urlparse import urlparse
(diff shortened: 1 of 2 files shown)


portal/views/auth.py
```diff
@@ -571,6 +572,16 @@ def validate_application_origins(form, field):
             if not url_validation(url, require_tld=False):
                 raise validators.ValidationError("Invalid URL")
 
+    def validate_callback_url(form, field):
```
ivan-c: Could you add a docstring here, describing what this validation function does?

tests/test_auth.py
```diff
@@ -65,15 +65,17 @@ def test_client_bad_add(self):
     def test_client_edit(self):
```
ivan-c: Could you add another test, testing that submitting a CB url with a different domain does in fact cause an error (regression test)?

OptimusRhine:
@ivan-c Thanks Ivan! Added description to that new validation function, and added a new negative test.

tests/test_auth.py
```diff
@@ -65,15 +65,28 @@ def test_client_bad_add(self):
...
-        self.assertEquals(client.callback_url, 'http://tryme.com')
+        self.assertEquals(client.callback_url, test_url)
+
+        invalid_url = "http://invalid.org"
+        rv2 = self.client.post('/client/{0}'.format(client.client_id),
+                data=dict(callback_url=invalid_url,
+                          application_origins=origins,
+                          application_role=INTERVENTION.DEFAULT.name))
+        # 200 response, because page is reloaded with validation errors
+        self.assert200(rv2)
```
ivan-c: Could you make the assertion more specific, eg check for the specific error in the reponse (`rv2.data`)?

PR #1181, opened 2017-07-26
Escape HTML entities in user.display_name

OptimusRhine:
https://www.pivotaltracker.com/story/show/149043645

* added `cgi.escape()` to `User.display_name` return, so that when said `display_name` is pulled for `intervention_strategy` strings, etc, we pre-escape potentially dangerous HTML entity tags ('<', '>', and '&').

### DIFF
--- a/portal/models/user.py
+++ b/portal/models/user.py
@@ -1,4 +1,5 @@
 """User model """
+from cgi import escape
 from datetime import datetime
 from dateutil import parser
 from flask import abort, current_app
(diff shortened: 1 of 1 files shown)


ivan-c:
Worth noting I found [`flask.escape` ](http://flask.pocoo.org/docs/0.10/api/#flask.escape)which does the same thing, but allegedly has some way to prevent double escaping by returning a `Markup` object instead of an actual string.

```
>>> import cgi
>>> import flask
>>> tag = "<script>"

>>> cgi.escape(tag)
'&lt;script&gt;'
>>> cgi.escape(cgi.escape(tag)) # double escaping happening (&)
'&amp;lt;script&amp;gt;'

>>> flask.escape(tag)
Markup(u'&lt;script&gt;')
>>> flask.escape(flask.escape(tag)) # No double escaping
Markup(u'&lt;script&gt;')

```

PR #1182, opened 2017-07-26
front-end changes to convert consent date based on patient timezone

achen2401:
address this story: https://www.pivotaltracker.com/story/show/149497511

@ivan-c 
Ivan, here is a little detail regarding my first attempt using Intl object, which failed in browsers other than Chrome:
I included polyfill.io e.g.<script src="https://cdn.polyfill.io/v2/polyfill.min.js?features=Intl.~locale.en"></script>  error received: RangeError: timeZone is not supported.  

Turning to use moment library instead.  works cross browsers when testing.


ivan-c:
Didn't have a chance to try this out, going to revert

PR #1187, opened 2017-07-27
Check for NUL in access_token request form

OptimusRhine:
https://www.pivotaltracker.com/story/show/148694671
https://www.pivotaltracker.com/story/show/148455555

* added check for NUL (`\0`) char in request form fields for `oauth/token' endpoint

This is a tricky situation, because our code in `auth.access_token()` is wrapped by the actual built-in flask oauth token_handler functionality; prior to this in fact, our `access_token()` method did nothing but `return None`. So we can't just wrap it in a try/except and handle the `ValueError`, because the sqlalchemy logic is happening outside of what we can control. We also can't modify the `request.form` fields (and just convert the data into something acceptable), because the object's immutable.

Current approach is just to straight-up check for the invalid NUL char in all the fields before continuing on. Open to other approaches though, especially if there's a straight-up validation check for sqlalchemy string literals that I'm not aware of.

### DIFF
--- a/portal/views/auth.py
+++ b/portal/views/auth.py
@@ -1047,6 +1047,9 @@ def access_token():
               description: The authorized scopes.
 
     """
+    for field in request.form:
+        if '\0' in request.form[field]:
+            abort(400, "invalid {} string".format(field))
     return None
 
 


pbugni:
@OptimusRhine I agree, this is complicated.

The check you added looks fine, except I don't believe it'll ever get run.  Don't you think such an input would raise an error in the `@oauth.token_handler` decorator first?

As I recall, this issue was filed to confront noise from the penetration testing.  We may just have to live with the noise - I don't believe it generated an actual exploit?

OptimusRhine:
@pbugni No, whatever code we add here gets run before the rest of the built-in `token_handler` code. I tested it, and this fix definitely catches NUL chars in the input fields before anything else is attempted with them. So it _is_ effective, even if it's a little hacky.

If we don't care about that error popping up in the logs, I'm fine with not merging this. But it _does_ work as a validation check, if we do want to avoid that noise.

PR #1190, opened 2017-07-27
eproms patients list performance fix

achen2401:
address story: https://www.pivotaltracker.com/story/show/149615045

- remove code to load assessment status in the backend and retrieve the data via ajax instead
  assessment statuses on the visible rows will be retrieved first
  then the rest will be retrieved via a series of ajax calls (15 each) separated by a mili-seconds 
  separation  

- fix frontend sorting column issue
- add helper methods to admin tool

hope this will take.  tested in different browsers

### DIFF
--- a/portal/static/js/admin.js
+++ b/portal/static/js/admin.js
@@ -106,14 +106,15 @@ AdminTool.prototype.getData = function(userString, callback) {
                         });
         self.ajaxRequests.push(ajaxRequest);
 };
-AdminTool.prototype.loadData = function(list, callback) {
+AdminTool.prototype.loadData = function(list, callback, timeout) {
     var self = this;
     self.requestsCounter = list.length;
     $("#admin-table-error-message").text("");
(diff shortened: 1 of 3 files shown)


portal/static/js/admin.js
```diff
@@ -17,143 +19,176 @@ AdminTool.prototype = Object.create(OrgTool.prototype);
...
+        };
+   });
+   this.initUserList = _userIds;
+   return _userIds;
+};
+AdminTool.prototype.abortRequests = function(callback) {
+   //NEED TO ABORT THE AJAX REQUESTS OTHERWICH CLICK EVENT IS DELAYED DUE TO NETWORK TIE-UP
+   if (AT.ajaxRequests.length > 0) {
+      AT.ajaxRequests.forEach(function(request, index, array) {
+          console.log("WTF??")
```
pbugni: looks like we could omit this from the log?  :)

PR #1194, opened 2017-07-28
Fold validate_local_origin into validate_client_origin

OptimusRhine:
* added local validations (origin's netloc matches the server_name, or origin is a local path) to `validate_client_origin()`
* removed `validate_local_origin()`
* changed calls of `validate_local_origin()` to `validate_client_origin()`
* updated tests

There are some cases popping up where we use the `next=` redirect for either local OR client origins. For example, with `/user/register?next=...`, we sometimes pass a local URL, sometimes a client, depending on how the user was sent to the register page.

As far as I am aware, there are no cases where we want to allow client origins but specifically _exclude_ local origins (or vice versa). So, rather than create a _third_ origin validation method for situations allowing both client and local origins (to handle the likes of `/user/register`), it makes more sense to keep things simple, and merge them into one single validation.

This should solve the issue we're seeing in stg.us.truenth.org, where users attempting to register via SR are getting a '401 Unauthorized: Failed to validate origin' error.

### DIFF
--- a/portal/models/auth.py
+++ b/portal/models/auth.py
@@ -332,7 +332,7 @@ def save_token(token, request, *args, **kwargs):
     return tok
 
 
-def validate_client_origin(origin):
+def validate_origin(origin):
     """Validate the origin is one we recognize
 
     For CORS, limit the requesting origin to the list we know about,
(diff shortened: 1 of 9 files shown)


pbugni:
I love that you merged this.  But I think the name `validate_client_origin` is now misleading, as it isn't necessarily a client origin we're validating.  How about simply `validate_origin`?

PR #1196, opened 2017-07-31
address bootstrap table updateByRow method issue - exposed when testing in IE

achen2401:
the fix last week to address blank rows rendered in patients list in IE is only partial as it only addressed the symptom, **not** the source of the issue.  
I took the time to study the Boostraptable JS code, and realized that the Bootstrap method used to update the assessment status, updateRowByUniqueID, will actually **wipe out** data for the row, if only partial data (e.g. status) is provided for updating - hence the blank rows was displayed in IE.  The issue was not detected in other browsers because a call to reset the table view is made subsequently within the same method.  In IE, the call to reset table view was not made likely due to the interruption in ajax call for updating row data, caused possibly by the limit of concurrent ajax calls in IE or network tie-up related to concurrent ajax calls (which was fixed in last week's fix).

These changes will:
- remove bandaid code
- provide entire row data, including updated assessment status, to the Bootstrap method for updating row

### DIFF
--- a/portal/static/js/admin.js
+++ b/portal/static/js/admin.js
@@ -8,6 +8,7 @@ function AdminTool (userId) {
   this.initUserList = [];
   this.ajaxRequests = [];
   this.ajaxAborted = false;
+  this.arrData = {};
   OrgTool.call(this);
 };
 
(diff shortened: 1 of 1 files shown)


achen2401:
doing some code clean-up, will re-opening this when I am done.

PR #1197, opened 2017-07-31
add input validation for < & > characters, remove string from settimeout calls

achen2401:
add some frontend fixes to safeguard against xss data insert:

- add input validation check against < and > characters (only for now) 
- based on Content Security Policy documentation, https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP, remove string from setTimeout calls.

### DIFF
--- a/portal/static/css/eproms.css
+++ b/portal/static/css/eproms.css
@@ -1845,12 +1845,11 @@ table.profile-audit-log TH,
 }
 .error-message,
 .report-error-message {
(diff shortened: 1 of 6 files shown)


pbugni:
@ivan-c is correct, but this still seems like a reasonable front end change.  Any opposition to merging as is?

ivan-c:
@pbugni Nope, was only waiting on your feedback on whether or not to wait on the backend implementation

PR #1198, opened 2017-07-31
front end translation changes - using i18next library

achen2401:
*NOT ready to be merged yet*  pending Ivan's review
Ivan, I enclosed the following files just to detail the steps I took to extract JSON from the js files and then translate json into POT file:
**convert.js** (this contains code that translates JSON into POT)
**gulpfile.js** (see this one for the tasks I ran to extract text from JS files into JSON)
**package.json** is just that for my local nodeJS projects that details the dependencies needed to run my gulp tasks
**NOTE, I will delete the above files after you have reviewed since they are just sample files, and I have since make some changes in the js (I haven't re-run the tasks to re-generate).  The paths referenced in those files only referenced those on my mac**
translation.json file contains JSON that resulted from running i18next scanning tool to extract text from main.js and gil.js (NOTE the location of this JSON file can be changed, I just need to make the necessary adjustment in the code to look for where it is) 

Ivan, just to note that this can stand as is without JSON file containing translated text. No translation will take place, that is all.


achen2401:
will resolve conflict and re-create this PR

PR #1202, opened 2017-08-01
Enhance admin table

OptimusRhine:
https://www.pivotaltracker.com/story/show/149756859

* updated version of bootstrap-table from 1.10.1 to 1.11.1
* added bootstrap-table-export and bootstrap-table-filter-control extensions
* added table export functionality to /admin table
* added table filter functionality (for certain fields) to /admin table
* updated /admin API, to allow for `org_list` url param (to filter users in table based on org)
* added groundwork for `FILTER LIST BY SITE` button, including generifying the `initOrgsList()` function

### DIFF
--- a/portal/static/js/admin.js
+++ b/portal/static/js/admin.js
@@ -238,7 +238,7 @@ AdminTool.prototype.getUserOrgs = function() {
   if (this.userOrgs.length == 0) this.setUserOrgs(this.userId);
   return this.userOrgs;
 };
-AdminTool.prototype.initOrgsList = function(request_org_list) {
+AdminTool.prototype.initOrgsList = function(request_org_list, origin) {
     //set user orgs
     var self = this;
     self.setUserOrgs();
@@ -294,8 +294,8 @@ AdminTool.prototype.initOrgsList = function(request_org_list) {
                    if ($(this).is(":checked")) orgsList.push($(this).val());
                 });
                if (orgsList.length > 0) {
-                  location.replace("/patients/?org_list=" + orgsList.join(","));
-               } else location.replace("/patients");
+                  location.replace("/" + origin + "/?org_list=" + orgsList.join(","));
+               } else location.replace("/" + origin);
             });
         });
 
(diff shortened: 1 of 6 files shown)


portal/views/portal.py
```diff
@@ -595,10 +595,34 @@ def home():
 def admin():
     """user admin view function"""
     # can't do list comprehension in template - prepopulate a 'rolelist'
-    users = User.query.filter_by(deleted=None).all()
+
+    request_org_list = request.args.get('org_list', None)
+
+    if request_org_list:
+        org_list = set()
+
```
pbugni: This block looks familiar.  Are we doing this elsewhere?  Is a refactor of common code in order?

OptimusRhine:
@achen2401 **Not urgent**, but whenever you get a chance, can you assist a bit on this?

I've got the export/filter functions working fine, but I'm also trying to add a `FILTER LIST BY SITE` button (same as we current have on the /patients table).

I've laid the groundwork - the API endpoint now allows for an `org_list` param which it uses to filter the users it passes to the template, and the endpoint also passes the selected orgs to the template in `org_list`, so the template know's what to mark as selected - but I'm encountering issues when I start to actually try pulling in the admin.js and/or initializing an `AdminTool()`.

Does admin.js need reconfiguring to work with /admin (along with continuing to work with /patients )? Or is some other approach in order?

portal/views/portal.py
OptimusRhine: Well, we're doing something _similar_ in the /patients page. But that logic is different enough (because there, `org_list` also depends on the `current_user`'s orgs, the relative role levels, etc), that I don't think it's worth trying to break it into a shared method.

portal/static/js/admin.js
```diff
@@ -238,7 +238,7 @@ AdminTool.prototype.getUserOrgs = function() {
   if (this.userOrgs.length == 0) this.setUserOrgs(this.userId);
   return this.userOrgs;
 };
-AdminTool.prototype.initOrgsList = function(request_org_list) {
+AdminTool.prototype.initOrgsList = function(request_org_list, origin) {
```
OptimusRhine: Makes sense, I'll change that up

achen2401:
@OptimusRhine 
Adam, I will have a look.  It seems we can incorporate the methods (for AdminTool object) in admin.js for adding filtering by functionality.

achen2401:
@OptimusRhine 
Adam, I think Filter by Site functionality should work now.  I didn't have a lot of chances to test out my changes.  Let me know if you see any problem.  Thank you very much for making the changes.

OptimusRhine:
@achen2401 Thanks so much! I'll go ahead and test it out.

OptimusRhine:
Looks good! I think this should now be ready to merge.

PR #1203, opened 2017-08-01
fix patients performance issue

achen2401:
Fixes include - 
reducing the size of each ajax request
reducing the number of rows rendered each time (e.g. render each row as the data comes back now, instead of delayed rendering)

### DIFF
--- a/portal/static/js/admin.js
+++ b/portal/static/js/admin.js
@@ -5,9 +5,10 @@ function AdminTool (userId) {
   this.requestsCounter = 0;
   this.userId = userId;
   this.userOrgs = [];
-  this.initUserList = [];
   this.ajaxRequests = [];
   this.ajaxAborted = false;
+  this.arrData = {};
+  this.patientsIdList = [];
   OrgTool.call(this);
 };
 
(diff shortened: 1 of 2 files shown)


portal/static/js/admin.js
```diff
@@ -198,9 +181,9 @@ AdminTool.prototype.abortRequests = function(callback) {
 AdminTool.prototype.getUserIdArray = function(_userIds) {
   var us = "", ct = 0, arrUsers = [];
   if (!_userIds) {
-     _userIds = this.getInitUserList();
+     _userIds = this.getPatientsIdList();
   };
-  var max_ct = Math.max(__patients_list.length/10, 25);
+  var max_ct = Math.max(_userIds.length/10, 10);
```
pbugni: Is this what you want?  Won't a really large `_userIds` lead to a value greater than 10?

portal/static/js/admin.js
```diff
@@ -99,18 +122,15 @@ AdminTool.prototype.getData = function(requests, callback) {
...
+                              setTimeout(function() { self.fadeLoader(); }, 500);
                             }
                             else {
-                              self.fadeLoader();
-                              __index = 0;
                               if (callback) setTimeout(function() { callback.call(self);}, 300);
-                              setTimeout(function() { $("#adminTable tr[data-uniqueid]").show(); }, 300);
-                            }
+                            };
                         }).fail(function(xhr) {
```
pbugni: can we transmit the failure back to the server for logging?

PR #1205, opened 2017-08-02
Feature/dogpile assessment status

pbugni:
Implement our first dogpile cache for the expensive `overall_assessment_status` lookup per user.

Using a one hour timeout, the only invalidation being done is on a POST of assessment data, and then it only invalidates the named user.

Left in debugging log statements so it's easy to see when a cache miss occurs.  Pull up the /patients view - log entries like the following will only show up if the data for the patient is NOT in the cache:

```
DEBUG in assessment_status [.../portal/models/assessment_status.py:436]:
CACHE MISS: portal.models.assessment_status 198 221
```

### DIFF
--- a/portal/dogpile.py
+++ b/portal/dogpile.py
@@ -1,0 +1,19 @@
+"""Module for global dogpile cache regions"""
+from flask_dogpile_cache import make_region
+
+
+def key_mangler(key):
+    """Maintain unique redis keys for dogpile needs
+
+    Mangle the key with a prefix to prevent accidental cache collisions
+    """
+    return 'dogpile_cache:' + key
+
+
+# Region for caching with hourly updates
+hourly_cache = make_region(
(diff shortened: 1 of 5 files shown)


portal/dogpile.py
```diff
@@ -0,0 +1,19 @@
...
+    """
+    return 'dogpile_cache:' + key
+
+
+# Region for caching with hourly updates
+hourly_cache = make_region(
+    key_mangler=key_mangler).configure(
+          'dogpile.cache.redis',
+          expiration_time=3600,
+          arguments={'url': "127.0.0.1:11211"}
```
ivan-c: Is this actually for redis? I think port 11211 is memcached
Should probably be configurable like our other redis URLs...

pbugni:
I believe this was effectively merged already from the master branch.  Closing.

PR #1210, opened 2017-08-02
account creation fixes

achen2401:
initially addressing this story: https://www.pivotaltracker.com/story/show/149407045
- unable to reproduce the issue, tried quite a different combination of steps, but in any event, firm up validation of fields by display error messages when needed (e.g. trigger error message display on focusout event)
- combine common code used between patient and staff creation pages into reusable template for better management


portal/templates/patient_profile_create.html
```diff
@@ -104,330 +104,12 @@ <h4 class="tnth-headline left-indent-top">{{ _("New Patient") }}</h4>
     {%- endif -%}
 </form>
 <a id="redirectLink" href="">&nbsp;</a>
+{%- from "profile_macros.html" import accountCreationScript -%}
```
pbugni: Seems like the pattern is to do all the imports at the top of the file.  If there's reason to do it here, fine.

portal/templates/profile_macros.html
```diff
@@ -2232,3 +2235,342 @@ <h4 id="clinicalDataLoc" class="profile-item-title index-item">{% if person.id =
     <button id="updateProfile" type="submit" class="btn btn-lg btn-tnth-primary">{{ _("Save") }}</button>
     <span id="confirmMsg" class="text-muted tnth-hide small-text">&nbsp;&nbsp;<em>{{ _("Profile changes have been saved") }}</em></span>
 {%- endmacro %}
+{% macro accountCreationScript() -%}
+    <script>
+    var AccountCreationObj = function (roles) {
+        this.attempts = 0;
+        this.max_attempts = 3;
+        this.params = null;
+        this.roles = roles ? roles : [{'name': '{{ROLE.PATIENT}}'}, {'name': '{{ROLE.WRITE_ONLY}}'}];
```
pbugni: Not sure I love defaulting to PATIENT & WRITE_ONLY.  Shouldn't we just enforce passing the correct set of roles based on the context?

portal/templates/patient_profile_create.html
achen2401: no reason, Paul. I have moved the imports to the top.

portal/templates/profile_macros.html
achen2401: yeah, make sense, Paul. I have removed the defaulting code.

PR #1219, opened 2017-08-04
admin js cleanup

achen2401:
general admin js clean-up
- add  descriptive comments
- remove unused code (e.g. arrData is not used)
- remove unnecessary calls to function (e.g. remove fadeLoader call as it is already called on page load)
- add check for null values in parameters in methods


pbugni:
Thanks for the clean up and some comments!

PR #1223, opened 2017-08-07
Replace variables in incoming LR content

OptimusRhine:
https://www.pivotaltracker.com/story/show/148510441

* modified `VersionedResource` and `UnversionedResource` classes, to allow for optional `variables` param
  * when returning a resource's asset, now attempts formats asset with provided `variables` (or returns an error if a requested asset variable is missing from the provided `variables` dict)
* added unit tests

### DIFF
--- a/portal/models/app_text.py
+++ b/portal/models/app_text.py
@@ -275,7 +275,12 @@ class UnversionedResource(object):
     def asset(self):
         """Return asset if available else error message"""
         if self._asset:
-            return format_asset_attributes(self._asset, self.variables)
+            try:
+                return self._asset.format(**self.variables)
+            except KeyError, e:
+                self.error_msg = "Missing asset variable {}".format(e)
+                current_app.logger.error(self.error_msg +
+                                         ": {}".format(self.url))
         return self.error_msg
 
 
@@ -335,7 +340,12 @@ class VersionedResource(object):
     def asset(self):
         """Return asset if available else error message"""
         if self._asset:
-            return format_asset_attributes(self._asset, self.variables)
+            try:
+                return self._asset.format(**self.variables)
+            except KeyError, e:
+                self.error_msg = "Missing asset variable {}".format(e)
+                current_app.logger.error(self.error_msg +
+                                         ": {}".format(self.url))
         return self.error_msg
 
     def _permanent_url(self, generic_url, version):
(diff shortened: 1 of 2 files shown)


portal/models/app_text.py
```diff
@@ -233,7 +234,7 @@ def name_key(**kwargs):
 
 class UnversionedResource(object):
     "Like VersionedResource for non versioned URLs (typically local)"
-    def __init__(self, url, asset=None):
+    def __init__(self, url, asset=None, variables={}):
```
pbugni: Classic python gotcha.  You don't want to assign a default variable to something like an empty dict.  Even if the current implementation doesn't modify `variables` - code changes over time and these problems sneak in.

I'd suggest making that `variables = None` and then in the body, assign variables to {} if needed.  Here's more on the topic:  http://effbot.org/zone/default-values.htm

portal/models/app_text.py
```diff
@@ -249,6 +250,7 @@ def __init__(self, url, asset=None):
         """
         self._asset, self.error_msg, self.editor_url = None, None, None
         self.url = url
+        self.variables = variables
```
pbugni: A reasonable pattern:
`self.variables = variables if variables else {}`

portal/models/app_text.py
```diff
@@ -273,14 +275,14 @@ def __init__(self, url, asset=None):
...
-            return self._asset
+            return format_asset_attributes(self._asset, self.variables)
         return self.error_msg
 
 
 class VersionedResource(object):
     """Helper to manage versioned resource URLs (typically on Liferay)"""
 
-    def __init__(self, url):
+    def __init__(self, url, variables={}):
```
pbugni: Same issue with default args getting evaluated once.

portal/models/app_text.py
```diff
@@ -416,3 +419,14 @@ def app_text(name, *args):
         raise ValueError(
             "AppText with name '{}' defines more parameters "
             "than provided: `{}`".format(name, *args))
+
+
+def format_asset_attributes(asset, variables):
+    asset_out = ""
+    x = 0
+    for match in finditer(r"\${[^}]+}", asset):
```
pbugni: This feels like a reinvention of string format.  Can we not simply leverage that?  Something like:
```
asset_out = asset.format(**variables)
```

OptimusRhine: I wasn't sure if this would work, because LR is sending the variables in `${...}` format, whereas string format just looks for `{...}`, right? I guess we could just use string format and then remove all $'s, but that runs the risk of removing $'s unrelated to the variables (edge case, but it could happen)

OptimusRhine:
This now properly handles cases where we don't have the correct variables to match the incoming LR content (logs an error)

PR #1226, opened 2017-08-08
timeout duration fix for patients list

achen2401:
found these issues while testing:
- shorten the terribly long timeout (2 minutes) to 3 seconds for each ajax call for assessment status in patients list.  The long timeout was accidentally left behind while testing.
- shorten the timeout duration for aborted ajax call (so the callback can be called in due speed)
- fixed so that loading spinner will display before (instead of during) user is redirected to next page (e.g. profile or account creation page

### DIFF
--- a/portal/static/js/admin.js
+++ b/portal/static/js/admin.js
@@ -151,7 +151,7 @@ AdminTool.prototype.abortRequests = function(callback, showLoader) {
           try {
             if (parseInt(request.readyState) != 4) {
                 /*
-                 * aborting the request here quite immediately, instead of waiting for the
+                 * aborting the request here to quit immediately, instead of waiting for the
                  * maximum timeout specified
                  */
                 request.timeout = 100;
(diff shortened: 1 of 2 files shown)


portal/static/js/admin.js
```diff
@@ -141,25 +141,34 @@ AdminTool.prototype.updateData = function() {
...
    //NEED TO ABORT THE AJAX REQUESTS OTHERWISE CLICK EVENT IS DELAYED DUE TO NETWORK TIE-UP
    if (self.ajaxRequests.length > 0) {
       self.ajaxAborted = true;
       self.ajaxRequests.forEach(function(request, index, array) {
+
           try {
-            if (request.readyState != 4) {
+            if (parseInt(request.readyState) != 4) {
+                /*
+                 * aborting the request here quite immediately, instead of waiting for the
```
pbugni: You probably meant `quit immediately` ?

achen2401: thank you, Paul, for catching that.

PR #1236, opened 2017-08-10
Modify ScheduledJobs via API & site_persistence

OptimusRhine:
https://www.pivotaltracker.com/story/show/150117398

* added `from_json()` and `as_json()` methods to ScheduledJob model
  * `from_json()` looks up existing jobs by name. if job is found, updates relevant fields; otherwise, creates new job entry
* added uniqueness constraint to `ScheduledJob.name` field (includes migration)
* added robust API for ScheduledJobs, including endpoints to create, edit, view, and/or delete jobs
  * only accessible by admins
* added unit tests for all new ScheduledJob API endpoints (and therefore for new `from_json()` and `as_json()` methods as well)
* added ScheduledJob loading to `site_persistence.import()`, and ScheduledJob exporting to `site_persistence.export()`


portal/migrations/versions/2fdf0e8b8e08_.py
```diff
@@ -0,0 +1,26 @@
...
+revision = '2fdf0e8b8e08'
+down_revision = '7b4733a907f9'
+
+from alembic import op
+import sqlalchemy as sa
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.create_unique_constraint(None, 'scheduled_jobs', ['name'])
```
pbugni: IIRC this will create problems with the upgrade / downgrade loop, when the constraint isn't named.  See some of the other migrations where a unique constraint is added.  Test w/ `flask db downgrade && flask db upgrade`

portal/site_persistence.py
```diff
@@ -382,6 +387,24 @@ def update_intervention(intervention_json):
...
+        # ScheduledJobs shouldn't be order dependent, now is good.
+        jobs_seen = []
+        for j in objs_by_type['ScheduledJob']:
+            job = ScheduledJob.from_json(j)
+            db.session.add(job)
+            jobs_seen.append(j['name'])
+
+        # Delete any ScheduledJobs not named
+        if not keep_unmentioned:
+            jobs_seen.append("__test_celery__")
```
pbugni: Not sure I get the addition of `__test_celery__`.  Is that one you just want added in there, and expect would never be named in the persistence file?

portal/views/scheduled_job.py
```diff
@@ -30,3 +32,48 @@ def jobs_list():
...
+    return jsonify(job.as_json())
+
+
+@scheduled_job_api.route('/api/scheduled_jobs/<int:job_id>')
+@roles_required(ROLE.ADMIN)
+@oauth.require_oauth()
+def get_job(job_id):
+    job = ScheduledJob.query.get(job_id)
+    if not job:
+        abort(400, 'job ID not found')
```
pbugni: I believe we generally treat missing content as 404s.

portal/site_persistence.py
OptimusRhine: @pbugni yea, check out https://github.com/uwcirg/true_nth_usa_portal/blob/develop/portal/celery_worker.py#L29 .

I wanted a test task that wouldn't be visible to users, but would help us debug on the backend to confirm that celery beat's working as intended for each instance (the job basically just updates its own `last_runtime` every hour).

This job will get recreated by the celery worker when it starts, if it doesn't already exist. But I added this line in here to handle the (very unlikely) scenario that we seed an instance *while* celery's running, so that celery isn't attempting to update a nonexistent job.

PR #1243, opened 2017-08-14
Better ScheduledJob Task param logic

OptimusRhine:
* modified ScheduledJob -> celerybeat loading logic, so as to not assume that every referenced task will take `job_id` as a param
  * now, if `job_id` is present in the job's `kwargs`, it will automatically get populated with the correct ID value when loading into a celery task

### DIFF
--- a/portal/celery_worker.py
+++ b/portal/celery_worker.py
@@ -7,7 +7,7 @@ the flask background tasks to run.
 
 Launch in the same virtual environment via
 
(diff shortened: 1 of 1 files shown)


OptimusRhine:
On second thought, gonna keep `job_id` as an assumed param for ScheduledJob tasks, but abstract it out into a decorator. Closing this PR for now.

PR #804, opened 2017-04-18
Session Monitor

ivan-c:
* Listen and respond to session events (creation and expiration) and act on them
  * Session initialization
      1. Add user_id/session_id mapping
  * Session expiration
      1. Delete user_id/session_id mapping
      2. Perform logout event callback

[NB about gevent and WSGI](https://stackoverflow.com/questions/9692089/why-gevent-on-a-flask-app-with-apache-mod-wsgi-is-raising-notimplementederror/9710418#9710418)

PR #1246, opened 2017-08-15
Reporting dashboard v2

OptimusRhine:
https://www.pivotaltracker.com/story/show/150063689

* added Usage Statistics section to Reporting Dashboard
  * includes registrations and logins (incl. intervention breakdowns)
  * currently displays by day, month, year, and all time
    * eventual implementation will be a filterable graph - this is more a data POC
* added Encounter and user registration data collection to backend, for new Usage Stats section
* added 'Unspecified' option for orgs, for users with no orgs yet (different from users with 'none of the above' org)

### DIFF
--- a/portal/views/portal.py
+++ b/portal/views/portal.py
@@ -1046,8 +1046,7 @@ def get_reporting_counts():
                 counts['organizations'][org.name] += 1
         counts['registrations'].append(user.registered)
 
-    for enc in Encounter.query.filter_by(auth_method='password_authenticated',
-                                         status='finished'):
+    for enc in Encounter.query.filter_by(auth_method='password_authenticated'):
         st = enc.start_time
         counts['encounters']['all'].append(st)
         user = get_user(enc.user_id)
--- a/portal/views/portal.py
+++ b/portal/views/portal.py
@@ -1003,7 +1003,7 @@ def reporting_dashboard():
 
 @dogpile_cache.region('hourly')
 def get_reporting_counts():
-    """Cachable interface for expensive data lookups
+    """Cachable interface for expensive reporting data queries
 
     The following code is only run on a cache miss.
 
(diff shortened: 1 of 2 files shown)


portal/views/portal.py
```diff
@@ -1036,8 +1039,20 @@ def get_reporting_counts():
             counts['interventions'][interv.description] += 1
             if (any(doc.intervention == interv for doc in user.documents)):
                 counts['intervention_reports'][interv.description] += 1
-        for org in user.organizations:
-            counts['organizations'][org.name] += 1
+        if not user.organizations:
+            counts['organizations']['Unspecified'] += 1
```
pbugni: Hmm, we have "none of the above" in play elsewhere.  See organizations.id == 0.

portal/views/portal.py
```diff
@@ -1036,8 +1039,20 @@ def get_reporting_counts():
...
+        else:
+            for org in user.organizations:
+                counts['organizations'][org.name] += 1
+        counts['registrations'].append(user.registered)
+
+    for enc in Encounter.query.filter_by(auth_method='password_authenticated',
+                                         status='finished'):
```
pbugni: FWIW, we have a bug around `finished` encounters, which will get cleaned up when Ivan's [Session Monitor](https://github.com/uwcirg/true_nth_usa_portal/pull/804) solution is ready, that is, encounters only get marked finished if the browser session isn't killed.

You might ignore the `status` for now.

portal/views/portal.py
OptimusRhine: Yep, that 'none of the above' org is already taken into account when looking at the `User.organizations`. This is to handle cases where the user has not yet been assigned to _any_ orgs, _including_ the 'none of the above' org; see Justin's comments in [the PT story here](https://www.pivotaltracker.com/story/show/150063689).

pbugni: Gotcha.  Comments always help when such a situation isn't obvious - but now that I've read that, it is.

PR #1248, opened 2017-08-15
Cleanup env settings

ivan-c:
Harmonize postgres env var configuration- only use [PG* environmental variables](https://www.postgresql.org/docs/9.6/static/libpq-envars.html)


.travis.yml
```diff
@@ -15,7 +15,10 @@ env:
   global:
     - FLASK_APP="$(readlink -f manage.py)"
     - SQLALCHEMY_DATABASE_TEST_URI='postgresql://postgres:@localhost/portal_unit_tests'
```
pbugni: Might be a good idea to build `SQLALCHEMY_DATABASE_TEST_URI` from its constituent parts defined below, just to keep 'em in sync...

ivan-c: Done!

PR #1255, opened 2017-08-16
Add textual descriptions to reporting dashboard tables

OptimusRhine:
https://www.pivotaltracker.com/story/show/150063689

* added text descriptions for each table in /reporting
  * note that these descriptions are VERY rough and technical; will require input from Justin and/or Paul to rewrite in a more user-friendly format

### DIFF
--- a/portal/templates/reporting_dashboard.html
+++ b/portal/templates/reporting_dashboard.html
@@ -6,6 +6,9 @@
 
     <h2 class="tnth-headline">{{ _("Usage Statistics") }}</h2>
     <div>
+        <p>{{ _("Registrations are collected from the User.registration timestamps of all Users without the Test Role") }}</p>
+        <p>{{ _("Logins are collected from the start_time timestamps of Encounters whose auth_method='password_authenticated'") }}</p>
+        <p>{{ _("Intervention Logins are filtered based on the login Encounter's User's User.interventions") }}</p>
         <table id="usageStatsTable"
                class="table table-striped table-hover table-condensed"
                data-show-columns="true">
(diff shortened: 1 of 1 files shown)


OptimusRhine:
@pbugni @mcjustin Please take a look at these descriptions for the tables on the /reporting page. Some are very rough/technical, because I'm not sure how they translate on the user end (aka how the analysts from Mo, etc would understand these descriptions). I'd definitely appreciate any input on rewriting these in a more user-friendly fashion.

portal/templates/reporting_dashboard.html
```diff
@@ -6,6 +6,9 @@ <h3 class="tnth-headline">{{ _("Reporting Dashboard") }}</h3>
 
     <h2 class="tnth-headline">{{ _("Usage Statistics") }}</h2>
     <div>
+        <p>{{ _("Registrations are collected from the User.registration timestamps of all Users without the Test Role") }}</p>
```
pbugni: It looks like we're recording the time of account creation rather than registration in user.registered.  Those would be the same for an account created during registration, but incorrect for accounts created by staff.  Bug in `user.promote_to_registered`

portal/templates/reporting_dashboard.html
```diff
@@ -6,6 +6,9 @@ <h3 class="tnth-headline">{{ _("Reporting Dashboard") }}</h3>
...
+        <p>{{ _("Logins are collected from the start_time timestamps of Encounters whose auth_method='password_authenticated'") }}</p>
```
pbugni: Worth noting password_authenticated includes FB and Google auth - although whether those are available is externally configured.

portal/templates/reporting_dashboard.html
```diff
@@ -6,6 +6,9 @@ <h3 class="tnth-headline">{{ _("Reporting Dashboard") }}</h3>
...
+        <p>{{ _("Intervention Logins are filtered based on the login Encounter's User's User.interventions") }}</p>
         <table id="usageStatsTable"
```
pbugni: Worth noting user_interventions only exist for interventions that control their user list manually.  This won't count, for example, decision support or other interventions based on complicated strategy rules.

pbugni:
@mcjustin Any wordsmithing here, or are we good to go?  Review comments above should be considered.

portal/templates/reporting_dashboard.html
mcjustin: @OptimusRhine can you please add that clarification?
so instead of this:
<p>{{ _("Intervention counts are tallied from User.interventions (e.g. a User with both the 'Care Plan' and 'Symptom Tracker' interventions, would add 1 to both Interventions' counts)") }}</p> 
this:
<p>{{ _("Intervention counts only apply to those interventions that control their subject list manually (eg Sexual Recovery, Care Plan, Community of Wellness). They are tallied from User.interventions (e.g. a User with both the 'Care Plan' and 'Community of Wellness' interventions, would add 1 to both Interventions' counts)") }}</p> 

In the next iteration we'll probably need to move P3P, Symptom Tracker and WiserCare to their own section.

mcjustin: @OptimusRhine once you've made that change we can merge it in - I'm certain it will need more, but I'd like to see these changes in context.

OptimusRhine: @mcjustin OK, I've made that change. @pbugni sounds like we're good to merge.

portal/templates/reporting_dashboard.html
pbugni: @mcjustin did you file this one already?  Not critical, but a pretty easy fix and the current info would be misleading.

OptimusRhine: @mcjustin can we get a PT story for this issue (bug: `user.registration` timestamp records account creation time, not user registration time)? Once that's filed, we can merge these changes.

OptimusRhine: Looks like we've got a story for this issue (https://www.pivotaltracker.com/story/show/150502476). @pbugni I think we can merge this PR now.

PR #1257, opened 2017-08-17
Cleanup deploy script

ivan-c:
* Remove partial conditional deploy logic (all steps run with each invocation now)
* Exit on first error
* Remove verbose/force options

### DIFF
--- a/bin/deploy.sh
+++ b/bin/deploy.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/bash -e
 
 usage() {
     echo "$0 - Simple script to make deployments of fresh code a one command operation" >&2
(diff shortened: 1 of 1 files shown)


bin/deploy.sh
```diff
@@ -81,78 +71,21 @@ if [[ -z $BRANCH ]]; then
...
+python setup.py egg_info --quiet
 
 # Restart apache if application is served by apache
 if [[ "${GIT_WORK_TREE}" == "/srv/www/"* ]]; then
-    if [[ $VERBOSE ]]; then
-        echo "Restarting apache"
-    fi
+    echo "Restarting services"
     sudo service apache2 restart
+    sudo service celeryd restart
```
pbugni: ohh - sticky.  celeryd is now named per vhost, and we should also restart the celerybeat-${VHOST} script.  I don't believe we have sudo access for user truenth to restart either of these.

ivan-c: We should be able to get that from the current path pretty easily. I'll circle back once Adam finalizes the celerybeat work

PR #1268, opened 2017-08-21
MailResource class, implemented with user invite email

OptimusRhine:
https://www.pivotaltracker.com/story/show/150369091

* added new MailResource class in app_text module
  * instead of `.asset`, has `.subject` and `.body`
    * both fields have variable fields replaced
    * `.body` field is rendered in a specific way, as to not destroy html formatting
* added UserInviteEmail_ATMA class for 'profileSendEmail invite email' in app_text module
* in `render_template()` calls of 'profile.html', create and pass through MailResource object for invite email info
* properly pass through and use info from invite email MailResource object in profile template (incl. 'profile_macros.html')

Note: This will also require new app_text entry for ('profileSendEmail invite email', '{config[LR_ORIGIN]}/c/portal/truenth/asset/mail?version=latest&uuid=32a6713e-5051-8079-0ed3-0a7eeb607c03') on both persistence files:
TrueNTH - https://github.com/uwcirg/TrueNTH-USA-site-config/pull/48
ePROMS - https://github.com/uwcirg/ePROMs-site-config/pull/53

### DIFF
--- a/portal/models/app_text.py
+++ b/portal/models/app_text.py
@@ -422,8 +422,8 @@ class MailResource(object):
         self.variables = variables or {}
         try:
             response = requests.get(url)
-            self._subject = response.json().get('subject')
-            self._body = response.json().get('body')
+            self._subject = str(response.json().get('subject'))
+            self._body = str(response.json().get('body'))
             self.url = self._permanent_url(
                 generic_url=url, version=response.json().get('version'))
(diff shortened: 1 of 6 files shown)


OptimusRhine:
@pbugni OK - Added a test (and in the process, improved how we're handling string conversion with the `MailResponse.body` field)

PR #1270, opened 2017-08-21
Use pep8 complaint operators for SQL null checks.

pbugni:
SQLAlchemy uses operator overloading in filter statements, but one can't
overload the is operator.

As of SQLAlechemy 0.7.9 the is_() and isnot() functions became
available.
http://docs.sqlalchemy.org/en/rel_1_0/core/sqlelement.html#sqlalchemy.sql.operators.ColumnOperators.is_

### DIFF
--- a/portal/models/encounter.py
+++ b/portal/models/encounter.py
@@ -126,7 +126,7 @@ def finish_encounter(user):
     now = datetime.utcnow()
     # Look for any stale encounters needing to be closed out.
     query = Encounter.query.filter(Encounter.user_id==user.id).filter(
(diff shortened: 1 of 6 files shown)


pbugni:
@ivan-c this resolves my primary concern with running a tool like `autopep8`.  Glad to have found these operators as I've always found SQL null checks fragile.

ivan-c:
@pep8speaks pep8ify

ivan-c:
Bleh, [known issue](https://github.com/OrkoHunter/pep8speaks/issues/54)- fixed it myself...

PR #1264, opened 2017-08-18
profile title changes, css fixes

achen2401:
patient profile title changes, address this story: 
https://www.pivotaltracker.com/story/show/150118632
for patient profile title:
- use name if available, otherwise use email
- css fixes
- fix email saving (for no email)

PR #1272, opened 2017-08-22
Revert "remove trans tag from email"

ivan-c:
Reverts last commit of #1264

The `user_email` template var is only in scope inside the trans block. 

It isn't strictly required to include the user's email as a template variable, but doing so gives the translator a little more context as to what the text *Profile for* refers to. 

Ideally when we mark strings for translation they should represent (as closely as possible) a complete sentence, thought or element. This is a little more of a grey area since it's just a heading, but the id/email are implied to be part of the heading and should be included if possible.

### DIFF
--- a/portal/templates/profile.html
+++ b/portal/templates/profile.html
@@ -40,7 +40,7 @@
                     {%- else -%}
                         {%- if current_user.id != user.id -%}
                             {%- if user.email -%}
-                            {% trans user_email=user.email %}Profile for {% endtrans %} {{ user_email }}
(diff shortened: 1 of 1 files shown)


ivan-c:
@achen2401 @pbugni Let me know if you have any questions, thanks for tolerating my pedantic nagging :)

pbugni:
@ivan-c Keep it coming - you haven't achieved pedantic yet :)

I may have mislead @achen2401 here - my point was that we would never want to translate a users {name, email, address}, but I get your point - the translator isn't translating said text, but the tags are helpful for context, and possibly a change in word order.

PR #1274, opened 2017-08-22
Changing /patients column label "Truenth ID" to "TrueNTH ID"

mcjustin:
@achen2401 if you could just make sure I'm not stomping on a key or something...


achen2401:
@mcjustin Justin, looks good.  I always forget that NTH should be capitalized!

PR #1275, opened 2017-08-22
Extend QuestionnaireBanks for SymptomTracker needs

OptimusRhine:
https://www.pivotaltracker.com/story/show/150348469

* updated QuestionnaireBank model
  * add `intervention_id` field
  * allow `NULL` values for `organization_id` field
  * add new CheckConstraint, where `intervention_id` and `organization_id` cannot both be NULL, nor can they both have values (aka at least one is required, but they're mutually exclusive)
* added migration for new fields/constraints
* updated `qbs_for_user()` logic, to pull the QBs associated with the user's interventions (via `User.interventions`) as well
* added tests for intervention-based QBs
* fixed minor issue with existing org QB lookup test


OptimusRhine:
Er, don't merge yet - need to fix a migration issue

portal/migrations/versions/26fb7947cd1e_.py
```diff
@@ -0,0 +1,43 @@
...
+    op.alter_column('questionnaire_banks', 'organization_id',
+                    existing_type=sa.INTEGER(),
+                    nullable=True)
+    op.create_foreign_key('questionnaire_banks_intervention_id_fkey', 'questionnaire_banks',
+                          'interventions', ['intervention_id'], ['id'])
+    op.create_check_constraint(
+               'ck_qb_intv_org_mutual_exclusion',
+               'questionnaire_banks',
+               'NOT(organization_id IS NULL AND intervention_id IS NULL) '
+               'AND NOT(organization_id IS NOT NULL AND intervention_id IS NOT NULL)')
```
pbugni: bonus points!

PR #1280, opened 2017-08-23
fix consent withdrawn verbiage, organization check in profile fix

achen2401:
per conversation with Justin related to this story:  https://www.pivotaltracker.com/story/show/150118994
- populate consent withdrawn text via css as it is different between EPROMS and TRUENTH

bug fix - receiving error when trying to load profile for patient that has no affiliated org - add check for existence of user's first org

this is for this upcoming release


achen2401:
need to make further fix on this, closing this for now

PR #1282, opened 2017-08-23
fix patients list swipe event in mobile devices

achen2401:
found this issue when testing in mobile devices:
when swiping rows in patient list right to see hidden columns - redirection to profile page occurred

- made the fix so redirection only occurred on touch event of rows in mobile devices

this is a bug - should be included in this release


achen2401:
non sequitor - if we get the assessment status from backend - close this for now

PR #1283, opened 2017-08-23
Feature/patients list assessment status backend changes

achen2401:
related to this story: https://www.pivotaltracker.com/story/show/149615045
- retrieve user's assessment status in backend
- remove ajax call that retrieve assessment status
- css fixes

### DIFF
--- a/portal/views/patients.py
+++ b/portal/views/patients.py
@@ -99,7 +99,7 @@ def patients_root():
                      UserIntervention.intervention_id.in_(ui_list)))
         patients = patients.union(ui_patients)
 
-    #get assessment status only if it is needed as specified by config
+    # get assessment status only if it is needed as specified by config
     if 'status' in  current_app.config.get('PATIENT_LIST_ADDL_FIELDS'):
         patient_list = []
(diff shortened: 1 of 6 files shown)


portal/views/patients.py
```diff
@@ -99,8 +99,16 @@ def patients_root():
...
         patients = patients.union(ui_patients)
 
+    # get assessment status only if it is needed as specified by config
+    if 'status' in current_app.config.get('PATIENT_LIST_ADDL_FIELDS'):
+        patient_list = []
+        for patient in patients:
+            patient.assessment_status = overall_assessment_status(patient.id)
+            patient_list.append(patient)
+        patients = patient_list
+
```
pbugni: I'm not sure I grok the pattern of creating a secondary list `patient_list` and then copying that back over the top of `patients` after iteration.  I don't believe you need to use `patient_list` here at all - unless I'm missing something?

pbugni: I'm wrong - the iteration over patients is returning a copy and won't modify the original.  Nevermind.

PR #1284, opened 2017-08-23
Feature/communication model

pbugni:
Probably missing a few details, but this should implement the lion's share of automating the sending of email reminders from site_persistence values for CommunicationRequests.

NB - Only the task of queueing messages is scheduled at this time - the actual send is pending decisions on how to handle existing users.


portal/models/app_text.py
```diff
@@ -461,6 +461,7 @@ def subject(self):
                 self.error_msg = "Missing subject variable {}".format(e)
                 current_app.logger.error(self.error_msg +
                                          ": {}".format(self.url))
+                raise
         return self.error_msg
```
pbugni: Ah - great point.  Yes, I think the interface should be strict, so we can avoid unsightly messages landing in inboxes everywhere!

pbugni:
Putting this feature on hold till post August release.  Closing PR for now.

PR #1286, opened 2017-08-24
Feature/frontend i18next translation changes

achen2401:
address this story:  https://www.pivotaltracker.com/story/show/150506204
first attempt at incorporating the i18next library
details on how I arrived at the translation.json/pot files are specified in the gulpfile (i.e. i18next_gulp.js) that I created and ran when testing.
I tried to tag everything where applicable.
debugging mode is enabled in development so can see which translation key/text is targeted
feedback on this is much appreciated.

**not for this release, but the next one**


portal/i18next_gulpfile.js
```diff
@@ -0,0 +1,186 @@
...
+    console.log("converting JSON to POT...");
+    del(['translations/js/src/translation.pot']);
+    i18nextConv.i18nextToPot('en', fs.readFileSync(translationSourceDir+nameSpace+".json"), options).then(save(translationSourceDir+nameSpace+".pot"));
+
+});
+
+/*
+ * combine newly created pot file to existing messages.pot file ???
+ * do we need this step??
+ */
```
OptimusRhine: @ivan-c Do we want to merge the POT files here? It might be better to keep them separate, right?

I guess the main question is, how does this process fit in with our existing workflow? Should they be integrated (aka we'd want run this and upload the resulting POT file to smartling as part of our `manage.translation_upload()` process), or separate (in which case I'd imagine we'd want to do a separate round of uploading/downloading here)?

portal/i18next_gulpfile.js
```diff
@@ -0,0 +1,186 @@
...
+gulp.task('i18nextConvertPOToJSON', function() {
+  console.log("converting po to json ...")
+  const options = {/* you options here */}
+
+  del(['static/files/**/translation.json']);
+
+   /*
+    * translating po file to json for supported languages
+    */
```
OptimusRhine: Do we need to rely on a maintained list of supported languages here? Or is it possible to iterate through each dir in `portal/translations/`, and if a PO file is found in that lang's subdir (or more accurately, in `portal/translations/<lang_name>/LC_MESSAGES`, we generate a corresponding JSON file?

I think the goal is to just be able to maintain our list of "supported" languages in smartling, and have the upload/download steps work for all languages found (for uploading, we just need one POT file for all the languages, and for downloading, we get a zip that contains PO files for all the supported translation languages found in the smartling project). At least, that's how we designed the existing smartling integration stuff.

achen2401: @OptimusRhine  Thank you, Adam, this makes sense.  I will make the fix.

portal/i18next_gulpfile.js
achen2401: @OptimusRhine Adam, that is also my question too...yeah wait and see what Ivan says on this.

ivan-c: I'm tempted to keep them separate, (maybe `portal/translations/frontend.pot`?) at least for now, if only because it would be more work to separate them later if we changed our minds.

We'll still want to merge all the PO files for the same language, but I think that makes sense to do after downloading them from Smartling.

Thoughts?

portal/i18next_gulpfile.js
```diff
@@ -0,0 +1,186 @@
...
+  });
+});
+
+/*
+ * concating all necessary i18next JS files into one
+ * don't run this unless files have been changed
+ */
+gulp.task('i18nextConcatScripts', function() {
+  console.log("concat i18next scripts ...")
+  var i18nextMain = request('https://unpkg.com/i18next/i18next.js')
```
ivan-c: Can these libraries be used directly from unpkg.com? unpkg appears to be a CDN so they be able to better serve users in more remote locations.

portal/i18next_gulpfile.js
achen2401: separate sounds good.  How you are guys handling this in CPRO?  I am curious..is it only a separate pot file for front end as well?

portal/i18next_gulpfile.js
```diff
@@ -0,0 +1,186 @@
...
+    .pipe(source('i18nextMain.js'));
+  var i18nextXHRBackend = request('https://unpkg.com/i18next-xhr-backend/i18nextXHRBackend.js')
```
ivan-c: Would probably be a good idea to [pin the version number](https://before-you-ship.18f.gov/infrastructure/pinning-dependencies/) (eg https://unpkg.com/i18next-xhr-backend@1.4.2/i18nextXHRBackend.js). More difficult to maintain for updates, but that way we know for sure what versions work.

portal/i18next_gulpfile.js
```diff
@@ -0,0 +1,186 @@
...
+    .pipe(source('i18nextXHRBackend.js'));
+  var i18nextLnDetection = request('https://unpkg.com/i18next-browser-languagedetector/i18nextBrowserLanguageDetector.js')
```
ivan-c: Would probably be a good idea to [pin the version number](https://before-you-ship.18f.gov/infrastructure/pinning-dependencies/) (eg https://unpkg.com/i18next-xhr-backend@1.4.2/i18nextXHRBackend.js). More difficult to maintain for updates, but that way we know for sure what versions work.

portal/package.json
```diff
@@ -0,0 +1,31 @@
+{
+  "name": "truenth",
+  "version": "0.0.1",
+  "description": "fills the front-end needs including incorporating i18next for frontend translation needs",
+  "main": "main.js",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "author": "Amy",
+  "license": "ISC",
```
ivan-c: I think we're using [BSD as our license](https://github.com/uwcirg/true_nth_usa_portal/blob/develop/LICENSE)

portal/i18next_gulpfile.js
achen2401: yup, will remove this task - it was meant for me to test.  The libraries are loaded directly unpkg/cdn (see changes in layout.html)

portal/i18next_gulpfile.js
achen2401: yeah, will remove this task.  was actually meant for me to test.

portal/i18next_gulpfile.js
achen2401: k

portal/package.json
achen2401: k will change

portal/static/js/main.js
```diff
@@ -256,55 +257,55 @@ var fillViews = {
...
         content = $("#patMeta input[name='pca_localized']:checked").closest("label").text();
-        if (hasValue(content)) $("#pca_localized_view").html("<div>" + content + "</div>");
-        else $("#pca_localized_view").html("<p class='text-muted'>No answer provided</p>");
+        if (hasValue(content)) $("#pca_localized_view").html("<div>" + i18next.t(content) + "</div>");
+        else $("#pca_localized_view").html("<p class='text-muted'>" + __NOT_PROVIDED_TEXT + "</p>");
     },
     "deceased": function() {
         if ($("#boolDeath").is(":checked")) {
-            $("#boolDeath_view").text("Patient has deceased");
+            $("#boolDeath_view").text(i18next.t("Patient has deceased"));
```
ivan-c: I *think* that should read "Patient *is* deceased"

portal/static/js/main.js
```diff
@@ -586,9 +587,18 @@ var fillContent = {
...
                  please see portal.css (for Truenth) and eproms.css (for EPROMs) for detail
              ********/
-            var headerArray = ['Organization', '<span class="eproms-consent-status-header">Consent Status</span><span class="truenth-consent-status-header">Status</span>',
-                                '<span class="agreement">Agreement</span>',
-                                '<span class="eproms-consent-date-header">' + (typeof CONSENT_DATE_LABEL != "undefined" ? CONSENT_DATE_LABEL : "Consent Date") + '</span><span class="truenth-consent-date-header">Registration Date</span> <span class="gmt">(GMT)</span>'];
+            var headerEnum = {"consentStatus": i18next.t("Consent Status"),
+                              "status": i18next.t("Status"),
+                              "agreement": i18next.t("Agreement"),
+                              "consentDate": i18next.t("Consent Date"),
+                              "registrationDate": i18next.t("Regiatration Date"),
```
ivan-c: Typo :)

portal/static/js/main.js
```diff
@@ -1669,24 +1680,24 @@ OrgTool.prototype.getDefaultModal = function(o) {
                 '<div class="modal-content">' +
                 '<div class="modal-header">' +
                 '<button type="button" class="close" data-dismiss="modal" aria-label="Close"><span aria-hidden="true">&times;</span></button>' +
-                '<h4 class="modal-title">Consent to share information</h4>' +
+                '<h4 class="modal-title">' + i18next.t("Consent to share information") + '</h4>' +
                 '</div>' +
                 '<div class="modal-body">' +
                 '<h4>Terms</h4>' +
-                '<p>I consent to sharing information with the <span class="consent-clinic-name">' + orgName + '.</span></p>' +
+                '<p>' + i18next.t("I consent to sharing information with the ") + '<span class="consent-clinic-name">' + orgName + '.</span></p>' +
```
ivan-c: Can you try i18next interpolation here? Some languages may have the org name in the beginning of the sentence.

portal/static/js/main.js
```diff
@@ -1669,24 +1680,24 @@ OrgTool.prototype.getDefaultModal = function(o) {
...
-                '<input type="radio" name="toConsent" id="' + orgId + '_consent_no" data-org="' + orgId + '"  value="no"/>No</label>' +
+                '<input type="radio" name="toConsent" id="' + orgId + '_consent_no" data-org="' + orgId + '"  value="no"/>' + i18next.t("No") + '</label>' +
                 '</div>' +
                 '<div id="' + orgId + '_consentAgreementMessage" class="error-message"></div>' +
                 '</div>' +
                 '<br/>' +
                 '<div class="modal-footer" >' +
                 '<div id="' + orgId + '_loader" class="loading-message-indicator"><i class="fa fa-spinner fa-spin fa-2x"></i></div>' +
-                '<button type="button" class="btn btn-default btn-consent-close" data-org="' + orgId + '" data-dismiss="modal" aria-label="Close">Close</button>' +
+                '<button type="button" class="btn btn-default btn-consent-close" data-org="' + orgId + '" data-dismiss="modal" aria-label="Close">' + i18next.t("Close") + '</button>' +
```
ivan-c: `aria-label` should probably also be translated

portal/static/js/main.js
```diff
@@ -2256,7 +2268,7 @@ var tnthAjax = {
                     }).fail(function(xhr) {
                         //console.log("request to delete consent failed.");
                         //console.log(xhr.responseText)
-                        var errorMessage = "Server error occurred removing consent: " + xhr.responseText;
+                        var errorMessage = i18next.t("Server error occurred removing consent: " + xhr.responseText);
```
ivan-c: I think you may need to use interpolation here

portal/i18next_gulpfile.js
ivan-c: We've done it in a hacky way- [one JS file per language](https://gitlab.cirg.washington.edu/svn/dhair2/tree/master/app/webroot/js/localization), would be nice to switch to i18next there too.

achen2401:
@ivan-c  Ivan, I made changes based on most of your comments.  A few things I found when I was making the changes...
Regarding i18next interpolation, I found that in order for that to do, the key and the corresponding translated text will need to be in the source json file already.
Example code in question:
original code in main.js:
"consent to share information with " + orgName
tried changing it to:
i18next.t("consent_key", {orgname: orgName});
for the key to be translated, I needed to add this to the source json for this to work:
{"consent_key": "consent to share information with {{orgname}}"}
hmm..that means we have to maintain some of these key and value pairs?
So my hack was to separate out the text to be translated:
i18next.t("consent to share information with: " )
i18next.t(orgName)
so that each will have a corresponding key and its associated translated text in the source file.
And I am not sure if some of the text generated at runtime will be translated - example will be the error text that are generated as a result of a failed ajax call - I did look into the on missing key event of i18next (the event that runs when a key/value pair is missing from source json file) - it seems that you can run custom code in it to capture some of these missing key/value pairs - but what to do with those is my question - do we store them somewhere and upload those to smartling as well?

achen2401:
@ivan-c @OptimusRhine @pbugni 
Closing this for now - will re-open after the release

PR #1291, opened 2017-08-28
Test for TrueNTH ST AccessStrategy

OptimusRhine:
https://www.pivotaltracker.com/story/show/150532380

* added new unit test for TrueNTH's SymptomTracker (aka self_management) AccessStrategy
  * includes new Biopsy check

### DIFF
--- a/tests/test_intervention.py
+++ b/tests/test_intervention.py
@@ -818,8 +818,8 @@ class TestIntervention(TestCase):
 
     def test_truenth_st_conditions(self):
         # Test the list of conditions expected for SymptomTracker in truenth
-        ds_sm = INTERVENTION.SELF_MANAGEMENT
-        ds_sm.public_access = False
+        sm = INTERVENTION.SELF_MANAGEMENT
+        sm.public_access = False
         user = self.test_user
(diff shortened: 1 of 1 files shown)


OptimusRhine:
In conjunction with new ST Biopsy condition:
https://github.com/uwcirg/TrueNTH-USA-site-config/pull/50

tests/test_intervention.py
```diff
@@ -816,6 +816,101 @@ def test_eproms_p3p_conditions(self):
         user, ds_p3p = map(db.session.merge, (user, ds_p3p))
         self.assertFalse(ds_p3p.display_for_user(user).access)
 
+    def test_truenth_st_conditions(self):
+        # Test the list of conditions expected for SymptomTracker in truenth
+        ds_sm = INTERVENTION.SELF_MANAGEMENT
```
pbugni: *really* minor, but the `ds_*` pattern belongs to decision_support_{p3p, wisercare}.  Probably just want to use `sm` here.

OptimusRhine: Ah right, that makes sense

PR #1294, opened 2017-08-29
Invalidate assessment cache for user on consent post.

pbugni:


pbugni:
@achen2401 noticed we failed to catch manual updates to a user's consent date, and continued to show the cached value for assessment status.  This confront that.

PR #1296, opened 2017-08-29
Add variable list to MailResource

OptimusRhine:
https://www.pivotaltracker.com/story/show/148510441

* added `variable_list` to `MailResource`, which provides a list of all requested variables from the LR content
* updated `MailResource` tests


OptimusRhine:
(rebased)

PR #1299, opened 2017-08-30
display error message if orgs failed to load in account creation page

achen2401:
When the ajax call failed to load organizations in account creation page, the UI remains blank - there is no indication that a failure has occurred.
The fix is to display error message to user if that does occur.


pbugni:
@achen2401 should this be merged into the release branch?

Second question - can we invoke a retry w/o asking the user to reload the page?

achen2401:
@pbugni Paul, invoking a second try makes sense.  Will make that fix.   I think these changes can wait until the next release.

achen2401:
per standup - will re-create this based off of the release branch

PR #1301, opened 2017-08-30
Strict matching on QuestionnaireStatus filter in /patients

OptimusRhine:
* toggled strict search on the data filter for the QuestionnaireStatus column of the /patients table (so that selecting 'Completed', for example, won't also show rows with 'Partially Completed' status)


pbugni:
@OptimusRhine  Please reopen to merge into branch `release/v17.8.29`

OptimusRhine:
Ah right, gotcha.

PR #1303, opened 2017-08-30
using image optim to compress some large images

achen2401:
address this story:
https://www.pivotaltracker.com/story/show/150686621

compress large images files for performance - saved about 7% per file
I checked that the large svg file - profile-image@1x.svg - is not currently used by any page


pbugni:
Thanks @achen2401 - is it worth running through the rest of the files?

Lets delete the ones that aren't being used at all.

achen2401:
Paul - some of the images were at the optimal size so didn't get compressed - I did delete some unused images files.

PR #1309, opened 2017-08-30
Add configurable profiler to assist in tracking down bottlenecks

pbugni:
Disabled unless configuration is added - shows significant detail on *every* request in the logs.

### DIFF
--- a/portal/app.py
+++ b/portal/app.py
@@ -7,6 +7,7 @@ import sys
 import requests_cache
 from flask import Flask
 import redis
+from werkzeug.contrib.profiler import ProfilerMiddleware
 
 from .audit import configure_audit_log
 from .config import DefaultConfig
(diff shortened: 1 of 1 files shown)


ivan-c:
Neat! As I always ask: would it make sense to auto-enable this conditional on `SYSTEM_TYPE`? 

Are there any performance implications from having it enabled?

pbugni:
> would it make sense to auto-enable this conditional on SYSTEM_TYPE?

No, I see this as a toggle you'd turn on when investigating performance issues.  Too much bulk in the logs on *every* request.

> Are there any performance implications from having it enabled?

I'm sure there is a minor hit, mostly from the logging.  Profiling certainly would cost a bit as well.  Again, I only expect to use when debugging performance issues.

PR #1312, opened 2017-08-31
Use new email templates for CRV & IRONMAN

OptimusRhine:
* add logic to pull specific invite email templates from LR if user's top org is CRV or IRONMAN
  * see https://github.com/uwcirg/ePROMs-site-config/pull/55 for specific LR url vals
* added and populated MailResource variables for new template


pbugni:
Well done.  Thank you.

PR #1326, opened 2017-09-05
Simple lookup view for LR uuids

pbugni:
@mcjustin I added this simple view function to lookup the article ID from the configured LR.  Not protected, perhaps it should be - but seems pretty harmless.


pbugni:
Looks like Victor added an info page in Liferay to lookup the UUID directly on a piece of content.  Closing this PR as it's no longer needed.

PR #1331, opened 2017-09-06
consolidate script codes, load static scripts after html rendering to…

achen2401:
a few changes that may help speeding up loading of pages:
- consolidate static js code (moving static/boostrap.rowlink.js and static/validator.min.js to main.js) so only loading one static file
- move loading of static JS files to bottom of html template since it is rendering-blocking
- making loading of external export and filter table plugin js files optional (since they are only required in patients list and admin pages)
- removed unused js code

### DIFF
--- a/portal/static/js/main.js
+++ b/portal/static/js/main.js
@@ -18,7 +18,6 @@ function equalHeightBoxes(passClass) {
         $('.'+passClass).height(maxHeight);
     }
 }
-
 // Return an XHR without XHR header so  it doesn't need to be explicitly allowed with CORS
 function xhr_function(){
     // Get new xhr object using default factory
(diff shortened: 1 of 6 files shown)


pbugni:
Thank you @achen2401 .  I don't mind merging this, but before I do, I see for example, on the /patients page, YSlow tells me we're still pulling 10 distinct JS files.  Is there more room for improvement (or should that wait for another PR)?

achen2401:
@pbugni 
Paul, I did think about consolidating them into one static file but after discussing this with Ivan, I was dissuaded from doing that. Those external JS libraries (e.g. bootstrap and font awesome libraries) were being requested from CDN, which is now using HTTP2 protocol that will load the assets in parallel, therefore allowing a page requesting those resources to load faster.  Ivan referred me to a helpful article:
https://blog.cloudflare.com/http-2-for-web-developers/
The static JS/CSS files (e.g. main.js, validator.js, rowlink.js) that we have in portal are served via HTTP1.1 protocol, which is rendering blocking.  That is why I consolidated those.

PR #1334, opened 2017-09-06
resized actual footer images, remove width/height attributes for the …

achen2401:
another fix to help performance - 
resized  the actual sizes of footer images and removed size attributes (i.e. height and width) of those images to avoid re-painting of page at runtime.


pbugni:
Nice!

PR #1342, opened 2017-09-08
Custom export filenames

OptimusRhine:
* modified options to use custom filenames for Staff, Admin, and Patient list exports
  * custom filename is `<table_object>List_<datetimestamp` (e.g. `StaffList_08092017150707.csv`)

### DIFF
--- a/portal/static/js/main.js
+++ b/portal/static/js/main.js
@@ -3977,7 +3977,7 @@ function containHtmlTags(text) {
 };
 function __getExportFileName(prefix) {
     var d = new Date();
-    return (prefix?prefix:"exportList__")+("00" + d.getDate()).slice(-2)+("00" + (d.getMonth() + 1)).slice(-2)+d.getFullYear()+("00" + d.getHours()).slice(-2)+ ("00" + d.getMinutes()).slice(-2)+ ("00" + d.getSeconds()).slice(-2);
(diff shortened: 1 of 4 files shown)


OptimusRhine:
@achen2401 Would you mind taking a look at this, when you get a chance? My `exportOptions` changes worked perfectly for the Admin and Staff tables, but for some reason the same code isn't working for the Patient table (it's just using the default name)...

achen2401:
@OptimusRhine 
Adam, I figured out why the export file naming wasn't working for patients list - moving the initiation of table outside of document ready block seems to resolve it - I am still not sure why it wasn't working if it was in that block yet.. also I hope it is okay that I moved the code for naming the export file name to one function call?  Let me know if you find any issue with my changes?

OptimusRhine:
@achen2401 Looks good to me, and also works for me all on the tables now - thanks! My only question would be, was it intentional to use two underscores in the filename, instead of one (e.g. `StaffList__` instead of `StaffList_`)? I think one might look cleaner, but I'm fine with two if that was an intentional change.

achen2401:
@OptimusRhine 
Adam, I think I really meant to just use one underscore there, sorry for the confusion! I agree, one underscore is cleaner.

OptimusRhine:
@achen2401 Ah ok cool, updated it to just use one then. Thanks for your help here!

PR #1349, opened 2017-09-12
Include frontend translation files in Smartling processes

OptimusRhine:
https://www.pivotaltracker.com/n/projects/1225464/stories/150506204/comments/178799487

* modified Smartling upload/download functions to be file-agnostic, cleaned up code
* added `frontend.pot` upload step, and `frontend.po` download step
* ran full test, which thus generated a new `messages.pot` file, and new PO/MO files (both `messages` and `frontend`) for fr_CA


OptimusRhine:
(Please don't merge until @ivan-c 's had a chance to review)

OptimusRhine:
@achen2401 Right now, we're bringing the entries from the downloaded PO file into the existing PO file. But we use the `polib` library to cycle through the entries in the downloaded file, try to look up each entry in the existing file, and overwrite with the new downloaded translations if an existing entry is found (with new entries appended to the end of the existing files). So the file should only grow with new entries.

This design means if a translation/row is somehow deleted from Smartling, we don't lose it in our PO files. Also, if we decide to add entries outside of Smartling (e.g. manually check them in), they won't get lost on the next Smartling download. At least, I _think_ those are the reasons why Ivan and I agreed on that approach...

achen2401:
@OptimusRhine 
Adam, thank you very much for the explanation.  It makes sense and I like that we are covering grounds with this approach.

PR #1352, opened 2017-09-12
minify GIL stylesheet

achen2401:
minify GIL stylesheet
another change to enhance performance


ivan-c:
Did you "minify" the .less files too? Unless they're easier to read it shouldn't affect anything, but you probably already knew that :)

achen2401:
@ivan-c 
Ivan, css file is minified, not the less file :-P
I added minifying css as part of my compiling process now (via gulp taskrunner)

ivan-c:
Ah ok, I saw that the less file had changed but wasn't sure if you were using a tool to collapse some of the whitespace. Thanks for the explanation!

PR #1358, opened 2017-09-14
Feature/frontend accessibility optimization fixes

achen2401:
accessibility fixes:
add aria label to input fields
add alt text to images
optimization fixes:
compress png images via https://tinypng.com/ - I think they were left out last time
uglify static js file from external libraries (e.g. kiosk_function.js, vendor.js)
get minified version of i18next libraries from unpkg


achen2401:
Ivan - I very much agree about gzip compression - can we, as you mentioned, move this step into our "deploy" process? I think it will help with the performance quite a bit.
The performance I saw when I tested our portal pages via Chrome's lighthouse plugin, was quite discouraging:
https://chrome.google.com/webstore/detail/lighthouse/blipmdconlkpinefehnmjammfjpmpbjk?hl=en
It did suggest using gzip to compress text.

achen2401:
@ivan-c Ivan, I un-uglify the JS files for now.  I still think it is worthwhile for us to look into gzip compression to compress our files.

mcjustin:
@achen2401 when I think 'accessibility' I think accommodating users w/ compromised vision, hearing, etc. This change is more about performance, isn't it?

achen2401:
@mcjustin 
Justin, the aria label changes are accessibility changes (i.e. for screen reader used by vision impaired users) and yeah, others are performance fixes.

ivan-c:
>Ivan - I very much agree about gzip compression - can we, as you mentioned, move this step into our "deploy" process? I think it will help with the performance quite a bit.
@achen2401 What's your tool of choice (and invocation) for minifying/uglyifying JS?

It might have to wait until we deploy with docker since generating the minified files will probably require a good chunk of the NodeJS stack, or changing our deploy.sh script significantly.

PR #1359, opened 2017-09-14
Feature/qb dates

pbugni:
Significant refactor to QuestionnaireBanks:
- The timing (start, overdue, expired) is now at the QuestionnaireBank level rather than the individual Questionnaires.
- Date fields take relativedelta values to handle months, etc.
- Only one QB should apply to a user at any time - available at QuestionnaireBank.most_current_qb(user).
- Simplified logic for assessment status based on current_db logic
- Upgrade breaks QNR.questionnaire_bank_id - will need to re-establish once new QBs are loaded in.
- Mock QBs should match the current spec for ePROMs and SymptomTracker
- New site_persistence branch (https://github.com/uwcirg/ePROMs-site-config/pull/56) will be required with this commit.


pbugni:
@OptimusRhine I agree - we need better glue to track which iteration a QNR references in the case of recurring.

Just added an `iteration_count` to QBQ, and modified `recur.active_interval_start` to bubble this number up.  Outstanding work to associate QNR -> QBQ when `iteration_count` != 0.

Given the timing - I think we should merge this and address the problems moving forward.  I see both site persistence branches needing this PR have already been merged (meaning problems on `flask sync` if this isn't merged as well).

OptimusRhine:
@pbugni Agreed. Looked at the new changes (incl. commenting out the migration), and it looks good to me.

PR #1362, opened 2017-09-15
Reporting Dashboard v2.1

OptimusRhine:
https://www.pivotaltracker.com/story/show/150828841

* moved reporting stat generation to separate model
  * fixed minor errors in stat generation (corrected some checks, etc)
  * added stats for intervention access (not checking if `interv` in `user.interventions`, but rather checking `interv.display_for_user(user).access`)
  * added renaming of 'Decision Support' to 'Decision Support P3P' (just within the stats, not the actual db entries)
* updated reporting template page
  * corrected and added minor details to existing descriptions
  * moved Intervention Report counts to separate table (as users can have interv reports even if they don't have those intervs in their `user.interventions`)
  * added Intervention Access table
  * fixed date checks for registration/logins
  * no longer showing blank-named rows in tables (mostly relevant for ePROMs interventions)
* added celery task for refreshing reporting stats
  * ~**TO DO:** add entry for task to persistence files~ (see comment below)
* added reporting stats tests


OptimusRhine:
Adding ScheduledJob to TrueNTH:
https://github.com/uwcirg/TrueNTH-USA-site-config/pull/54

portal/models/i18n.py
```diff
@@ -254,6 +254,6 @@ def merge_po_into_master(po_path, language, fname):
 
 @babel.localeselector
 def get_locale():
-    if current_user() and current_user().locale_code:
+    if session and current_user() and current_user().locale_code:
```
pbugni: Lets move the check for session into `current_user`

portal/models/reporting.py
```diff
@@ -0,0 +1,81 @@
...
+    stats['intervention_access'] = defaultdict(int)
+    stats['intervention_reports'] = defaultdict(int)
+    stats['organizations'] = defaultdict(int)
+    stats['registrations'] = []
+    stats['encounters'] = defaultdict(list)
+
+    interventions = Intervention.query.all()
+
+    for user in User.query.filter_by(active=True):
+        if ROLE.TEST in [r.name for r in user.roles]:
```
pbugni: There is a `user.has_role(ROLE.TEST)` method for this...

portal/models/reporting.py
```diff
@@ -0,0 +1,81 @@
...
+                    stats['patients']['dx-t'] += 1
+                if any((obs.codeable_concept == CC.PCaLocalized
+                        and obs.value_quantity == CC.FALSE_VALUE)
+                       for obs in user.observations):
+                    stats['patients']['meta'] += 1
+
+        for interv in interventions:
+            desc = interv.description
+            if desc == 'Decision Support':
+                desc = 'Decision Support P3P'
```
pbugni: Hmm - there are multiple `decision support` options.  I'm not sure we should be forcing this to P3P here.  Would it be safer to switch on intervention.name?

portal/models/reporting.py
```diff
@@ -0,0 +1,81 @@
...
+            if interv.display_for_user(user).access:
+                stats['intervention_access'][desc] += 1
+            if interv in user.interventions:
+                stats['interventions'][desc] += 1
+            if (any(doc.intervention == interv for doc in user.documents)):
+                stats['intervention_reports'][desc] += 1
+
+        if not user.organizations.all():
```
pbugni: `user.organizations.count()`

portal/models/reporting.py
```diff
@@ -0,0 +1,81 @@
...
+from .procedure_codes import known_treatment_not_started
+from .role import ROLE
+from .user import User
+
+
+@dogpile_cache.region('hourly')
+def get_reporting_stats():
+    """Cachable interface for expensive reporting data queries
+
+    The following code is only run on a cache miss.
```
pbugni: I know there's logging in the celery task, but I think it'd be good to note this is running (catch any overlooked cache misses).

OptimusRhine:
@pbugni Updated with your suggested changes. Note that I had to modify your recent change to `AssessmentStatus.__str__`, because when testing with eproms-stg data I was getting some AssessmentStatus objects where `self.qb_data.qb == None`

PR #1368, opened 2017-09-18
Store QBR iteration num on QNR

OptimusRhine:
https://www.pivotaltracker.com/story/show/150859216

* added `qb_iteration` field to QNRs
  * if `None`, assume QB has no iteration (aka is baseline)
  * otherwise, signifies the iteration number of the QBRecur at the time of QNR receipt
* added migration for above field
* changed `QuestionnaireBank.most_current_qb(user)` to return a tuple of `(QuestionnaireBank, iteration_num)`, where `iteration_num` is the iteration count of the current QBRecur (if it exists; else, `interation_num` is `None`)
  * updated existing calls of `most_current_qb()`
* updated QNR [POST] logic to properly set the QB and IC, using the new `most_current_qb()`
* updated test for above QNR [POST], to add a UserConsent for the QB's org (so that `most_current_qb()` properly found and returned a QB)


portal/models/questionnaire_bank.py
```diff
@@ -256,7 +256,8 @@ def validate_classification_count(qbs):
 
     @staticmethod
     def most_current_qb(user, as_of_date=None):
-        """Return "most current" qb for user, or None
+        """Return tuple of "most current" qb for user (or None), and recurrence
```
pbugni: Silly documentation style, but we'll try and stick to it.

First line should be < 80 chars, just a one liner.  If longer description is needed, it should be in its own paragraph:

```
"""Return two item tuple (qb, iteration_count) for user

Returns tuple of "most current" qb for user (or none if N/A) and the recurrence iteration number if the found QB is recurring, else None.
"""
```

portal/models/assessment_status.py
```diff
@@ -105,7 +105,7 @@ class QuestionnaireBankDetails(object):
     """
     def __init__(self, user):
         self.user = user
-        self.qb = QuestionnaireBank.most_current_qb(user)
+        self.qb = QuestionnaireBank.most_current_qb(user)[0]
```
pbugni: More "pythonic" style is to throw away the part of the return tuple you don't need:

`self.qb, _ = QuestionnaireBank.most_current_qb(user)`

OptimusRhine: @pbugni Gotcha. Updated all existing calls this way.

PR #1372, opened 2017-09-19
QNR to QB/IC association migration

OptimusRhine:
https://www.pivotaltracker.com/story/show/150859216

* modified `QuestionnaireBank.most_current_qb()`, `QuestionnaireBank.calculated_start()`, `Recur.active_interval_start` to properly accept/handle `as_of_date` param
  * updated existing calls
* added migration to updated existing QuestionnaireResponses with `questionnaire_bank_id` and `qb_iteration` values

### DIFF
--- a/portal/models/assessment_status.py
+++ b/portal/models/assessment_status.py
@@ -86,7 +86,7 @@ def qb_status_dict(user, questionnaire_bank):
     if not questionnaire_bank:
         return d
     trigger_date = questionnaire_bank.trigger_date(user)
-    start, _ = questionnaire_bank.calculated_start(trigger_date, None)
+    start, _ = questionnaire_bank.calculated_start(trigger_date)
     overdue = questionnaire_bank.calculated_overdue(trigger_date)
     expired = questionnaire_bank.calculated_expiry(trigger_date)
     for q in questionnaire_bank.questionnaires:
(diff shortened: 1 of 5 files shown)


portal/models/assessment_status.py
```diff
@@ -86,7 +86,7 @@ def qb_status_dict(user, questionnaire_bank):
     if not questionnaire_bank:
         return d
     trigger_date = questionnaire_bank.trigger_date(user)
-    start, _ = questionnaire_bank.calculated_start(trigger_date)
+    start, _ = questionnaire_bank.calculated_start(trigger_date, None)
```
pbugni: The default param for that second arg is None, no need to be explicit, right?

portal/models/questionnaire_bank.py
```diff
@@ -329,14 +326,14 @@ def calculated_start(self, trigger_date):
 
     def calculated_expiry(self, trigger_date):
         """Return calculated expired date (UTC) for QB or None"""
-        start, ic = self.calculated_start(trigger_date)
+        start, ic = self.calculated_start(trigger_date, None)
```
pbugni: Same default arg issue.

portal/models/assessment_status.py
OptimusRhine: Well, there's no default param for that arg in `calculated_start()`. Although you're right, probably easier to add it there.

PR #1378, opened 2017-09-20
Current QB in /patients list

OptimusRhine:
https://www.pivotaltracker.com/story/show/150112352

* added new namedtuple object `QBD` to QuestionnaireBank module
* modified `QB.calculated_start()` and `QB.most_current_qb()` to return `QBD` namedtuple
  * updated existing calls to handle new return object properly
* added `QBD` info to assessment cache
* added display version of current QB to patient information for /patients list
  * display version is a combination of the QB name prefix (e.g. 'IRONMAN'), the classification (e.g. 'Baseline', 'Recurring'), and if recurring, the month total, tabulated through a combination of the current Recur attributes and the iteration count
* added new 'Visit' column to /patients page table, which contains the display version of the current QB (per above)

### DIFF
--- a/portal/models/questionnaire_bank.py
+++ b/portal/models/questionnaire_bank.py
@@ -315,8 +315,8 @@ class QuestionnaireBank(db.Model):
         be non zero if it takes multiple iterations to reach the active cycle.
 
         :return: namedtuple QBD (datetime of the questionnaire's start date,
-            iteration_count, recurrence, and not-yet-populated QB field);
-            QBD(None, None, None, None) if N/A
+            iteration_count, recurrence, and self QB field);
+            QBD(None, None, None, self) if N/A
 
         """
         # On recurring QB, deligate to recur for date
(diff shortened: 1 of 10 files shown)


portal/models/questionnaire_bank.py
```diff
@@ -278,41 +283,40 @@ def most_current_qb(user, as_of_date=None):
...
-                                                         as_of_date)
-                if relative_start is None:
+                qbd = qb.calculated_start(trigger_date, as_of_date)
+                if qbd.relative_start is None:
                     # indicates QB hasn't started yet, continue
                     continue
                 expiry = qb.calculated_expiry(trigger_date)
-                last_found = qb
-                last_found_ic = ic
+                last_found = qbd._replace(questionnaire_bank=qb)
```
pbugni: seems like the iteration should be captured in the `qbd._replace()` as well?

portal/models/questionnaire_bank.py
```diff
@@ -321,22 +325,25 @@ def calculated_start(self, trigger_date, as_of_date=None):
                 (relative_start, ic) = recurrence.active_interval_start(
                     trigger_date=trigger_date, as_of_date=as_of_date)
                 if relative_start:
-                    return (relative_start, ic)
-            return (None, None)  # no active recurrence
+                    return QBD(relative_start=relative_start, iteration=ic,
+                               recur=recurrence, questionnaire_bank=None)
```
pbugni: Why don't we populate QBD.questionnaire_bank here?  isn't it `self`?

portal/models/questionnaire_bank.py
OptimusRhine: It should already be captured (because `qbd` is returned from `calculated_start()`, which contains the current iteration)

pbugni: ah!  thanks.

OptimusRhine:
@pbugni 
> Question: would it make sense to add the assessment_status to the namedtuple, or do you think it's cleaner as is?
I think it's cleaner as-is. Otherwise we have to populate `assessment_status=None` throughout all the `calculated_start()` and `most_current_qb()` code, just to make the return of `overall_assessment_status()` a bit cleaner. Seems excessive.

> 2nd question: i have no idea how this works, but as we're changing what get's cached, will we need to basically invalidate the whole assessment_cache on deploy?
...that's a good question, I also don't know how it works. I'm hoping it invalidates on deploy? I'm not sure how best to handle that, without adding a bunch of checks in the code (that will be extraneous as soon as the cache is refreshed).

portal/models/questionnaire_bank.py
OptimusRhine: Good point, I'll fix that

pbugni:
I have a feeling we're going to need to manually purge the cache as the return type (not the passed parameters) have changed.  We'll find out shortly :)

PR #1379, opened 2017-09-20
intervention access checks only apply to patients.

pbugni:
@OptimusRhine @mcjustin Can you check if this is the desired behavior?

This change will restrict the lookup for intervention access and intervention reports to only patients.

(it should also significantly cut down on the noise in the error logs)

### DIFF
--- a/portal/models/reporting.py
+++ b/portal/models/reporting.py
@@ -51,16 +51,17 @@ def get_reporting_stats():
                        for obs in user.observations):
                     stats['patients']['meta'] += 1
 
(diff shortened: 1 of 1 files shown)


pbugni:
I'm merging now to reduce the error emails.  We can revert if this isn't the desired change.

mcjustin:
@pbugni hmm sexual recovery should be available to partners as well - that's the only non-patient that comes to mind here though.
"This change will restrict the lookup for intervention access and intervention reports to only patients."

PR #1388, opened 2017-09-20
Feature/gil stylesheet fix 2

achen2401:
found original source of syntax error in gil stylesheet and re-ran less


portal/static/less/gil.less
```diff
@@ -2662,7 +2662,7 @@ input[type="search"] {
 .checkbox-inline input[type="checkbox"] {
     position: absolute;
     margin-left: -20px;
-    margin-top: 4px;
+    margin-top: 4px \9;
```
pbugni: This looks odd - should that `\9` really be there?

achen2401: Paul, that is css shorthand for IE9 and lower style

PR #1389, opened 2017-09-20
fix answer display for questionnaire response that doesn't have answer

achen2401:
per slack chat with Justin related to questionnaire responses sent by Care Plan:
questionnaire id should populate question column
answer code should populate the answer column

### DIFF
--- a/portal/templates/profile_macros.html
+++ b/portal/templates/profile_macros.html
@@ -1229,8 +1229,8 @@
                                                  '</tr>';
                         };
                     });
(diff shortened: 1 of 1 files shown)


mcjustin:
@achen2401 actually, did you confirm w/ @ivan-c ? We've had recent changes around this and more in flight...

achen2401:
@mcjustin 
Justin, Yes, I did confirm with Ivan before I made the changes.

PR #1392, opened 2017-09-21
Adding exception handling and enhanced logging in tracking down celery

pbugni:
problems.


pbugni:
May opt to roll back, sending to staging for debugging help.

PR #1399, opened 2017-09-21
Move visit_name to separate function, add tests

OptimusRhine:
* moved QB visit name construction code to separate function (in QB module)
  * added check for `qbd.questionnaire_bank` value
* created unit tests for QB visit name, for baseline and recurring scenarios

### DIFF
--- a/portal/models/questionnaire_bank.py
+++ b/portal/models/questionnaire_bank.py
@@ -2,6 +2,7 @@
 from collections import namedtuple
 from datetime import datetime
 from flask import current_app, url_for
+from re import sub
 from sqlalchemy import UniqueConstraint, CheckConstraint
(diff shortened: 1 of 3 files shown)


portal/models/questionnaire_bank.py
```diff
@@ -462,3 +463,21 @@ def add_if_not_found(self, commit_immediately=False):
             self.id = existing.id
         self = db.session.merge(self)
         return self
+
+
+def visit_name(qbd):
+    if not qbd.questionnaire_bank:
+        return None
+    name = sub('_', ' ', qbd.questionnaire_bank.name).split()[0]
```
pbugni: string.replace should cut it here - quicker and safer than re, unless you need pattern matching.

PR #1401, opened 2017-09-21
Create audit entries for generated email messages

OptimusRhine:
* added logic to `EmailMessage.send_message()`, to create an audit entry upon successful send
  * `Audit.user_id` and `Audit.subject_id` are set as the sender and first recipient, if found - otherwise, they default to `EmailMessage.user_id` (if that's also not found, no audit created)


portal/models/message.py
```diff
@@ -37,6 +39,18 @@ def send_message(self):
         message.html = fill(self.body, width=280)
         mail.send(message)
 
+        user = User.query.filter_by(email=self.sender).first()
+        user_id = user.id if user else self.user_id
```
pbugni: Hmm - the truth is, it's often the "system" sending the email, not the user.

@mcjustin do we need some sort of system account for this?  Little misleading to suggest the user sent themselves an email reminder, for example.

mcjustin: @pbugni good point, yes we should distinguish that this was done by the system automatically.

PR #1411, opened 2017-09-25
Quicker & quieter intervention access check

OptimusRhine:
* turned off InterventionStrategy logging by default, unless turned on via config setting
* added new `Intervention.quick_access_check(user)` function
  * returns `True` as soon as access is confirmed (or `False` if no access confirmed), rather than making additional checks and/or building a `DisplayDetails` object
  * skips html-generating strategies (they normally just return `True` anyway)
* modified reporting stats function to use new `quick_access_check()`
* added to existing InterventionStrategy-related Intervention tests, to confirm that `Intervention.quick_access_check(user)` acts the same as `Intervention.display_for_user(user).access`

Before these changes, reporting dashboard took 13min to load with truenth stg data.
After these changes, reporting dashboard took ~8min~ (7min with recent changes) to load with same truenth stg data.
In other words, while we should definitely continue to make the check more efficient, this change at least is an incremental improvement.

### DIFF
--- a/portal/models/intervention.py
+++ b/portal/models/intervention.py
@@ -157,7 +157,7 @@ class Intervention(db.Model):
         return DisplayDetails(access=access, intervention=self,
                 user_intervention=ui)
 
-    def quick_access_check(self, user, silent=False):
+    def quick_access_check(self, user):
         """Return boolean representing given user's access to intervention
 
         Somewhat complicated method, depending on intervention configuration.
@@ -177,7 +177,9 @@ class Intervention(db.Model):
         """
         # 1. check strategies for access
         for func in self.fetch_strategies():
-            if func(intervention=self, user=user, silent=silent):
+            if func.__name__ == 'update_user_card_html':
+                return True
+            if func(intervention=self, user=user):
                 return True
 
         # 2. check user_intervention for access
--- a/portal/models/intervention.py
+++ b/portal/models/intervention.py
@@ -165,22 +165,17 @@ class Intervention(db.Model):
         should have access to an intervention.  The first 'true' found
         is returned (as to make the check as quick as possible).
 
-        1. call each strategy_function in intervention.access_strategies.
-           Note, on rare occasions, a strategy may alter the UserIntervention
-           attributes given the circumstances.
+        1. check if the intervention has `public_access` set
         2. check for a UserIntervention row defining access for the given
            user on this intervention.
-        3. check if the intervention has `public_access` set
+        3. call each strategy_function in intervention.access_strategies.
 
         @return boolean representing 'access'.
 
         """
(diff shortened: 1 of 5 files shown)


OptimusRhine:
@pbugni Sure, that makes sense (and should make things cleaner, if we're talking about just putting the check for `LOG_DEBUG_STRATS` into the `_log()` func, and removing all new `silent` param stuff).

I assumed that the logging served some important purpose, but if it's mostly just noise, definitely seems prudent to have it off by default...

portal/models/intervention.py
```diff
@@ -157,6 +157,46 @@ def display_for_user(self, user):
...
+            if func(intervention=self, user=user):
+                return True
+
+        # 2. check user_intervention for access
+        ui = UserIntervention.query.filter_by(
+            user_id=user.id, intervention_id=self.id).first()
+        if ui and ui.access == 'granted':
+            return True
+
+        # 3. check intervention scope for access
```
pbugni: @OptimusRhine I *believe* we can run this first for the quick check.  You'd need to review what we do alter in the access strategies - but IIRC that only alters things like the html displayed in the card - doesn't alter access itself?

OptimusRhine: @pbugni I think you're right, it doesn't look like any of the strategies modify `public_access`.

It also looks like we could reverse the order here entirely (i.e. first check for `public_access`, _then_ check the `UserInterventions`, _then_ check the access strategies), unless I'm missing something. That would definitely speed things up...

pbugni: I'm sure I thought there was good reason for the original order, but if you feel you've adequately confirmed none of the current strategies would *need* to run before the other checks, lets change it.

Please double check - I'm sure I only chose this more inefficient order as it was necessary at the time.

OptimusRhine: Well, the order actually still makes since for `display_for_user()`:
* The InterventionStrategy loop needs to happen:
  * ...before the UserIntervention query, because the UserIntervention is returned as part of the `DisplayDetails` object. So if a UI is generated in the IS loop, it needs to happen first.
  * ...no matter what, because of the potential for HTML to be generated, etc
* Given the Strategy loop always needs to happen, the `public_access` check placement doesn't really matter, as long as its there - moving it won't change the runtime of the function.

But none of these are concerns for the quick access check, so I'm going to go ahead and switch that order up, see if it helps things.

OptimusRhine: (switching it up improved things by ~1min)

OptimusRhine:
@pbugni Only other instance that didn't require the `DisplayDetails` return from `display_for_user()`, was in the QuestionnaireBank model. Updated that instance to use `quick_access_check()` instead.

PR #1416, opened 2017-09-26
Properly handle bad access tokens

OptimusRhine:
* modified `access_via_token()`, so that when the embedded `verify_token()` fails, we actually redirect to the returned redirect, rather than treating it as a `user_id`
* added test for bogus token (confirm redirect to landing)


pbugni:
Perfect - thank you.

PR #1417, opened 2017-09-26
Celery init factory

ivan-c:
* Move app and celery initialization into separate module (`factories`)
* Initialize celery from a factory function
* Name tasks and call by name (with `celery.send_task()`) to prevent cyclic imports

Adapted from:
* [Configuring Celery](http://flask.pocoo.org/docs/0.12/patterns/celery/)
* [Using Celery with Flask Factories](https://citizen-stig.github.io/2016/02/17/using-celery-with-flask-factories.html)

See also:
* [Celery and the Flask Application Factory Pattern](https://blog.miguelgrinberg.com/post/celery-and-the-flask-application-factory-pattern)
* [Celery integration with Flask](http://shulhi.com/celery-integration-with-flask/)

### DIFF
--- a/portal/celery_worker.py
+++ b/portal/celery_worker.py
@@ -14,14 +14,12 @@ from .database import db
 from factories.celery import create_celery
 from factories.app import create_app
 from .models.scheduled_job import ScheduledJob
+import tasks
 
 
 app = create_app()
 celery = create_celery(app)
 app.app_context().push()
-
-# Todo: fix 'RuntimeError: Working outside of application context.'
-import tasks
 
 @celery.on_after_configure.connect
 def setup_periodic_tasks(sender, **kwargs):
--- a/portal/celery_worker.py
+++ b/portal/celery_worker.py
@@ -10,17 +10,18 @@ Launch in the same virtual environment via
   $ celery worker -A portal.celery_worker.celery --loglevel=info
 
 """
-from .app import create_app
 from .database import db
-from .extensions import celery
+from factories.celery import create_celery
+from factories.app import create_app
 from .models.scheduled_job import ScheduledJob
-import tasks
 
-assert celery  # silence pyflake warning
 
 app = create_app()
+celery = create_celery(app)
 app.app_context().push()
 
+# Todo: fix 'RuntimeError: Working outside of application context.'
+import tasks
 
 @celery.on_after_configure.connect
 def setup_periodic_tasks(sender, **kwargs):
--- a/portal/factories/app.py
+++ b/portal/factories/app.py
@@ -171,9 +171,6 @@ def configure_extensions(app):
     # flask-session - Server side sessions
     session.init_app(app)
 
-    # celery - task queue for asynchronous tasks
-    # celery.init_app(app)
-
     # babel - i18n
     babel.init_app(app)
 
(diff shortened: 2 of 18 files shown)


portal/views/portal.py
```diff
@@ -1199,7 +1204,7 @@ def communicate(email):
 
 @portal.route("/post-result/<task_id>")
 def post_result(task_id):
-    r = post_request.AsyncResult(task_id).get(timeout=1.0)
+    r = AsyncResult(task_id, app=celery).get(timeout=1.0)
```
OptimusRhine: Do we need a `celery = create_celery(current_app)` initialization here?

portal/celery_worker.py
```diff
@@ -10,17 +10,18 @@
...
 from .models.scheduled_job import ScheduledJob
-import tasks
 
-assert celery  # silence pyflake warning
 
 app = create_app()
+celery = create_celery(app)
 app.app_context().push()
 
+# Todo: fix 'RuntimeError: Working outside of application context.'
```
pbugni: What is the source of this?  This will happen when you try and access something like `current_app` outside of a valid context - such as using it at file scope rather than within a function called during a valid context.

portal/factories/app.py
```diff
@@ -172,7 +172,7 @@ def configure_extensions(app):
     session.init_app(app)
 
     # celery - task queue for asynchronous tasks
-    celery.init_app(app)
+    # celery.init_app(app)
```
pbugni: lets get it out of here - no value in leaving comments around :)

portal/models/auth.py
```diff
@@ -152,7 +153,9 @@ def notify(self, data):
 
         # Use celery asynchronous task 'post_request'
         kwargs = {'url': self.callback_url, 'data': formdata}
-        res = post_request.apply_async(kwargs=kwargs)
+        celery = create_celery(current_app)
```
pbugni: I do think we want to use a singleton pattern, rather than create another celery on every request such as this.

portal/views/scheduled_job.py
```diff
@@ -1,15 +1,16 @@
...
 
 from ..audit import auditable_event
 from ..database import db
-from ..extensions import celery, oauth
+from ..extensions import oauth
 from ..models.role import ROLE
 from ..models.scheduled_job import ScheduledJob
 from ..models.user import current_user
 
+from ..factories.celery import create_celery
```
pbugni: can we keep the imports alphabetized?  there's a section for external followed by a line of whitespace and then the locals.

portal/factories/celery.py
```diff
@@ -0,0 +1,27 @@
...
+from celery import Celery
+
+
+__celery = None
+
+def create_celery(app):
+    global __celery
+    if __celery:
+        return __celery
+    celery = Celery(
```
pbugni: missed one - i'm sure you meant to assign the return to `__celery` here...

portal/factories/celery.py
```diff
@@ -0,0 +1,27 @@
...
+        app.import_name,
+        broker=app.config['CELERY_BROKER_URL']
+    )
+    celery.conf.update(app.config)
+    TaskBase = celery.Task
```
pbugni: and return it...

PR #1440, opened 2017-09-29
ScheduledJobs feature enhancements

OptimusRhine:
* added `last_status` field to ScheduledJob
  * added migration
* changed ScheduledJob-updating logic from tasks
  * now updates both `ScheduledJob.last_runtime` and `ScheduledJob.last_status`
  * cleaned up task calls to reflect new update functionality, and use better error handling
* added manual task trigger functionality to ScheduledJob
  * includes model function to actually run the task and API endpoint to trigger the model function
  * added tests
* updated /scheduled_jobs page to be cleaner, and include a 'Run Job' button (using trigger functionality described above)


OptimusRhine:
Looks like this is causing some funky timeout issues in the test...closing for now, will reopen when I've got a fix.

portal/models/scheduled_job.py
```diff
@@ -80,13 +82,28 @@ def crontab_schedule(self):
...
                       )
 
+    def trigger(self):
+        from .. import tasks
+        func = getattr(tasks, self.task, None)
+        if func:
+            args_in = self.args.split(',') if self.args else []
+            kwargs_in = self.kwargs or {}
+            try:
+                msg = func(*args_in, job_id=self.id, **kwargs_in)
```
pbugni: Granted, you're naming the `job_id` param, but shouldn't the order be `(job_id, *args, **kwargs)` ?  that's what i'm more familiar with seeing and i think it'll possibly bind incorrectly otherwise.

OptimusRhine: Mmm, I think it'll work this way...at least, it worked for my testing (albeit with empty args/kwargs), and that's what I've been doing in `celery_worker.py`. But I can test in the suggested order and see if that works.

OptimusRhine: Oh, nevermind, you're right. Changing now (and adding additional tests).

OptimusRhine:
Ugh...a test is timing out in Travis, despite working fine for me locally. Digging in deeper...

OptimusRhine:
Tests fixed, good to go.

PR #1446, opened 2017-09-30
Cleanup deb-builder

ivan-c:
* Update `builder` (docker image for building debian packages) to use stretch
* Remove backport cruft

### DIFF
--- a/docker/Dockerfile.build
+++ b/docker/Dockerfile.build
@@ -4,8 +4,9 @@ ENV DEBIAN_FRONTEND noninteractive
 
 RUN \
     apt-get update && \
-    apt-get dist-upgrade -y && \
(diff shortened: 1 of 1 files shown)


docker/Dockerfile.build
```diff
@@ -1,25 +1,7 @@
-FROM debian:jessie
+FROM debian:stretch
```
pbugni: Great  - no problem, but I wonder if this is a push to get our other TrueNTH systems up to stretch to be in alignment?

ivan-c: It would be nice, but luckily it looks like there isn't much of a difference to our portal code. This did serve well as a smoke-test though :)

PR #1452, opened 2017-10-03
Fix version generation

ivan-c:


ivan-c:
Can't seem to change description...
* setuptools_scm now prepends a letter (corresponding to the SCM) to the hash.
We may need to delete the `.eggs/` directory which may contain older versions of setuptools_scm, downloaded during the installation of the portal package

PR #1454, opened 2017-10-03
Convert communication q_due_date to user's TZ

OptimusRhine:
https://www.pivotaltracker.com/story/show/151540337

* modified `questionnaire_due_date` for LR-generated Communication emails, to convert from UTC to the user's TZ (as that may modify the due date by +/- 1), if user TZ is set
* added tests for `questionnaire_due_date` (without and with user TZ)
  * added new `setdate` param to `bless_with_basics()`, to make sure we can test a time where converting to another timezone would affect the day

### DIFF
--- a/portal/models/communication.py
+++ b/portal/models/communication.py
@@ -1,7 +1,6 @@
 """Communication model"""
 from collections import MutableMapping
 from flask import current_app, url_for
-from pytz import utc, timezone
 import regex
 from sqlalchemy import UniqueConstraint
 from sqlalchemy.dialects.postgresql import ENUM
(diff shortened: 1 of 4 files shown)


portal/models/communication.py
```diff
@@ -127,7 +128,13 @@ def _lookup_questionnaire_due_date():
         trigger_date = qb.trigger_date(user)
         due = (qb.calculated_overdue(trigger_date) or
                qb.calculated_expiry(trigger_date))
-        return due.strftime('%-d %b %Y') if due else ''
+        due_date = ''
+        if user.timezone and due:
+            local = utc.localize(due).astimezone(timezone(user.timezone))
```
pbugni: Can we make this a function on user or perhaps better in `date_tools`?  I would expect there are several places where we'll need to return the timezone adjusted date / time.

OptimusRhine: Sure thing

OptimusRhine:
@pbugni Should the `intervention_strategies` fix be part of this PR? Or should we go ahead and get this change in, then add that separately?

pbugni:
@OptimusRhine fair - let's file as a separate issue.  Merging.

PR #1467, opened 2017-10-04
indent code in initial queries, fix memory leak and add delay between

achen2401:
adding some fixes for initial queries after testing in various browsers:
- assign item organization id to a private variable (add var), it was in the global scope and was creating some unexpected behavior in IE
-  when saving organization information - allow a slight delay in saving to allow consent to be saved
- indent code in initial queries


achen2401:
adding some fixes for safari, hope they will take, will reopen when ready

PR #1468, opened 2017-10-04
Localize AssessmentStatus card_html display dates

OptimusRhine:
https://www.pivotaltracker.com/story/show/151666519

* populate assessment_status card_html dates (due date, completion date, etc) based on user's local timezone, not just always UTC
* add tests for above display dates
  * includes update to `test_assessment_status.mock_qr()`, to add a timestamp param

### DIFF
--- a/tests/test_assessment_status.py
+++ b/tests/test_assessment_status.py
@@ -20,8 +20,8 @@ from portal.models.fhir import QuestionnaireResponse
 from tests import TestCase, TEST_USER_ID
 
 
-def mock_qr(user_id, instrument_id, status='completed'):
-    today = datetime.utcnow()
+def mock_qr(user_id, instrument_id, status='completed', timestamp=None):
+    today = timestamp or datetime.utcnow()
     qr_document = {
         "questionnaire": {
             "display": "Additional questions",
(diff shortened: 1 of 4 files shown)


tests/test_assessment_status.py
```diff
@@ -20,8 +20,8 @@
 from tests import TestCase, TEST_USER_ID
 
 
-def mock_qr(user_id, instrument_id, status='completed'):
-    today = datetime.utcnow()
+def mock_qr(user_id, instrument_id, status='completed', timestamp=None):
+    today = timestamp or datetime.utcnow()
```
pbugni: can we change `today` to reflect potential reality :)

how about:
`timestamp = timestamp or datetime.utcnow()`

tests/test_intervention.py
```diff
@@ -398,17 +399,28 @@ def test_card_html_update(self):
         self.assertTrue(
             user.display_name in ae.display_for_user(user).card_html)
 
+        dt = datetime(2017, 06, 10, 20, 00, 00, 000000)
```
pbugni: putting a zero in front of an integer, means "octal" - in this case they're equal, but I think we should avoid it unless we want an octal value:
```
>>> 06 == 6
True
>>> 011 == 11
False
```

tests/test_assessment_status.py
OptimusRhine: Isn't all time relative anyway, even the word "today"? Maybe the test is just living in the past...

:)

Yea, I'll update that

PR #1482, opened 2017-10-09
Exclude org name from QB visit status

OptimusRhine:
https://www.pivotaltracker.com/story/show/150112352

* exclude org name (IRONMAN, CRV) from QB visit status
* flip month display around ('X Month' -> 'Month X')
* updated tests to look for new visit name display format


OptimusRhine:
Er, this should be against

PR #1484, opened 2017-10-09
frontend error logging changes

achen2401:
add frontend error logging for failed ajax calls
- refactor code to have a common ajax call function which:
       1.  will log error (via ajax call to server) if the ajax call fails.
       2.  allows callback upon ajax success or failure

**Not for this release**
can be merged to the develop branch after this is reviewed.


pbugni:
@achen2401 Love this change.  I take it you feel it's risky to include in the current release?  (guess we have to draw the line somewhere...)

achen2401:
@pbugni 
Paul, yeah, a bit risky because I did some re-factoring there.  so some rounds of testing are required to make sure nothing is broken.

mcjustin:
Agreed - not for this release

PR #1485, opened 2017-10-09
Correct crv baseline questionnaire rank w/ migration

pbugni:
(FKs prevent site_persistence fix from working)

@mcjustin look at the `desired_order` list to review the order they end up in.


pbugni:
@mcjustin this is a fix for the immediate problem.

Site_persistence may need to become more intelligent in time - for now, it works great unless small mods like this are requested on existing data.  Generally, it looks for changes, leaves alone matches, creates new and destroys old for modified data.  Can't handle this sort of change where we just want to fix an attribute and leave all the existing related data alone.

PR #1490, opened 2017-10-09
Flush redis before updating DB

ivan-c:
**NB: This will merge into RC**
* Call `FLUSHDB` on configured redis DB (`REDIS_URL`) before updating database

### DIFF
--- a/manage.py
+++ b/manage.py
@@ -5,6 +5,7 @@ FLASK_APP=manage.py flask --help
 """
 import os
 import click
+import redis
 
 import alembic.config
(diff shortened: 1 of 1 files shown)


ivan-c:
Yeah, it's a tradeoff between maintaining what needs to be invalidated and speed. I figured the temporary slowdown was OK since some downtime will be expected whenever we're updating a deployment.

pbugni:
Agreed.  We can always make use of the /scheduled_jobs API and run a few manually right after deploy.

PR #1491, opened 2017-10-10
Communications preview page

OptimusRhine:
https://www.pivotaltracker.com/story/show/151268523

* added `preview` function for Communications
  * generates a message, but does NOT send
  * includes new endpoint (`/communicate/preview/<id>`) for returning preview message data in json format
* added admin-only Communications list page (accessible at `/communicate`)
  * entries have 'preview' button, which creates popup with a preview of the email (_whether or not it has actually been sent yet_) using the new endpoint
* added 'Admin Tools' section to `/admin` page, which includes links to `/communicate` and `/scheduled_jobs`
* added tests


OptimusRhine:
@achen2401 
I'd definitely appreciate any suggestions/improvements for the UI changes here - mainly:

* the `/communicate` page
  * the preview popups from that page
* the new 'Admin Tools' section of the `/admin` page

Not a huge deal, since this all should be only visible to admins anyway, but if you have a few moments to spare it could definitely use the Amy magic. Thanks!

OptimusRhine:
@achen2401 The changes look great! Especially that switch to a modal popup - it's definitely smoother, cleaner, and avoids that popup blocker issue. Thanks so much!

portal/models/communication.py
```diff
@@ -252,6 +252,34 @@ def generate_and_send(self):
         self.message.send_message()
         self.status = 'completed'
 
+    def preview(self):
```
pbugni: Can we avoid repeating such similar code blocks between `generate_and_send` and `preview`?  (i.e. both use a common block?)

OptimusRhine: Sure thing - I've moved that repeated code into a separate function.

OptimusRhine: @pbugni (whenever you get a chance, this is ready for another review)

mcjustin:
@ivan-c since Paul is out, can you please review and merge if acceptable?

portal/models/communication.py
```diff
@@ -244,14 +229,49 @@ def generate_and_send(self):
...
+        if current_app.config.get('DEBUG_EMAIL', False):
+            # hack to restart trace when in loop from celery task
+            # don't want to reset if in the middle of a request
+            from celery import current_task
+            establish_trace(
+                "BEGIN trace as per DEBUG_EMAIL configuration",
+                reset_trace=current_task is None)
+
+        user = User.query.get(self.user_id)
+        if not user.email or '@' not in user.email:
```
ivan-c: Does the `@` need to be unicode (`u'@'`)?

portal/templates/communications.html
```diff
@@ -0,0 +1,116 @@
+{% extends "layout.html" %}
+{% block main %}
+    <h1>Communications</h1>
```
ivan-c: Tag with `<trans>`/`__()` please :)

portal/models/communication.py
OptimusRhine: It looks like this doesn't matter either way (`'@' in 't@st'`, `u'@' in 't@st'`, `'@' in u't@st'`, and `u'@' in u't@st'` all return `True`), but I'll change it anyway

PR #1494, opened 2017-10-10
Remove old WSGI entrypoint

ivan-c:
Supplanted by wsgi.py common entrypoint (Gunicorn & Apache)


ivan-c:
Should be OK to remove now- updated all the Apache deployments to use the new WSGI entrypoint

PR #1498, opened 2017-10-10
Fix accidentally deleted "sender" attribute

ivan-c:


OptimusRhine:
We may not actually need this PR after all...I moved a lot of the `generate_and_send()` code out to a new function anyway, including the `MailResource()` init, and added in the sender param there - see #1491

ivan-c:
@OptimusRhine Cool, will close/delete then

PR #1505, opened 2017-10-12
initial queries buttons fixes

achen2401:
minor styling fix for buttons on initial queries page:
- remove css for opacity for buttons on initial queries - not necessary anymore as they are being hidden and displayed via js


achen2401:
close this one - will combine these styling changes with my upcoming branch for initial queries

PR #1506, opened 2017-10-12
Invalidate overall_assessment_status on task run

OptimusRhine:
* added invalidation step to `cache_assessment_status()`, on top of the refresh (refresh only works if the cache is already invalidated or expired)

### DIFF
--- a/portal/tasks.py
+++ b/portal/tasks.py
@@ -188,6 +188,7 @@ def update_patient_loop(update_cache=True, queue_messages=True):
 
     for user in valid_patients:
         if update_cache:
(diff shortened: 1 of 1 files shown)


pbugni:
@OptimusRhine I still think it should work without, but if you're certain it doesn't, let's merge.

OptimusRhine:
I'm not certain about anything when it comes to dogpile :)

I just know that when I tested locally, just the `refresh` didn't seem to make a difference, unless either the hourly time limit had already passed, or it was preceded by the `invalidate`. Ultimately up to you though, I'm OK with or without it.

PR #1511, opened 2017-10-13
Move request cache duration to formal config

ivan-c:
* Default value should appear in /settings now
* Updated value on prod to 8 hours

### DIFF
--- a/portal/config.py
+++ b/portal/config.py
@@ -42,6 +42,8 @@ class BaseConfig(object):
         'REQUEST_CACHE_URL',
         REDIS_URL
     )
+    REQUEST_CACHE_EXPIRE = 10
+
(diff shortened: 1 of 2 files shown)


ivan-c:
I updated the value on prod to 8 hours, but we should be flushing redis on deploy so invalid entries shouldn't be an issue. 

Justin also mentioned it might be worthwhile to pre-populate the cache on deploy (maybe we could iterate through apptext table looking for links?)

pbugni:
@ivan-c, @mcjustin if we want to create a script to pre-populate the cache, I think that's a fine idea, but a bit more complicated than iterating through the apptext table.  For example, we have dynamic values in several strings such as organization name.  Still, very reasonable, just needs a story...

PR #1513, opened 2017-10-13
Hotfix/update AUA orgs

OptimusRhine:
To be merged in conjunction with https://github.com/uwcirg/ePROMs-site-config/pull/75

* added new migration for updated AUA org structure
  * org 21100 renamed to "Australian Urology Associates (AUA)"
  * org 21110 changed to 21300, parent changed to 21000, and renamed to "Australian Prostate Cancer Research Centre - Queensland"
  * org 21120 changed to 21400, parent changed to 21000, and renamed to "Eastern Health"

Note that this is a "trial run" of the org ID changing process; once this is confirmed working as expected (or once any issues are fixed), we'll add this as a function in `portal.models.organization`, for future use.


portal/migrations/versions/b814b51681bf_.py
```diff
@@ -0,0 +1,103 @@
...
+    for oa in session.query(OrganizationAddress).filter_by(organization_id=old_id):
+        oa.organization_id = new_id
+    session.commit()
+
+    session.execute('DELETE FROM organizations WHERE id = {}'.format(old_id))
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    update_org_id(21110, 21300, 21000, "Australian Prostate Cancer Research Centre - Queensland")
```
ivan-c: Using kwargs (instead of positional) would make this a little easier to read

PR #1514, opened 2017-10-16
fix biopsy q ui issue - allowing checking of no/unknown questions

achen2401:
@mcjustin Justin, I stumbled upon this,  when I was in the initial queries page.   It seemed I couldn't check the 'no' or 'unknown' option for the biopsy question anymore.
I traced back and found the culprit code - that was added in this PR that addressed Firefox specific behavior a little while ago:
https://github.com/uwcirg/true_nth_usa_portal/pull/1475/files

This PR should fix the issue.  And I think this should be added as a hotfix. I see this is affecting demo and production.
The fixes are:
- instead of setting "checked" property of biopsy question to false, simply display the biopsy date input fields if the biopsy question is checked.
- allow data-trigger-event attribute, if specified, to be used to specify event type that triggers an event
- remove white spaces in comments

### DIFF
--- a/portal/static/js/main.js
+++ b/portal/static/js/main.js
@@ -247,9 +247,9 @@ var fillViews = {
                 content = $("#patBiopsy input[name='biopsy']:checked").closest("label").text();
                 content += "&nbsp;&nbsp;" + displayDate;
             } else {
-                f.prop("checked", false);
                 content = $("#patBiopsy input[name='biopsy']:checked").closest("label").text();
             };
+            if ($("#patBiopsy input[name='biopsy']").is(":checked")) $("#biopsyDateContainer").show();
             if (hasValue(content)) $("#biopsy_view").html("<div>" + i18next.t(content) + "</div>");
             else $("#biopsy_view").html("<p class='text-muted'>" + __NOT_PROVIDED_TEXT + "</p>");
         };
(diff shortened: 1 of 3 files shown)


mcjustin:
@ivan-c this will need to be another hotfix. Is master the proper merge-to destination?

mcjustin:
@ivan-c this will need to be another hotfix. Is master the proper merge-to destination? We'll need to test at stg.us.truenth.org first...

ivan-c:
@mcjustin If we're hot-fixing, master is the right place; we'll just have to update stg* to master (as opposed to develop as it typically is).

mcjustin:
I'm going to note behavior that I'm seeing with this fix on stg.us. I don't know how much of this was introduced in this change, or how much has been this way for a while now...

Patient 11627:
For "Have you had a prostate cancer biopsy?", I answered "No". The 'NEXT' button appeared [good], but so did the date fields for the 'Yes' option [*not good, but not a critical issue*].
I wasn't asked about Diagnosis or localized/metastacized [good].

My profile then says:
Have you had a prostate cancer biopsy? | No [good]
Have you been diagnosed with prostate cancer? | No (my biopsy was negative) [**The no part is correct, but the second part isn't exactly true, since there was no biopsy. Not a critical issue**]
Is the prostate cancer only within the prostate? | No (the cancer is in other parts of my body, too) [**This is certainly a bug. Is it in the data, or just the UI?**]

So @achen2401 , if you could please look at that last item in bold.

PR #1521, opened 2017-10-17
TN232 and TN 233 update checkbox content for Truenth and Eproms

achen2401:
address stories:
TN 232:  https://jira.movember.com/browse/TN-232
TN 233:  https://jira.movember.com/browse/TN-233

for Eproms:
update content for each checkbox item
for Truenth:
add a new checkbox for privacy policy

also clean-up code (added indents, removed white space, etc., styling fixes)

This needs to go out in conjunction with the change in Truenth persistence file (i.e. a new required core data: privacy policy) : PR:  https://github.com/uwcirg/TrueNTH-USA-site-config/pull/61


portal/templates/initial_queries.html
```diff
@@ -844,7 +765,90 @@ <h4 class="profile-item-title"> {{ _("Thank you.") }}</h4>
...
+
+    /*
+     * the flow here :
+     * get still needed core data
+     * populate all required fields
+     * get incomplete fields thereafter
+     * note: need to delay gathering incomplete fields to allow fields to be render
+     */
+    configObj.initConfig(function() {
+
```
ivan-c: Would it make sense to move this section/functionality to a separate file, or does core data get acquired across multiple template/pages?

portal/templates/initial_queries_macros.html
```diff
@@ -8,31 +8,99 @@ <h4 class="tnth-headline">{{ _("Terms of Use") }}</h4>
...
+                        <span class="terms-tick-box-text">&nbsp;By checking this box, I confirm that I have read and accept the website <a
+                          href="{{privacy_url}}" class="required-link">privacy policy</a> and <a
+                          href="{{terms_url}}" class="required-link">terms</a></span>
+                        {% endtrans %}
+                        <input type="hidden"
+                          data-type="terms"
+                          data-agree="false"
+                          data-core-data-type="privacy_policy"
+                          data-tou-type="privacy policy"
+                          data-url="{{privacy_url}}"/>
```
ivan-c: Do these template variables (`privacy_url`, `tou_url`) really work outside the `{% endtrans %}` block?

portal/templates/profile_macros.html
```diff
@@ -510,24 +515,26 @@ <h4 class="text-success registration-label">{{_('registered')}}</h4>
...
-           <div class="state-selector-container selector-show"><select id="stateSelector" class="form-control"></select></div>
-           <div id="fillOrgs" userId="{{person.id if person}}" class="orgs-by-state" {%if person and person.has_role(ROLE.PATIENT) and person.id == current_user.id %}patient_view="true"{%endif%}>
-               <p class="clinic-prompt prompt selector-show">{{ _("What is your main clinic for prostate cancer care?")}}</p>
-               <div class="main-state-container selector-show">
+        <div id="clinics">
+            <div id="userOrgs" class="customized">
+               <input type="hidden" id="stock_consent_url" value="{{ url_for('portal.stock_consent', org_name='placeholder', _external=True)}}" />
+               <p class="prompt text-muted">{{ _("In what state ")}}{% if current_user and current_user.has_role(ROLE.STAFF)%}{{_("is the patient ")}}{%else%}{{_("are you ")}}{% endif %}{{_("currently receiving prostate cancer care?") }}</p>
```
ivan-c: You'll likely need to split this into two complete sentences for translation (see below). Breaking up a sentence into multiple translation calls makes it impossible to translate (each string is translated individually).
```
{% if staff %}
__("Sentence A")
{% else %}
__("Sentence B")
{% endif %}
```

portal/templates/gil/base.html
```diff
@@ -37,15 +37,17 @@
...
-        url: url,
-        cache: false,
-        success: function(response) {
-            result = $(response).find("#csrf_token");
-            $("#modal-login > div > div > div > form").append('<input id="csrf_token" name="csrf_token" type="hidden" value="'+ result.attr("value") +'">');
-        }
-    });
+      {% if not request.args.get("disableLinks", "") %}
+        var url = "{{url_for('user.login')}}";
+        $.ajax({
```
ivan-c: Is it possible to use the CSRF token directly in the template here ([eg](https://github.com/uwcirg/true_nth_usa_portal/blob/develop/portal/templates/layout.html#L132))? This was a hack I originally added when we were hosting the GIL outside the portal codebase.

achen2401:
@ivan-c Ivan, FYI, still making changes based on your comments.  hope to be done soon.

portal/templates/initial_queries.html
achen2401: Ivan, yes, it does, the code is growing.  I think I put the chunk there because of the dependency the code has on config variables, config.REQUIRED_CORE_DATA and config.CONSENT_WITH_TOP_LEVEL_ORG.  I know that breaking it out now into a separate file will break it.
I looked but don't think there is currently an API to get those values? If so, I can certainly break them out into a separate file.

portal/templates/gil/base.html
achen2401: Ivan, turns out we do currently send in the CSRF token in the form now.  (there was a round of fixes to add this back when we were patching up security risks of our system).  so I removed this ajax call - seems redundant now.

portal/templates/initial_queries_macros.html
achen2401: thank you for catching this Ivan!  I think fixed another place where it has the same issue. let me know if you find issue with my fix though.

portal/static/js/main.js
```diff
@@ -359,7 +360,7 @@ var ConsentUIHelper = function(consentItems, userId) {
             var org = OT.orgsList[item.organization_id];
             touContent += "<tr data-tou-type='" + item.type + "'>";
             touContent += "<td><span class='eproms-tou-table-text'>" + (org && hasValue(org.name) ? i18next.t(org.name) : "--") + "</span><span class='truenth-tou-table-text'>TrueNTH USA</span></td>";
-            touContent += "<td><span class='text-success small-text eproms-tou-table-text'>Agreed to <a href='" + item.agreement_url + "' target='_blank'><span class='text-capitalize'>" + i18next.t(item.display_type) + "</span></a></span><span class='text-success small-text truenth-tou-table-text'>" + i18next.t("Agreed to terms") + "</span></td>";
+            touContent += "<td><span class='text-success small-text eproms-tou-table-text'>" + i18next.t("Agreed to ") + "<a href='" + item.agreement_url + "' target='_blank'><span class='text-capitalize'>" + i18next.t(item.display_type) + "</span></a></span><span class='text-success small-text truenth-tou-table-text'>" + i18next.t("Agreed to terms") + "</span></td>";
```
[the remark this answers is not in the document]
achen2401: Ivan, I think this works --> i18next.t("Agreed to " + item.display_type) -- enclosing the whole text in the call to t.  
I think I remember why I didn't use interpolation - it seems the when I tried to transform json that contains interpolation (e.g. {
	"agree_to_text": "Agreed to {{agreement_type}}"	
})  to pot file, the pot file doesn't contain that key/value pair? 
I used the i18next tool to do the transformation.  If you look in frontend.pot now, you will see that the key/value pair is missing.  I have enclosed a custom.json that contains the interpolation.  Perhaps you can try if you have a moment?

portal/templates/initial_queries_macros.html
```diff
@@ -8,31 +8,99 @@ <h4 class="tnth-headline">{{ _("Terms of Use") }}</h4>
...
+                          <div class="display-view text-warning tnth-hide">{{_("You have previously provided your consent to the website during a visit at your treating site. If you would like a copy of this please contact your study contact.")}}<br/></div>
+                        </label>
+                         <label class="terms-label tnth-hide">
+                            <i class="fa fa-square-o fa-lg terms-tick-box"></i>
+                            {% trans privacy_url=url_for('portal.privacy', disableLinks="true"), terms_url=url_for('portal.terms_and_conditions', disableLinks="true") %}
+                              <span class="terms-tick-box-text">&nbsp;By checking this box, I confirm that I have read and accept the website <a href="{{privacy_url}}" class="required-link">privacy policy</a> and
+                                <a href="{{terms_url}}" class="required-link">terms</a></span>
```
ivan-c: Same issue as above, I don't *think* there's any translatable text below this link

portal/templates/initial_queries_macros.html
```diff
@@ -8,31 +8,99 @@ <h4 class="tnth-headline">{{ _("Terms of Use") }}</h4>
...
+                        data-agree="false"
+                        data-type="terms"
+                        data-tou-type="website terms of use"
+                        data-core-data-type="website_terms_of_use"
+                        data-url="{{tou_url}}">
+                        <i class="fa fa-square-o fa-lg terms-tick-box edit-view"></i>
+                        <span class="terms-tick-box-text">{{_("By checking this box, I confirm that I have read the information notice above and consent and agree to the processing of my personal information (including my health information) on the terms described in this Consent.") }}</span>
+                      </label>
+                      <br/>
+                      {% trans privacy_url=url_for('portal.privacy', disableLinks="true"), terms_url=url_for('portal.terms_and_conditions', disableLinks="true") %}
```
ivan-c: I think you'll want to move the `{% trans %}` block down this time, I don't think there's any English text in the `<label>` element

portal/static/js/main.js
```diff
@@ -355,11 +356,12 @@ var ConsentUIHelper = function(consentItems, userId) {
...
         //wording is not spec'd out for EPROMs. won't add anything specific until instructed
+
         this.touObj.forEach(function(item, index) {
             var org = OT.orgsList[item.organization_id];
             touContent += "<tr data-tou-type='" + item.type + "'>";
             touContent += "<td><span class='eproms-tou-table-text'>" + (org && hasValue(org.name) ? i18next.t(org.name) : "--") + "</span><span class='truenth-tou-table-text'>TrueNTH USA</span></td>";
-            touContent += "<td><span class='text-success small-text eproms-tou-table-text'>Agreed to <a href='" + item.agreement_url + "' target='_blank'><span class='text-capitalize'>" + i18next.t(item.display_type) + "</span></a></span><span class='text-success small-text truenth-tou-table-text'>" + i18next.t("Agreed to terms") + "</span></td>";
+            touContent += "<td><span class='text-success small-text eproms-tou-table-text'><a href='" + item.agreement_url + "' target='_blank'>" + i18next.t("Agreed to " + item.display_type) + "</a></span><span class='text-success small-text truenth-tou-table-text'>" + i18next.t("Agreed to terms") + "</span></td>";
```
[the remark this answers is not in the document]
achen2401: @ivan-c 
Ivan, thank you. A note here:  i18next.t("Agreed to {{ item.display_type }}") yields only "Agreed to" - "{{item.display_type}}" is ignored.  My understanding of how interpolation works in i18next is that text containing placeholder text (i.e. "Agreed to {{ item.display_type }}") actually resides in the source json file and needs to be associated with a key.  So, when we call i18next.t function in main.js, we can refer to the key/value pair by the key and inject into the function the value to be replaced accordingly.
In this case, for instance, in the source json (i.e. translation.json) there will a key/value pair {"agreed_text": "Agreed to {{item_type}}"}
And in the js code, we can call i18next.t using the key, "agreed_text", and provide the replacement text for the placeholder: i18next.t("agreed_text", {item_type: item.display_type}).  So in essence, we need to provide a custom json file containing key/translated text with placeholder that will allow interpolation to work.  To work around that, I find that i18next.t function takes broken-up strings, variable or concatenation of string or string and variable and translates them (that is why I did the concatenation here).  You can see that in debugging (i.e add queryString ?debugi18next=true to url).  I know this works a bit differently from gettext.
Sorry for the long-winded explanation - I hope that makes sense.

portal/static/js/main.js
```diff
@@ -355,11 +356,12 @@ var ConsentUIHelper = function(consentItems, userId) {
...
         //wording is not spec'd out for EPROMs. won't add anything specific until instructed
+
         this.touObj.forEach(function(item, index) {
             var org = OT.orgsList[item.organization_id];
             touContent += "<tr data-tou-type='" + item.type + "'>";
             touContent += "<td><span class='eproms-tou-table-text'>" + (org && hasValue(org.name) ? i18next.t(org.name) : "--") + "</span><span class='truenth-tou-table-text'>TrueNTH USA</span></td>";
-            touContent += "<td><span class='text-success small-text eproms-tou-table-text'>Agreed to <a href='" + item.agreement_url + "' target='_blank'><span class='text-capitalize'>" + i18next.t(item.display_type) + "</span></a></span><span class='text-success small-text truenth-tou-table-text'>" + i18next.t("Agreed to terms") + "</span></td>";
+            touContent += "<td><span class='text-success small-text eproms-tou-table-text'><a href='" + item.agreement_url + "' target='_blank'>" + i18next.t("Agreed to " + capitalize(item.display_type)) + "</a></span><span class='text-success small-text truenth-tou-table-text'>" + i18next.t("Agreed to terms") + "</span></td>";
```
ivan-c: I don't think you can pass dynamically-generated strings to `i18next.t()` without string formatting/interpolation. 

`capitalize()` won't be called until after `i18next-extraction` is run, so `item.display_type` won't be included in the POT file. Using interpolation allows you to substitute the value in *after* the string is translated by `i18next.t()`, so it should appear correctly in the translated text.

ivan-c:
Thanks for talking through the i18next issues you found! :)

PR #1522, opened 2017-10-17
TN-156 TN-263 display recipient types on form; pull recipient for msg

OptimusRhine:
https://jira.movember.com/browse/TN-261

* added `types` as input array on `contact.html` (both gil and non-gil) template render
  * pulls in any `Organization.email` with a valid email address
* added 'Enquiry Type' dropdown to `contact.html` (both gil and non-gil)
  * only appears if `types > 0` (aka, if there are actual options to choose from besides just the default
  * always includes, and starts selected as, 'Other' option (which is the same as the configured default)
* modified `/contact [POST]` to grab the recipient from the `type` field on form from `contact.html` (or, if nothing found, use the configured default)
  * throws an error if no `type` value found in the form AND no configured default value
  * only uses `type` from form, if it matches an existing `Organization.email` (otherwise, uses default)

### DIFF
--- a/portal/views/portal.py
+++ b/portal/views/portal.py
@@ -925,7 +925,7 @@ def contact():
         email = user.email if user else ''
         gil = current_app.config.get('GIL')
         recipient_types = []
-        for org in Organization.query.filter(Organization.email != None):
+        for org in Organization.query.filter(Organization.email.isnot(None)):
             if u'@' in org.email:
                 recipient_types.append((org.name, org.email))
         return render_template('contact.html' if not gil else 'gil/contact.html',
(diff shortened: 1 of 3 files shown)


portal/views/portal.py
```diff
@@ -948,11 +953,14 @@ def contact():
         abort(400, "No contact request body provided")
     body = u"From: {sendername}<br />Email: {sender}<br /><br />{body}".format(
         sendername=sendername, sender=sender, body=formbody)
-    recipients = current_app.config['CONTACT_SENDTO_EMAIL']
+    recipient = request.form.get('type')
```
pbugni: Hmm - this worries me.

Perhaps this sounds overly paranoid... but I know contact forms are frequently used to generate spam - seems like we're now opening up an email gateway, if we blindly take the recipient as a form variable.  I'd suggest either validating it's a known email address, or passing something used to look up the correct contact.

OptimusRhine: Ah, that's a good point. I'll validate that the email actually belongs to an org.

OptimusRhine: Modified the logic; now it only uses the recipient from the form if it matches an existing `Organization.email`; otherwise, it just always uses the default from the config

PR #1525, opened 2017-10-18
TN-124 TN-249 Store UI table sort order/filter values

OptimusRhine:
https://jira.movember.com/browse/TN-249

* added new TablePreference object
  * stores preferences regarding filtering & storing for given User (`user_id`) and UI Table (`table_name`)
    * sort field and order stored individually
      * order must be 'asc' or 'desc'
    * filter values stored in dict/json form (e.g. `{"<field_to_filter>": "<value_to_filter_on>"}`)
  * combination of `user_id` and `table_name` must be unique
  * includes `as_json()` and `from_json()` methods
    * `TablePreference.from_json()` only updates fields present in incoming json data, but it DOES update with blank/`None` vals (if users want to blank out their sort field, etc)
* added table_preferences table db migration
* added TablePreference API endpoints
  * `[GET]` and `[PUT]/[POST]` endpoints both located at `/api/user/<user_id>/table_preferences/<table_name>`
  * `[PUT]/[POST]` will create if user/table entry doesn't exist, otherwise just update
* added unit tests for API endpoints (and subsequently also for model functions)
* fixed issues with `test_site_persistence` and `test_scheduled_job`
  * unrelated to the actual change here, but the existence of the new tests caused the erroneous code in `test_site_persistence` to fail non-silently, and fixing that caused the erroneous code in `test_scheduled_job` to fail non-silently.


OptimusRhine:
Hooray, test issues (with the other, pre-existing tests) finally fixed :)

@pbugni This should be ready to merge

PR #1527, opened 2017-10-19
preserve column sorting in patients list, frontend

achen2401:
address TN-248
https://jira.movember.com/browse/TN-249

preserving column sorting in patients list 

CANNOT MERGE YET UNTIL PR #1525  
is merged.

@OptimusRhine Adam, if you have a minute, do you mind having a look at my changes?
I decided that calling API on the "on sort" event seems safer than calling it when the browser page is unloaded.  I found that the call is not reliably made when window is unloaded in some browsers (e.g. in IE).


achen2401:
@OptimusRhine Adam, thank you for looking at my changes. I had forgotten about the filter fields - I had meant to work on those and then was distracted.  I will add the changes.

achen2401:
@OptimusRhine Adam, I added the changes to preserve the filtering selection.  Be nice if you can have another look?

portal/static/js/admin.js
```diff
@@ -421,7 +421,91 @@ AdminTool.prototype.handleDownloadModal = function() {
...
+    	//get selected sorted field information on UI
+    	var sortedField = $("#adminTable th[data-field]").has(".sortable.desc, .sortable.asc");
+    	if (sortedField.length > 0) {
+    		data["sort_field"] = sortedField.attr("data-field");
+    		var sortedOrder = "desc";
+    		sortedField.find(".sortable").each(function() {
+    			if ($(this).hasClass("desc")) sortedOrder = "desc";
+    			else if ($(this).hasClass("asc")) sortedOrder = "asc";
+    		});
+    		data["sort_order"] = sortedOrder;
```
ivan-c: @OptimusRhine Could this cause a XSS if someone initially set their sorting preference for a column to be `alert("foo");`?

OptimusRhine: Well, I don't think this is an issue with the sorting column/order, because on the backend, we only store sort_order values of 'asc' or 'desc' (anything else throws an error)

This COULD come up with the filter fields (basically, someone type `alert("foo")` into the filter field for a given column, and we'd display it later).

However, (a) I don't know if this is actually possible (depends on how those filter fields are called/rendered/etc), and (b) even if it is, it would only affect that user who inputted that value in the first place. I think XSS attacks are only considered dangerous, if a script one user puts in could influence what's displayed/rendered for another user (like by putting scripting into their username, which shows up in patient lists, etc). So even if it's possible here, I would think it's not a worry, as no user would be influenced besides the one generating the script in the first place...

ivan-c: >only affect that user who inputted that value in the first place
Good point, had forgotten these would only be displayed for the associated user

>depends on how those filter fields are called/rendered/etc
Always keep in mind that bad guys can bypass the front-end validation entirely :)

PR #1530, opened 2017-10-20
TN-236 TN-297 User EmailMessages API

OptimusRhine:
https://jira.movember.com/browse/TN-297

* added `as_json()` method for EmailMessages
* added GET API endpoint for messages sent to a given user (`/api/user/<user_id>/messages`)
  * only accessible by those who should see the table in the UI - that is, Staff, InterventionStaff, and Admins
* added unit tests for above endpoint

### DIFF
--- a/portal/views/user.py
+++ b/portal/views/user.py
@@ -14,6 +14,7 @@ from ..models.audit import Audit
 from ..models.auth import Client, Token
 from ..models.group import Group
 from ..models.intervention import Intervention
+from ..models.message import EmailMessage
 from ..models.organization import Organization
 from ..models.role import ROLE, Role
 from ..models.relationship import Relationship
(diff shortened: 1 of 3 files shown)


OptimusRhine:
@achen2401 will this work for you, for TN-260?

achen2401:
@OptimusRhine 
adam, I think so! The API returns all the necessary fields I need for rendering the UI.

portal/views/user.py
```diff
@@ -1887,3 +1888,73 @@ def set_table_preferences(user_id, table_name):
...
+    user = current_user()
+    if user.id != user_id:
+        current_user().check_role(permission='view', other_id=user_id)
+        user = get_user(user_id)
+    if user.deleted:
+        abort(400, "deleted user - operation not permitted")
+
+    messages = []
+
+    for em in EmailMessage.query.all():
```
pbugni: This looks rather inefficient.  Any reason we can't query directly for the matching set of messages?

OptimusRhine: @pbugni I went with this route, because of a couple issues I bumped up with when testing:
* We want to look for `EmailMessages` where the `recipients` list contains the user email address exactly. But we can't just query where `recipients contains user.email` or `recipients like %user.email%, because that could get some inexact matches (for example, if `EmailMessages.recipients="blorp@gmail.com"` and `user.email="orp@gmail.com"`, we'd get a false match). So the matching there is fairly complex, if it's in-query
* `user.email` has different behavior in-query, than it does in straight variables. For example, say `user._email="__invite__blorp@gmail.com"`. When `user.email` is called directly (as I'm doing currently), we get `blorp@gmail.com` (which is what we want). However, when it's used in a sqlalchemy query, it's pulling the whole `__invite__blorp@gmail.com`. I mentioned this a few weeks ago in [this PT story comment thread](https://www.pivotaltracker.com/story/show/151111342). Something we should address at some point, but in the meantime, we need to be wary of using email as a param in sqlalchemy queries.

pbugni: @OptimusRhine fair enough.

Can you add a comment in the code, so we don't foolishly *optimize* those problems in

PR #1546, opened 2017-10-24
exercise and diet blueprint

pbugni:
Imported demo code as provided by CHCR, made some minor modifications to convert it to a valid flask blueprint, and modified system to use it.

NB - early demo code, needs styling, etc., but it does provide working example of all systems involved (LR, tags, blueprints, etc.)


pbugni:
Merging now, for demo purposes.

PR #1547, opened 2017-10-24
fix exercse and diet link in GIL about page

achen2401:
fix broken exercise and diet link in GIL About page


pbugni:
Thank you @achen2401

PR #1549, opened 2017-10-25
TN-136 Weekly site summary email

OptimusRhine:
https://jira.movember.com/browse/TN-136

* added `overdue_stats_by_org()` reporting function, which gets a list of (anonymous) overdue questionnaires, stored by days overdue and sorted by org
* added `generate_overdue_table_html()` reporting function, which generates an html table of overdue questionnaire stats
  * separate from `overdue_stats_by_org()`, because (in the email summaries) the table is generated on a per-user basis, but we only ever want to pull the master list of stats once
  * only shows orgs the given user has access to (that user's orgs, and any child orgs of those orgs)
  * if `top_org` is provided, also only shows orgs that are within the `top_org`'s org tree
  * takes a `cutoff_days` param, which dictates the split of columns
    * e.g. `cutoff_days=[7,14,21]` will result in columns of '1-7 Days', '8-14 Days', & '15-21 Days' (with the data split correctly between them)
  * includes per-org total column, and overall totals row
* added new celery task, to generate a site summary email
  * takes `top_org` and `cutoff_days` params (per the table generation function above)
  * only sends emails to Staff that belong to the given org (or any child orgs)
  * pulls table html from reporting function, pulls other email content from LR
* added table styling to default EmailMessage styling, tightened up restricted tag check
* added SiteSummaryEmail_ATMA AppTextModelAdapter
* added unit tests


OptimusRhine:
@achen2401 if you have some time, could you take a quick look at the styling changes I made in `portal.models.message`, for the default table styling in emails? What I have works as a proof of concept, but you may have some ideas of ways to make those tables look better. Thanks!

tests/test_reporting.py
```diff
@@ -73,3 +78,76 @@ def test_reporting_stats(self):
...
+            name='CRV', organization_id=crv.id,
+            start='{"days": 1}',
+            overdue='{"days": 2}',
+            expired='{"days": 90}')
+        qbq = QuestionnaireBankQuestionnaire(
+            questionnaire_id=epic26.id,
+            rank=0)
+        bank.questionnaires.append(qbq)
+
+        self.test_user.organizations.append(crv)
```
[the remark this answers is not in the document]
OptimusRhine: ...but `self.test_user` already exists from `__init__`...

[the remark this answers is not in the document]
OptimusRhine: @ivan-c Is there any way to have codacy ignore the unit test code? Besides false-positive "issues" like the one here, it looks like it may also interpret a lot of "clones" in the tests as well (because we repeat similar steps in different tests, etc).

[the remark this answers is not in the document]
ivan-c: @OptimusRhine Sorry about that- I thought Codacy kept everything separate (no PR comments etc). Let me disable checking the `tests/` dir but please let me know if it's nagging is too much and I'll turn it off.

portal/models/message.py
```diff
@@ -25,7 +25,31 @@ def log_message(message, app):
     "<html><head><title>TrueNTH email</title><style>"
     "body {"
     " font-size: 16px;"
-    "}"
+    "} "
```
ivan-c: Getting a bit long there, think it makes sense to move into a separate jinja template yet?

portal/models/reporting.py
```diff
@@ -82,3 +86,89 @@ def get_reporting_stats():
...
+        if user.has_role(ROLE.TEST):
+            continue
+        overdue = calculate_days_overdue(user)
+        if overdue > 0:
+            for org in user.organizations:
+                overdue_stats[org].append(overdue)
+    return overdue_stats
+
+
+def generate_overdue_table_html(cutoff_days, overdue_stats, user, top_org):
```
ivan-c: This might work better as a jinja template, especially if it gets anymore complicated. Concatenating HTML fragments can make it difficult to get a sense of the overall structure of something and is harder to maintain/extend.

portal/tasks.py
```diff
@@ -269,6 +273,56 @@ def send_user_messages(email, force_update=False):
...
+def generate_and_send_summaries(cutoff_days, org):
+    ostats = overdue_stats_by_org()
+    cutoffs = [int(i) for i in cutoff_days.split(',')]
+
+    ot = OrgTree()
+    top_org = Organization.query.filter_by(name=org).first()
+    if not top_org:
+        raise ValueError("No org with name {} found.".format(top_org))
+
+    if top_org:
```
ivan-c: Won't this always be `True` (otherwise above `ValueError` will be thrown)?

OptimusRhine: Yep, good point

portal/models/reporting.py
OptimusRhine: Yea, it got bloated as I added onto it. Moving it out to a jinja template now.

portal/models/message.py
OptimusRhine: Yea, I'll move this out to a jinja template as well.

OptimusRhine: Actually, not only does this cause the job to run slower (as we have to pass a whole bunch of generated email html through a `render_template` call), but I'm having issues with the styling getting dropped when it's moved out to a template.

I'm moving the reporting table out to a template, but I think we can just leave this emailmessage styling here for now (it's all in one chunk, like the intervention_strategies html, so it's still easily readable/understandable). If it gets more bloated, we can look at ways to abstract it out to a template.

PR #1558, opened 2017-10-26
TN-327 Fixing ePROMs terms

OptimusRhine:
https://jira.movember.com/browse/TN-327

* brought back in code that got dropped during the blueprint refactoring, so that when logged in, ePROMs users see `/about` content that's specific to their user and org (if Patient or Staff)

### DIFF
--- a/portal/eproms/views.py
+++ b/portal/eproms/views.py
@@ -139,7 +139,19 @@ def privacy():
 def terms_and_conditions():
     """ terms-and-conditions of use page"""
     user = current_user()
-    terms = VersionedResource(app_text(Terms_ATMA.name_key()))
+    if user:
+        organization = user.first_top_organization()
+        role = None
+        for r in (ROLE.STAFF, ROLE.PATIENT):
+            if user.has_role(r) and not role:
+                role = r
+        if not all((role, organization)):
+            role, organization = None, None
+
+        terms = VersionedResource(app_text(Terms_ATMA.name_key(
+            role=role, organization=organization)))
+    else:
+        terms = VersionedResource(app_text(Terms_ATMA.name_key()))
     return render_template(
         'eproms/terms.html', content=terms.asset, editorUrl=terms.editor_url, user=user)
 


portal/eproms/views.py
```diff
@@ -139,7 +139,19 @@ def privacy():
...
     """ terms-and-conditions of use page"""
     user = current_user()
-    terms = VersionedResource(app_text(Terms_ATMA.name_key()))
+    if user:
+        organization = user.first_top_organization()
+        role = None
+        for r in (ROLE.STAFF, ROLE.PATIENT):
+            if user.has_role(r) and not role:
+                role = r
+        if not all((role, organization)):
```
ivan-c: Is this necessary? shouldn't `first_top_organization()` return something false-y enough to be `None`?

portal/eproms/views.py
```diff
@@ -139,7 +139,19 @@ def privacy():
 def terms_and_conditions():
     """ terms-and-conditions of use page"""
     user = current_user()
-    terms = VersionedResource(app_text(Terms_ATMA.name_key()))
+    if user:
```
ivan-c: Should this logic be moved to the `VersionedResource` model? Seems it would be useful elsewhere (than this one view)

portal/eproms/views.py
OptimusRhine: Right, but if a `role` isn't found, then we don't want to send a value for `organization` either, and vice versa (because with `Terms_ATMA()`, if you have a null value for one param and a non-null value for the other, it throws an error; it only returns a value when either both params have values, or neither do)

portal/eproms/views.py
OptimusRhine: It kind of is (in that `Terms_ATMA()` returns one thing if `kwargs.get('organization') and kwargs.get('role')`, and another if not), but I don't think we want the actual org and role lookup logic in there...at least, it seems like Paul's been intentionally avoiding any sort of actual object query logic in those AppTextModelAdapters

portal/eproms/views.py
OptimusRhine: Actually, we COULD change `Terms_ATMA()` to never raise `ValueError`s, and just only return the org/role formatted terms if both params have values (otherwise return the non-formatted terms). Depends on if there's a good reason to raise those `ValueError`s...

ivan-c: >change Terms_ATMA() to never raise ValueErrors, and just only return the org/role formatted terms if both params have values
Sorry, this is what I was suggesting, just not sure how good of an idea it was

portal/eproms/views.py
ivan-c: Sorry I wasn't implying something as big as moving the org logic into any other model; you correctly guessed my vague suggestion in your other comment :)

PR #1565, opened 2017-10-30
TN-136 Minor fixes for site summary email

OptimusRhine:
https://jira.movember.com/browse/TN-136

* add explicit check for `user.email` (for cases when `user._email=<just a prefix>`, so `user.email=None`)
* change 'multiple QB' error logging to warning logging (so we're not spammed with emails about it)

### DIFF
--- a/portal/models/questionnaire_bank.py
+++ b/portal/models/questionnaire_bank.py
@@ -260,12 +260,17 @@ class QuestionnaireBank(db.Model):
         def validate_classification_count(qbs):
             if qbs and qbs[0].classification == 'recurring':
                 return
-            if len(qbs) > 1:
-                current_app.logger.warn(
-                    "multiple QuestionnaireBanks for {user} with "
-                    "{classification} found.  The UI won't correctly display "
-                    "more than one at this time.".format(
-                        user=user, classification=classification))
+            if (len(qbs) > 1):
+                errstr = ("multiple QuestionnaireBanks for {user} with "
+                          "{classification} found.  The UI won't correctly "
+                          "display more than one at this "
+                          "time.").format(user=user,
+                                          classification=classification)
+                systype = current_app.config.get('SYSTEM_TYPE', '').lower()
+                if systype == 'production':
+                    current_app.logger.error(errstr)
+                else:
+                    current_app.logger.warn(errstr)
 
         validate_classification_count(results)
         return results
--- a/portal/models/questionnaire_bank.py
+++ b/portal/models/questionnaire_bank.py
@@ -261,7 +261,7 @@ class QuestionnaireBank(db.Model):
             if qbs and qbs[0].classification == 'recurring':
                 return
             if len(qbs) > 1:
-                current_app.logger.error(
+                current_app.logger.warn(
                     "multiple QuestionnaireBanks for {user} with "
                     "{classification} found.  The UI won't correctly display "
                     "more than one at this time.".format(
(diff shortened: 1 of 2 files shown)


portal/tasks.py
```diff
@@ -304,7 +304,7 @@ def generate_and_send_summaries(cutoff_days, org_id):
     name_key = SiteSummaryEmail_ATMA.name_key(org=top_org.name)
 
     for user in User.query.filter_by(deleted_id=None).all():
-        if (user.has_role(ROLE.STAFF) and (u'@' in user.email)
+        if (user.has_role(ROLE.STAFF) and user.email and (u'@' in user.email)
```
ivan-c: Why not move those filters into the query?

OptimusRhine: * The `user.has_role(ROLE.STAFF)` filter _could_ be moved to the query, but it would make it way more complex (joining three tables, etc). Figured since we already had the check here, it'd be simpler code to just included it here.
* The `(top_org in ot.find_top_level_org(user.organizations))` can't be moved to the query
* The email checks act differently in sqlalchemy vs being called directly; that is, `user.email` does the whole prefix-removal-thing when called directly, but when called in a sqlalchemy query filter, it actually just returns `user._email`. I mentioned it to Paul [in this PR](https://github.com/uwcirg/true_nth_usa_portal/pull/1530#discussion_r146428689) and [in this PT story](https://www.pivotaltracker.com/n/projects/1225464/stories/151111342), but it has something to do with how we (apparently improperly) set up the hybrid property there. In any case, would rather just handle that here for now, until that property gets properly fixed.

OptimusRhine: Actually, thinking more about this - for the `user.email` checks, it shouldn't matter if we're included the prefixes or not (at least, for this initial check that the email address includes '@'; obviously, we don't want the prefixes when we actually pull the `user.email` value to use as the `EmailMessage.recipient`). So you're right @ivan-c , that can probably be moved into the query. I'll make that change now.

portal/models/questionnaire_bank.py
```diff
@@ -261,7 +261,7 @@ def validate_classification_count(qbs):
             if qbs and qbs[0].classification == 'recurring':
                 return
             if len(qbs) > 1:
-                current_app.logger.error(
+                current_app.logger.warn(
```
ivan-c: Would it be easy to condition this on `ROLE.TEST`? 

I feel like these would be valuable warnings for real patients, and would "encourage" us to properly flag out test patients as such :)

portal/tasks.py
ivan-c: >user.email
Thanks, forgot that was a method underneath and not actually in the DB

OptimusRhine: ACTUALLY actually, we do still want to check that `user.email is not None`, because in the User model, if `user._email` starts with the no_email prefix (e.g. `user._email == '__no_email__test@gmail.com'`), then `user.email` returns None. So it'd be possible that in the sqlalchemy query, the check to see if `user.email contains '@'` returns True (because it actually checks against `user._email`), but then we'd get errors when we try to actually send the EmaiMessage.

Long story short, gonna keep this as is for now :)

portal/models/questionnaire_bank.py
OptimusRhine: For one, in our non-prod instances, this is true for LOTS of users. Unless we want to flag almost all users in non-prod instances with `ROLE.TEST`, we're going to keep getting a _lot_ of noise from all those instances. For instance, with stg-eproms alone, one run of the site summary email process for just one of the top orgs generated 38 emails for this error, and it would have been more if it hadn't errored out (because of the `user.email` issue)

For two, with prod instances, how likely is it that we'll see this error, and proactively fix it (esp for real patients)? Seems more likely something to help troubleshoot, if a user's asking why the wrong QB is showing up in their UI.

OptimusRhine: (per Slack convo, going to keep this as `logger.error`, but only log in production)

PR #1566, opened 2017-10-30
orgs filtering fix in patients list

achen2401:
fixes per conversation with Justin and address issues found related to this story: https://jira.movember.com/browse/TN-124

- **remove** select filter for organization column
- preserve filters based on selection(s) from filter list control

@OptimusRhine Adam, it would be nice if you can have a look and see if my changes in patients.py are okay?  I figure it is easier to get filtered rows for selected organization(s) from the server side rather than client side

### DIFF
--- a/portal/views/patients.py
+++ b/portal/views/patients.py
@@ -56,10 +56,8 @@ def patients_root():
         # check user table preference for organization filters
         pref = TablePreference.query.filter_by(table_name='patientList',
                                                user_id=user.id).first()
-        if pref:
-            pref_data = pref.as_json()
-            if ('filters' in pref_data) and ('orgs_filter_control' in pref_data['filters']):
-                pref_org_list = pref_data['filters']['orgs_filter_control']
+        if pref and pref.filters:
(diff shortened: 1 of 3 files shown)


portal/views/patients.py
```diff
@@ -51,16 +52,24 @@ def patients_root():
     consented_users = [u.user_id for u in consent_query]
 
     if user.has_role(ROLE.STAFF):
-        request_org_list = request.args.get('org_list', None)
+        pref_org_list = None
+        # check user table preference for organization filters
+        pref = TablePreference.query.filter_by(table_name='patientList',
+                                               user_id=user.id).first()
+        if pref:
+            pref_data = pref.as_json()
```
OptimusRhine: This could be done without converting the pref to json, by just pulling the attributes from `pref`. For example:
```
if pref and pref.filters:
   pref_org_list = pref.filters.get('orgs_filter_control') # will return None if 'orgs_filter_control' isn't in pref.filters
```

PR #1581, opened 2017-11-01
Feature/patients list column selections

achen2401:
address this story:  https://jira.movember.com/browse/TN-347

- preserving user's column selection(s) on patients list
- add link to patient profile page per row instead of per cell as a column can be hidden now
- styling fixes, including fixing erroneous reference to source file in source map files


achen2401:
trying a better approach, closing for now

PR #1582, opened 2017-11-02
TN-294 Org timezones & Patient timezone assignment

OptimusRhine:
https://jira.movember.com/browse/TN-294

* added `_timezone` field to Orgs
  * accessible through `timezone` property/setter
    * property recursively inherits from parents if `not self._timezone`; defaults to 'UTC' if no parents have TZs
  * accessible in `as_fhir()` and `from_fhir()` methods as an Extension
    * URL: `http://hl7.org/fhir/StructureDefinition/user-timezone`
  * includes migration
* added logic to account creation (`[POST] /account`), where if the incoming user JSON contains one and only one (valid) Org, use that Org's TZ for the user's TZ
* added robust unit tests for Org TZ fields/logic/inheritance, and for Org->User TZ inheritance on account creation


OptimusRhine:
@pbugni I'm not sure `UserRaceExtension()` is related. Do you mean `UserTimezone()` and `OrganizationTimezone()`? In that case, they do share some code, but also use different base attribs (`org` vs `user`) that I think would prevent much of it from being moved to a parent class.

portal/models/extension.py
```diff
@@ -55,3 +58,36 @@ def apply_fhir(self):
         # Remove the stale concepts that weren't requested again
         for concept in remove_if_not_requested.values():
             self.children.remove(concept)
+
+
+class TimezoneExtension(CCExtension):
+    def __init__(self, org, extension):
```
pbugni: can we `s/org/source/`?

OptimusRhine: Ack, yea, slipped through

PR #1584, opened 2017-11-02
Allow build jobs to fail

ivan-c:
Allow build jobs to fail until I can figure the best way to clone to an exact commit

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -55,9 +55,9 @@ matrix:
       addons:
         postgresql: "9.4"
         sauce_connect: true
-  # Build artifacts only (no testing)
   allow_failures:
     - env: TOXENV=ui
+    - env: TOXENV=build
 
 before_install:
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -55,9 +55,9 @@ matrix:
       addons:
         postgresql: "9.4"
         sauce_connect: true
-  # Build artifacts only (no testing)
   allow_failures:
     - env: TOXENV=ui
+    - env: TOXENV=build
```
pbugni: `build` is a little ambiguous - can we rename that target (line 50 iiuc) to say `docker_build`?

ivan-c: what about `build_artifacts` - technically we're building a deb package *and* a docker image

pbugni: great - i think that's more clear.

pbugni:
sure - thanks.

PR #1587, opened 2017-11-03
add treatments/procedures section in account creation/profile for eproms

achen2401:
address this story:  https://jira.movember.com/browse/TN-185

- expose treatments/procedures section in account creation page for CRV staff in EPROMs
- hide section appropriately in Truenth and for IRONMAN staff via css
- moving JS code for account creation to a standalone JS file
- expose config API to allow access for frontend JS code
- styling fixes

NOTE: this needs to go out in conjunction to PR for EPROMs persistence file: https://github.com/uwcirg/ePROMs-site-config/pull/86


portal/static/js/accountCreation.js
```diff
@@ -0,0 +1,572 @@
...
+    };
+    this.__setDemo = function(returnedData) {
+        var responseData = returnedData && returnedData.data? returnedData.data : null;
+        var self = this;
+        if (responseData) {
+
+            self.userId = responseData["user_id"];
+
+            if (isNaN(self.userId)) {
+                self.__handleError("Invalid user id: " + self.userId);
```
ivan-c: Tag for translation please :)
You also might wanna give this a try- my best attempt at interpolation (working around i18next):
```
self.__handleError(i18next.t("Invalid user id: %d").replace("%d", self.userId));
```

portal/static/js/accountCreation.js
```diff
@@ -0,0 +1,572 @@
...
+        var self = this;
+
+        var treatmentRows = $("#pastTreatmentsContainer tr[data-code]");
+        if (treatmentRows.length == 0) {
+            self.__handleDisplay();
+            return false;
+        };
+
+        if (isNaN(self.userId)) {
+            self.__handleError("Invalid user id: " + self.userId);
```
ivan-c: Tag for translation please :)

portal/static/js/main.js
```diff
@@ -1172,15 +1172,24 @@ var fillContent = {
...
                 else if (creator == subjectId) {
                     creator = i18next.t("this patient");
                 }
                 else creator = i18next.t("staff member") + ", <span class='creator'>" + (hasValue(creatorDisplay) ? creatorDisplay: creator) + "</span>, ";
                 var dtEdited = val.resource.meta.lastUpdated;
                 dateEdited = new Date(dtEdited);
-                contentHTML += "<tr data-id='" + procID + "' data-code='" + code + "'><td width='1%' valign='top' class='list-cell'>&#9679;</td><td class='col-md-8 col-xs-8' valign='top'>" + (cPerformDate?cPerformDate:performedDate) + "&nbsp;--&nbsp;" + displayText + "&nbsp;<em>(" + i18next.t("data entered by ") + creator + i18next.t(" on ") + dateEdited.toLocaleDateString('en-GB', {day: 'numeric', month: 'short', year: 'numeric'}) + ")</em></td><td class='col-md-4 col-xs-4 lastCell text-left' valign='top'>&nbsp;" + deleteInvocation + "</td></tr>";
+
+                var creatorText = i18next.t("(data entered by {})");
+                var createDateText = i18next.t(" on {}");
```
ivan-c: You can't split sentences into multiple `t()` calls. You'll want to do something like this:
```
create_date = dateEdited.toLocaleDateString('en-GB', {day: 'numeric', month: 'short', year: 'numeric'}));

i18next.t("(data entered by {actor}) on {create_date}").replace("{actor}", creator).replace("{create_date}", create_date)
```

portal/static/js/main.js
```diff
@@ -1207,7 +1216,7 @@ var fillContent = {
 
         // If newEntry, then add icon to what we just added
         if (newEntry) {
-            $("#eventListtnthproc").find("tr[data-id='" + highestId + "'] td.lastCell").append("&nbsp; <small class='text-success'><i class='fa fa-check-square-o'></i> <em>" + i18next.t("Added") + "!</em></small>");
+            $("#eventListtnthproc").find("tr[data-id='" + highestId + "'] td.descriptionCell").append("&nbsp; <small class='text-success'><i class='fa fa-check-square-o'></i> <em>" + i18next.t("Added") + "!</em></small>");
```
ivan-c: Technically, I think the "!" should be inside the "Added"

portal/static/js/accountCreation.js
```diff
@@ -0,0 +1,572 @@
...
+            });
+        } else {
+            //Display Error Here
+            if (returnedData.error) {
+               this.__handleError(returnedData.error);
+               this.__handleButton();
+            };
+        };
+    };
+    this.__setProcedures = function() {
```
ivan-c: Should this be moved to `procedures.js`?

portal/static/js/accountCreation.js
```diff
@@ -0,0 +1,572 @@
...
+            $("#email").val("").attr("disabled", true).removeAttr("required").removeAttr("data-customemail");
+            setTimeout('$("#erroremail").text(""); $("#email").closest("form-group").removeClass("has-error"); $("#email").css("border-color", "#ccc");', 600);
+        }
+        else {
+            $("#erroremail").css("visibility", "visible");
+            $("#email").attr("disabled", false).attr("required", "required").attr("data-customemail", "true");
+        };
+    });
+
+    function getDateWithTimeZone(dObj) {
```
ivan-c: Should this be moved to a more general file (`main.js`)?

portal/views/portal.py
```diff
@@ -707,8 +707,8 @@ def settings():
 @oauth.require_oauth()
 def config_settings(config_key):
     key = config_key.upper()
-    # Only handing out LifeRay keys at this time
-    if key.startswith('LR_'):
+    # check key existence in application config
+    if key in current_app.config:
```
ivan-c: This is a security vulnerability since we store some secrets as config values :) (`SECRET_KEY`, `SQLALCHEMY_DATABASE_URI`, etc)

Is there a pattern in the config values you need so we can use something like `LR_` to allow access to some? 
If it's only a handful (less than ~10) it's probably fine to list them all.

ivan-c:
Also, let me know if any of the @codacy-bot comments are annoying or inaccurate! :)

portal/static/js/accountCreation.js
achen2401: @ivan, I didn't fix this here because there is code in this method that acts on the UI specific to account creation page - hard to make it work if the method is moved to procedures.js.  Otherwise, I addressed your comments.   I didn't address some of issues raised by codacy-bot - e.g. hasValue not defined, etc.  as the undefined functions were actually defined in main.js (I did have a comment in the js file that there is a dependency on main.js)

achen2401: Also, Ivan, let me know if you see any issue with my latest changes.

portal/views/portal.py
achen2401: @ivan Ivan, didn't think of that.  thank you for catching this!  I added a check for "consent_" prefix config variables (e.g. CONSENT_WITH_TOP_LEVEL_ORG, COSENT__PERMISSIBLE_ROLES) that are accessed by the front end currently.  There may be others but those are the ones needed by the accountCreation.js.

portal/static/js/accountCreation.js
```diff
@@ -0,0 +1,576 @@
+/*
+ * dependency: JQuery, main.js
```
ivan-c: Is there a programmatic way to do that with NPM or a more appropriate framework that makes sense? Should this include i18next (I see it calls `i18next.t()`)?

portal/static/js/accountCreation.js
ivan-c: Ah, thanks for the explanation :)

ivan-c: There isn't a way to pass those variables between files, is there (maybe as values to functions)? Seems like there should be a more robust way to know you won't have undefined variables.

achen2401: @ivan-c ivan, yeah, was thinking over the dependency issue here.. thinking to treat each JS file as a module and load each when needed via modular script loader like RequireJS (or CommonJS).  I will do more research into this.

portal/static/js/accountCreation.js
achen2401: @ivan-c ivan, yeah, my previous comment alluded to a third party modular script loader, or a nodeJS app, which I think follows commonJS module system.

PR #1589, opened 2017-11-03
TN-335 Move Patient invite email to API GET

OptimusRhine:
https://jira.movember.com/browse/TN-335

* added API for generating Patient invite email content (only accessible to Staff, StaffAdmin, and Admin)
* modified profile logic
  * no longer generates invite_email content before loading patient profile, which means...
    * invite_email errors won't prevent the profile from loading
    * the profile won't need to wait for the content to be pulled from LR and generated before loading
    * the template code is _much_ cleaner (no longer need to pass the invite_email param through 10 different macros)
  * now just makes a separate API call to new API endpoint when button is clicked (same as we're already doing for Staff registration emails)

### DIFF
--- a/portal/templates/profile.html
+++ b/portal/templates/profile.html
@@ -69,7 +69,7 @@
         {%- if user.id == current_user.id -%}
           {{user_profile(user, current_user, consent_agreements, user_interventions)}}
         {%- else -%}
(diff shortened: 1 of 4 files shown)


OptimusRhine:
Thanks Amy!

PR #1590, opened 2017-11-03
TN-348 Fix email HTML formatting bug

OptimusRhine:
https://jira.movember.com/browse/TN-348

* removed extraneous `<!DOCTYPE html><html>` from header and `</html>` from footer in email body
  * flask_mail Message (which we convert our EmailMessages to) already adds outside html tags to the body. Our current setup was causing the tags to get repeated, which was causing formatting issues within the email
  * tested to confirm that (a) the bug was fixed (see TN-348 for full description of issue), and (b) the rest of the email rendered as expected
* updated unit tests to reflect lack of declaration in `EmailMessage.body`

### DIFF
--- a/portal/models/message.py
+++ b/portal/models/message.py
@@ -21,8 +21,7 @@ email_dispatched.connect(log_message)
 
 
 EMAIL_HEADER = (
-    u"<!DOCTYPE html>"
-    "<html><head><title>TrueNTH email</title><style>"
+    u"<head><title>TrueNTH email</title><style>"
     "body {"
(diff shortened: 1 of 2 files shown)


achen2401:
@OptimusRhine adam, I will test. 
@pbugni @ivan-c 
paul and ivan, I wonder if you can wait to merge this.  I want to test this out a bit.

OptimusRhine:
@achen2401 Yep, just pulled and tested again. Looks great to me. Thanks for your help with this!

PR #1591, opened 2017-11-07
admin table filtering fixes

achen2401:
found this while testing - filtering by org(s) not working correctly in admin page

fixes include:
- add table identifier information to allow flexibility in saving table preference
- add table preference filtering information to admin view
- refactor admin.js to include table identifier code


achen2401:
i think i address most of the issues raised by codacy-bot.  if there are more, i will fix in the future PR.  
let me know if you guys see any issue with my changes.

portal/views/portal.py
```diff
@@ -429,15 +430,22 @@ def admin():
...
 
-    request_org_list = request.args.get('org_list', None)
+    user = current_user()
+
+    pref_org_list = None
+    # check user table preference for organization filters
+    pref = TablePreference.query.filter_by(table_name='adminList',
+                                           user_id=user.id).first()
+    if pref and pref.filters:
+        pref_org_list = pref.filters.get('orgs_filter_control')
```
ivan-c: You can remove the below condition (`if pref_org_list`) and un-indent the remaining code if you set the default `pref_org_list` to an empty string by default.
eg:
```
pref_org_list = pref.filters.get('orgs_filter_control', '')
org_list = set()
....
```

portal/static/js/admin.js
```diff
@@ -1,652 +1,781 @@
...
+              if (data["user_documents"] && data["user_documents"].length > 0) {
+                var existingItems = {}, count = 0;
+                /*
+                 * sort to get the latest first
+                 */
+                var documents = data["user_documents"].sort(function(a,b){
+                   return new Date(b.uploaded_at) - new Date(a.uploaded_at);
+                });
+                var content = "<table class='table-bordered table-condensed table-responsive tnth-table'>";
+                content += "<TH>" + i18next.t("Type") + "<TH>"+ i18next.t("Report Name") + "</TH><TH>" + i18next.t("Generated (GMT)") + "</TH><TH>" + i18next.t("Downloaded") + "</TH>";
```
ivan-c: HTML/content is easier to read and maintain if generated declarative, like with a template (instead of string concatenation).

I don't think we're including any libs that could help here (mustache), but it might be worth a second look if we start using one.

portal/static/js/admin.js
```diff
@@ -1,652 +1,781 @@
...
+      var tnthDates = self.getDependency("tnthDates");
+      /*
+       *  load the data sequentially
+       *  Note, NO concurrent ajax calls here,
+       *  one request will wait after the previous one has finished
+       */
+      var ajaxRequest = $.ajax ({
+                                type: "GET",
+                                url: "/api/consent-assessment-status",
+                                contentType: "application/json; charset=utf-8",
```
ivan-c: Is this necessary? Shouldn't we be setting a default for `contentType` in `ajaxSetup()`?

portal/static/js/admin.js
```diff
@@ -1,652 +1,781 @@
...
+      };
+      var userString = requests.shift();
+      if (!hasValue(userString)) {
+          return false;
+      };
+
+      var tnthDates = self.getDependency("tnthDates");
+      /*
+       *  load the data sequentially
+       *  Note, NO concurrent ajax calls here,
```
ivan-c: Seems like a good opportunity to try [Web workers](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API) if we're still having performance problems :)

portal/static/js/admin.js
```diff
@@ -1,652 +1,781 @@
...
+      if (hasValue(sortField) && hasValue(sortOrder)) {
+        data["sort_field"] = sortField;
+        data["sort_order"] = sortOrder;
+      } else {
+      	//get selected sorted field information on UI
+      	var sortedField = $("#adminTable th[data-field]").has(".sortable.desc, .sortable.asc");
+      	if (sortedField.length > 0) {
+      		data["sort_field"] = sortedField.attr("data-field");
+      		var sortedOrder = "desc";
+      		sortedField.find(".sortable").each(function() {
```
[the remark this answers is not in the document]
ivan-c: Please fix these moving forward if you can. I think most editors have a "leading tabs to spaces" function but let me know if you need help

ivan-c:
Let me know if any of the Codacy comments aren't exactly "helpful".

It's been pretty good with python but we had to disable one of its rules that was consistently wrong

portal/views/portal.py
achen2401: @ivan-c, ivan, didn't fix here as an else clause is attached to this statement to perform default action if the if clause fails here (in this case pref_org_list is null or empty)

portal/static/js/admin.js
achen2401: yeah, ivan, web work is a great idea.  will try that if we need to get assessment statuses via frontend in the future.  We have switched since to get them from backend.  This function is not in use at the moment.

PR #1606, opened 2017-11-14
Auto-push frontend strings to Smartling

ivan-c:
* Push new strings with same `tox` `translations` job
* Use `nodeenv` to install frontend dependencies (thanks @achen2401 !)

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -18,12 +18,6 @@ commands =
 changedir = docs
 commands = sphinx-build -W -n -b html -d {envtmpdir}/doctrees source {envtmpdir}/html
 
-[testenv:translations]
-passenv = SMARTLING_* {[testenv]passenv}
-commands =
-    flask sync
-    bash -c "if [[ "$TRAVIS_BRANCH" = "develop" ]]; then flask translation_upload ;fi"
-
 [testenv:ui]
 passenv = SAUCE_* {[testenv]passenv}
 commands =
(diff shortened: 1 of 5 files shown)


tox.ini
```diff
@@ -37,3 +31,28 @@ passenv = {[testenv]passenv} BRANCH GIT_REPO DOCKER_* COMPOSE_FILE
...
+deps =
+    nodeenv
+    {[testenv]deps}
+whitelist_externals =
+    nodeenv
+    /bin/bash
+commands =
+    nodeenv "{envtmpdir}/node_env"
+    bash -c 'source "{envtmpdir}/node_env/bin/activate" && npm --prefix "{toxinidir}/portal" install'
+    bash -c 'source "{envtmpdir}/node_env/bin/activate" && node "{toxinidir}/portal/node_modules/gulp/bin/gulp.js" --gulpfile "{toxinidir}/portal/i18next_gulpfile.js"'
```
achen2401: @ivan-c ivan, running the default gulp task here only extracts text in frontend code to a json file and then converts json file to pot file.  
Will we have another automated step that runs the conversion of po file(s) returned from Smartling into json file (containing translated text) which then can be consumed by the frontend?  Currently that is a gulp task of its own, i.e. "i18nextConvertPOToJSON".

ivan-c: Correct. We will still need to manually run that task (and commit the changes) every time new translations are added to Smartling.
The problem is that we can't commit to the portal git repo from TravisCI very easily, so we'll have to do it manually for now.

PR #1612, opened 2017-11-15
Feature/selectall orgs filter fix

achen2401:
fix the behavior of selectall checkbox in orgs filter control:
Jquery selector, :visible, also selects for elements that contain inline css of "display: none" - this resulted in all org elements including hidden ones being selected.
The fix is, in the onclick event of selectall checkbox, adding a check for the css property of "display:none" before checking an org element.

**Note** This will need to be merged **after** the hotfix has been merged to the develop branch as the hotfix contains admin.js code that does have code that has been added since the last release.


achen2401:
Ivan, yes #1613 needs to be merged first before this PR can be merged.

PR #1619, opened 2017-11-17
TN-185 - ePROMs - Manually enter treatment information

achen2401:
address https://jira.movember.com/browse/TN-185
add new treatment options for EPROMS

fix code to retrieve treatment options dynamically
fix previously hard-coded system information

will need to be reviewed by Paul
@pbugni FYI


portal/models/procedure_codes.py
```diff
@@ -158,6 +192,15 @@ def OtherProcedure(self):
             codings=[sno, ichom], text='Other procedure on prostate'
         ).add_if_not_found(True)
 
+    @lazyprop
+    def NoneOfTheAbove(self):
```
pbugni: Lets change the property name to `OtherPrimaryTreatment` (snomed, 999999999), if we need this one, so it doesn't collide with the pre-existing `NoneOfTheAbove` (truenth, 999).

pbugni:
Needing to move forward - I read Diana's comment - seems they are reluctant to change.  We'll alter if necessary down the road.

PR #1620, opened 2017-11-20
fix GIL footer position in patient's home page

achen2401:
noticing a gap in GIL footer in patient's home page when test.

- add styling fix to re-position the footer correctly
- add missing text to GIL's timeout modal


portal/gil/templates/gil/base.html
```diff
@@ -260,7 +260,7 @@ <h4 class="box-modal__copy">{{_("What is your main clinic for prostate cancer ca
         <div class="modal-content box-modal">
           <div class="box-modal__inner"><a class="box-modal__close" data-dismiss="modal" aria-label="{{_('Close')}}"></a>
             <h4 class="box-modal__title">{{_("Session Expiration Warning")}}</h4>
-            <span class="box-modal__copy">{{_("Your session will expire in approximately ")}}</span><span class="box-modal__copy text-warning" id='remaining-time'></span>
+            <span class="box-modal__copy">{{_("Your session will expire in approximately ")}}</span><span class="box-modal__copy text-warning" id='remaining-time'></span>  {{ _("seconds due to inactivity.") }}
```
pbugni: Wouldn't this be better in a single {% trans %} block for translator context?

portal/gil/templates/gil/base.html
```diff
@@ -387,13 +387,14 @@ <h4 class="box-modal__title">{{_("Session Timed Out")}}</h4><p class="box-modal_
                     onwarning: function() {$("#session-warning-modal").modal("show"); if (sessMon.modalShown) sessMon.intervalMonitor = setInterval(function(){ sessMon.ontimeout() }, 2 * 60 * 1000);}
               });
               window.sessMon = sessMon;
-               // Configure the session timeout warning modal
+              var warningText = i18next.t("Your session will expired in approximately [time] seconds due to inactivity.".replace("[time]",(sessMon.timeBeforeWarning / 1000)));
```
ivan-c: That should probably be "expire".

ivan-c: Could you keep consistent with your other `replace()` calls and `{}` (unless you're already using `[]` everywhere)?

ivan-c: Good idea moving this into the frontend by the way! :)

achen2401:
@ivan-c ivan, i think i addressed all the issues u raised. let me know if u see any problem with my changes.

portal/gil/templates/gil/base.html
```diff
@@ -387,13 +387,14 @@ <h4 class="box-modal__title">{{_("Session Timed Out")}}</h4><p class="box-modal_
                     onwarning: function() {$("#session-warning-modal").modal("show"); if (sessMon.modalShown) sessMon.intervalMonitor = setInterval(function(){ sessMon.ontimeout() }, 2 * 60 * 1000);}
               });
               window.sessMon = sessMon;
-               // Configure the session timeout warning modal
+              var warningText = i18next.t("Your session will expire in approximately {time} seconds due to inactivity.".replace("{time}",(sessMon.timeBeforeWarning / 1000)));
```
ivan-c: Whoops, almost missed that- can you make sure `replace()` is called on/after `t()`? (It looks like `replace()` is being called before `t()`)

achen2401:
@ivan-c ivan, i fixed the misplaced replace.  let me know if you see any more issue.

PR #1623, opened 2017-11-21
Feature/persistence file by type

pbugni:
Intermediate step for TN-375, all resourceTypes now live in their own respective persistence files.  Should be no content change to the persisted data in this PR - just how we manage it.


portal/config/model_persistence.py
```diff
@@ -0,0 +1,237 @@
...
+        self.sequence_name = sequence_name
+
+    def _log(self, msg):
+        current_app.logger.info(msg)
+
+    def __header__(self, data):
+        data['resourceType'] = 'Bundle'
+        data['id'] = 'SitePersistence v{}'.format(self.VERSION)
+        data['meta'] = {'fhir_comments': [
+            "export of dynamic site data from host",
```
ivan-c: Would it worthwhile to also include the `PERSISTENCE_DIR` config (eproms/gil)?

portal/config/model_persistence.py
```diff
@@ -0,0 +1,237 @@
...
+            result = self.update(o)
+            db.session.commit()
+            if self.lookup_field == 'identifier':
+                objs_seen.append(result.id)
+            else:
+                objs_seen.append(getattr(result, self.lookup_field))
+
+        # Delete any not named
+        if not keep_unmentioned:
+            if self.lookup_field == 'identifier':
```
ivan-c: I think you could remove the if/else block with a dict lookup (unless I missed something), something like this:

```
field = {'identifier':'id'}.get(self.lookup_field, self.lookup_field)
...

query = self.model.query.filter(~getattr(self.model, field).in_(objs_seen)) if objs_seen else []
```

manage.py
```diff
@@ -108,21 +105,23 @@ def seed(exclude_interventions=False, keep_unmentioned=False):
...
-        exclude_interventions = False
-
     # import site export file if found
-    SitePersistence().import_(exclude_interventions, keep_unmentioned)
+    SitePersistence().import_(keep_unmentioned=keep_unmentioned)
 
 
-@app.cli.command()
-def export_site():
+@click.option('--dir', '-d', default=None, help="Export directory")
```
ivan-c: Are you expecting we'll use manage.py to update orgs outside the release process (out of band)? If so, where will these changes be kept?

portal/config/model_persistence.py
pbugni: Different context. `self.lookup_field` is a ModelPersistence attribute, used to adapt different model classes - some are looked up by id, some name, some use the related 'identifier' table.

PR #1630, opened 2017-11-22
TN-342 Update UserConsent model/api; omit withdrawn from TNUSA staff patients

OptimusRhine:
https://jira.movember.com/browse/TN-342

* added 'status' field to UserConsent model
  * restricted to `('consented', 'suspended', 'deleted')` values
  * required; defaults to 'consented'
  * added to JSON methods
* migration for new UserConsent.status field
  * defaults to 'consented'; sets to 'suspended' or 'deleted' depending on `UserConsent.options` or `UserConsent.deleted_id`
* added new `/user/<user_id>/consent/withdraw [PUT, POST]` endpoint
  * only requires user_id (in path) and organization_id (in JSON) - similar to consent [DELETE]
  * replaces existing valid user/org consent with a matching one, minus the following changes:
    * status = 'withdrawn'
    * include_in_reports = True
    * send_reminders = False
    * staff_editable = False IF config['GIL'], else True
* updated patients list to only include users whose `consent.staff_editable` is True
* added new unit tests to test new functionality + new changes to existing functionality
* minor pep8 fixes throughout relevant files

### DIFF
--- a/portal/migrations/versions/9e5c1c6c4d64_.py
+++ b/portal/migrations/versions/9e5c1c6c4d64_.py
@@ -25,7 +25,7 @@ def upgrade():
     session = Session(bind=bind)
 
     status_enum = sa.Enum(
-        'consented', 'suspended', 'purged', name='status_enum')
+        'consented', 'suspended', 'deleted', name='status_enum')
     status_enum.create(op.get_bind(), checkfirst=False)
 
     op.add_column('user_consents',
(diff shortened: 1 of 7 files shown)


OptimusRhine:
NOTE:

This should **NOT** be merged until @achen2401 has a chance to add her changes for https://jira.movember.com/browse/TN-404

OptimusRhine:
@achen2401 There should already be a check there; look at https://github.com/uwcirg/true_nth_usa_portal/blob/develop/portal/models/user.py#L1140

`user.check_role("edit", other_id=user_id)` gets called during the /profile load (where `user` is the acting user, and `user_id` is the ID of the profile user). If `user` is Staff or StaffAdmin, and the profile user's consent does not have `staff_editable`, the check fails.

I haven't had a chance to actually test this though; if you're seeing different behavior in the portal, let me know, I can dig in and see why that's failing.

ivan-c:
@achen2401 Please merge if you're satisfied 👍

achen2401:
@ivan-c Ivan, I think we are waiting for @pbugni Paul to look at these changes.  It would be great if he can have a peek.

achen2401:
@OptimusRhine adam, you are right, just tried this again.  I think it is working as intended  I realized I was trying that out using an staff account that has admin privilege.

PR #1631, opened 2017-11-22
Feature/live org import

pbugni:
Implements https://jira.movember.com/browse/TN-375
org persistence can now be imported directly from admin's /settings page

### DIFF
--- a/portal/config/model_persistence.py
+++ b/portal/config/model_persistence.py
@@ -9,7 +9,7 @@ from ..database import db
 from ..dict_tools import dict_match
 from ..date_tools import FHIR_datetime
 from ..models.identifier import Identifier
-
+from ..trace import trace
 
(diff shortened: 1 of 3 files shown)


portal/config/model_persistence.py
```diff
@@ -99,6 +100,7 @@ def import_(self, keep_unmentioned):
                 query.delete(synchronize_session=False)
 
         self.update_sequence()
+        trace("Import of {} complete".format(self.model.__name__))
```
ivan-c: Why are you capturing these "log" messages in a trace instead of logging?

pbugni: that one is just noise when calling `flask seed` - it's extra that i wanted to view when calling live from the /settings page.  otherwise, you'll note the use of the common `_log()` method.

PR #1641, opened 2017-11-27
TN-396 Only count patients for overdue stats

OptimusRhine:
https://jira.movember.com/browse/TN-396

* modified overdue questionnaire reporting logic, so that we ONLY count users with the Patient role (while still ignoring any Patients that also have the Test role)


OptimusRhine:
Made against wrong branch; deleting.

PR #1645, opened 2017-11-28
Codings in site persistence

pbugni:
This should resolve the final details for https://jira.movember.com/browse/TN-185, namely display differences in the treatment types list.


pbugni:
Wrong branch - want this in the RC.

PR #1651, opened 2017-11-28
Fix treatment order

pbugni:
Another detail for TN-185 - corrected order of treatment options in dropdown to match spec.


pbugni:
Wrong branch!  Nevermind.

PR #1657, opened 2017-11-28
TN-393 Use unique name for CoM intervention reporting

OptimusRhine:
https://jira.movember.com/browse/TN-393

* modified intervention reporting, so that 'community_of_wellness' intervention shows up as 'Community of Wellness' (instead of the actual description 'Exercise and Diet')

### DIFF
--- a/portal/models/reporting.py
+++ b/portal/models/reporting.py
@@ -60,6 +60,8 @@ def get_reporting_stats():
                 desc = interv.description
                 if interv.name == 'decision_support_p3p':
                     desc = 'Decision Support P3P'
+                elif interv.name == 'community_of_wellness':
+                    desc = 'Community of Wellness'
                 if interv.quick_access_check(user):
                     stats['intervention_access'][desc] += 1
                 if interv in user.interventions:


portal/models/reporting.py
```diff
@@ -60,6 +60,8 @@ def get_reporting_stats():
                 desc = interv.description
                 if interv.name == 'decision_support_p3p':
                     desc = 'Decision Support P3P'
+                elif interv.name == 'community_of_wellness':
+                    desc = 'Community of Wellness'
```
pbugni: Is there any reason why we can't use the intervention.description fields for these?  (Alter in the config/*/Intervention.json files).

OptimusRhine: @pbugni 
We use the description for most of these. These are just two specific exceptions, and just for reporting.

For 'decision_support_p3p', they specifically wanted the 'P3P' tacked onto the end in the reporting (the normal description is just 'Decision Support').

For 'community_of_wellness', the description is 'Exercise and Diet', which is the same as the 'exercise_and_diet' intervention. Not sure why we have two interventions with the same description, but they specifically asked for c_o_w to have a different description in the reporting dashboard, so they could distinguish between the two.

OptimusRhine: As for whether or not we can just change the intervention's actual descriptions, that's a question for @mcjustin , as it would affect all displays/etc of these two interventions.

PR #1661, opened 2017-11-30
TN-436 cleanup Default Coding

OptimusRhine:
https://jira.movember.com/browse/TN-436

* remove the Default Coding object from config files
* migration to remove existing Default Coding objects from DBs
  * removes Default Coding object + related CodeableConcept object
  * sets any `Organization.default_locale_id` that pointed to the Default Coding, to `None`
  * sets any `User.locale_id` that pointed to the Default Coding, to `None`
  * downgrade restores the Default Coding/CodeableConcept (but does _not_ point any user/org locales back at Default, as we have no way of differentiating between user/org locales that were always `None`)

**NOTE: Please allow for unit tests to completely finish before merging**

### DIFF
--- a/portal/config/eproms/Coding.json
+++ b/portal/config/eproms/Coding.json
@@ -6049,12 +6049,6 @@
       "system": "http://www.ichom.org/medical-conditions/localized-prostate-cancer/"
     },
     {
-      "code": "",
-      "display": "Default",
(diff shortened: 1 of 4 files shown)


OptimusRhine:
(looks like there's an existing broken Communication test in develop; can wait for test fix before merging)

PR #1662, opened 2017-11-30
Need to include as_of_date when calculating expiry and due dates as well

pbugni:
This was creating stack problems when paper backdating hit date intervals outside of a valid recurrence.


pbugni:
wrong branch!

PR #1666, opened 2017-11-30
Moved Coding and CodeableConcept to their own modules

pbugni:
Also added IETF_LANGUAGE_TAG and LocaleConstants for consistent use.


pbugni:
This is a big change (# of files, not really anything serious) and I want these changes for other features... so I'm merging now.

PR #1671, opened 2017-12-01
TN-140 initial queries buttons styling fixes, and also JS refactor

achen2401:
address https://jira.movember.com/browse/TN-140, also taking the time to fix things in initial queires
-align finish button in initial queries
-re-indent html tags
-add fixes for errors returned from html validator
-move JS code (except for the shared JS code in template) to a standalone file as the template file is getting bloated


achen2401:
found issue with the change, will test some more

PR #1673, opened 2017-12-02
Use QB to pull QNR for AssessmentStatus

OptimusRhine:
https://jira.movember.com/browse/TN-298

* updated AssessmentStatus lookup of QuestionnaireResponses, to query based on the relevant QB (we're already assigning QNRs to QBs, just weren't utilizing that info in the lookup)
* updated unit tests to properly test QNR-QB assignment/lookup (e.g. `mock_qr()` now actually sets a QB for the QNR, so the lookup mechanism actually has something to query against)
* cleaned up unit tests (e.g. `mock_qr()` had a user_id param, but in 100% of cases was just using `TEST_USER_ID`, so param wasn't actually needed)

### DIFF
--- a/portal/models/assessment_status.py
+++ b/portal/models/assessment_status.py
@@ -11,11 +11,12 @@ from ..trace import trace
 from .user import User
 
 
-def recent_qnr_status(user, questionnaire_name):
+def recent_qnr_status(user, questionnaire_name, qb):
     """Look up recent status/timestamp for matching QuestionnaireResponse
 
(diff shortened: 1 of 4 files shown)


OptimusRhine:
Changed target branch. Tests should still be fine, but rerunning to confirm.

OptimusRhine:
@pbugni If this is targeting the new hotfix branch (release/v17.11.22.1), then I think this should be good to merge? Once you've reviewed it, that is.

PR #1674, opened 2017-12-04
During account promotion, exclude timezone so we retain the one

pbugni:
established during the original account promotion.

Fix for https://jira.movember.com/browse/TN-423

### DIFF
--- a/tests/test_user.py
+++ b/tests/test_user.py
@@ -990,6 +990,7 @@ class TestUser(TestCase):
                                   last_name='Better')
             other.birthdate = '02-05-1968'
             other.gender = 'male'
+            other.timezone = 'US/Eastern'
             self.shallow_org_tree()
             orgs = Organization.query.limit(2)
             other.organizations.append(orgs[0])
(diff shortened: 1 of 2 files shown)


tests/test_user.py
```diff
@@ -990,6 +990,7 @@ def test_merge(self):
                                   last_name='Better')
             other.birthdate = '02-05-1968'
             other.gender = 'male'
+            other.timezone = 'US/Eastern'
```
ivan-c: Should this be `America/New_York`? I think that's the standardized name...

ivan-c: Never mind, they are [two distinct values](https://stackoverflow.com/questions/9863625/difference-between-est-and-america-new-york-time-zones/9863679#9863679):

>EST is UTC - 5 hours. America/New_York is EST in the winter and E*D*T in the summer

pbugni: I'll note, those are simply testing values to confirm we keep the one we want.

PR #1675, opened 2017-12-04
Fix minor issue in settings POST

OptimusRhine:
* fixing minor issue with `dump_trace()` call in /settings POST

### DIFF
--- a/portal/views/portal.py
+++ b/portal/views/portal.py
@@ -627,7 +627,7 @@ def settings():
         form=form,
         apptext=apptext,
         organization_consents=organization_consents,
-        trace_data=dump_trace(),
+        trace_data=(dump_trace() if form.import_orgs.data else None),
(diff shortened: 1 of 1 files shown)


OptimusRhine:
@pbugni Sorry about that - it seemed like such a minor change, I just merged after the tests passed.

The issue was just that `dump_trace` is only imported `if form.import_orgs.data`, but `render_template(..., trace_data=dump_trace())` was called regardless of whether `form.import_orgs.data` - see https://github.com/uwcirg/true_nth_usa_portal/blob/develop/portal/views/portal.py#L607

So we could also solve this by moving the import line out of the conditional - if you'd prefer that, lemme know, and I'll put in another small fix.

pbugni:
This is my bug, and I'd like to fix it correctly.  Reversing this PR and will generate a new.

PR #1680, opened 2017-12-05
Revert "Fix minor issue in settings POST"

pbugni:
Reverts uwcirg/true_nth_usa_portal#1675


pbugni:
Rolling back the symptom fix - pending PR will correct source (bad import location).

PR #1682, opened 2017-12-05
Fix for TN-453 - `resume_identifier` rather than `resume_id`

pbugni:
Provide the assessment engine with the actual identifier value from the QuestionnaireResponse for more reliable session behavior.

### DIFF
--- a/portal/models/fhir.py
+++ b/portal/models/fhir.py
@@ -572,6 +572,30 @@ def aggregate_responses(instrument_ids, current_user):
 
     return bundle
 
+
+def qnr_document_id(
+        subject_id, questionniare_bank_id, questionnaire_name, status):
+    """Return document['identifier'] for matching QuestionnaireResponse
+
+    Using the given filter data to look for a matching QuestionnaireResponse.
+    Expecting to find exactly one, or raises NoResultFound
+
+    :return: the document identifier
+
+    """
+    qnr = QuestionnaireResponse.query.filter(
+        QuestionnaireResponse.status == status).filter(
+        QuestionnaireResponse.subject_id == subject_id).filter(
+        QuestionnaireResponse.document[
+            ('questionnaire', 'reference')
+        ].astext.endswith(questionnaire_name)).filter(
+        QuestionnaireResponse.questionnaire_bank_id ==
+        questionniare_bank_id).with_entities(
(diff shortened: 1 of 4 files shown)


portal/models/fhir.py
```diff
@@ -572,6 +572,30 @@ def aggregate_responses(instrument_ids, current_user):
...
 
+
+def qnr_document_id(
+        subject_id, questionniare_bank_id, questionnaire_name, status):
+    """Return document['identifier'] for matching QuestionnaireResponse
+
+    Using the given filter data to look for a matching QuestionnaireResponse.
+    Expecting to find exactly one, or raises NoResultFound
+
+    :return: the document identifier
```
ivan-c: The identifier is a string, could you add `:rtype: str`?

tests/test_assessment_status.py
```diff
@@ -21,14 +23,23 @@
 from tests import TestCase, TEST_USER_ID
 
 
-def mock_qr(instrument_id, status='completed', timestamp=None, qb=None):
+def mock_qr(
+        instrument_id, status='completed', timestamp=None, qb=None,
+        doc_id=None):
+    if not doc_id:
+        doc_id = randint(10000, 50000)
```
ivan-c: Could you test with a string, since the `identifier` is one?

tests/test_assessment_status.py
```diff
@@ -301,6 +312,25 @@ def mark_metastatic(self):
 
 
 class TestAssessmentStatus(TestQuestionnaireSetup):
+
+    def test_qnr_id(self):
+        qb = QuestionnaireBank.query.first()
+        mock_qr(
+            instrument_id='irondemog',
+            status='in-progress', qb=qb,
+            doc_id=21212)
```
ivan-c: Could you test with a string, since the `identifier` is one?

portal/models/fhir.py
pbugni: This is python, we don't type things unless strictly necessary.  The code will work for either type, whatever is stored in the json.

ivan-c: Fair enough, this is just more boilerplate to keep up to date when code changes. 
Any strong feelings about type-hinting? :)

PR #1683, opened 2017-12-05
TN-442 reconcile display of org hierarchy - fix orgs UI display

achen2401:
address TN-442:  https://jira.movember.com/browse/TN-442
fix display of organizations
- add css class to help align organizations in hierachy
- sort child organizations in alphabetical order


pbugni:
@achen2401  Looks like we have conflicts with all the usual suspects.  I look forward to a process that builds those on deploy rather than commit :)

PR #1686, opened 2017-12-06
TN-337 ScheduledJob 'active' check & additional enhancements

OptimusRhine:
https://jira.movember.com/browse/TN-337

* major modifications to the `/scheduled_jobs` UI
  * cleaner formatting
  * new 'active' toggle, to easily activate/inactive jobs
  * entries will now fade out and disable after being run/modified, until API response is returned (to avoid multiple triggers)
* moved all scheduled task logic to decorator (`@scheduled_task`)
  * updated existing tasks to use new decorator (now _much_ cleaner/simpler)
* added `ScheduledJob.active` check logic
  * Celery will still "run" inactive jobs, but (a) the actual task functionality won't kick off, and (b) the `last_runtime` and `last_status` fields won't be updated. Instead, Celery will just log that the task ran, but nothing happened, as the job was marked as inactive.
* updated tasks to take `job_id` as a kwarg, instead of an arg
  * updated `ScheduledJob.trigger()` and the celery_worker job load functions to compensate
* added additional logging to celery_worker job load (so you can actually see what ScheduledJobs are loaded on startup)
* updated `[PUT] /api/scheduled_job` to go off of job ID in the path (so now `/api/scheduled_job/<job_id>`)
* updated unit tests, added additional tests

### DIFF
--- a/portal/celery_worker.py
+++ b/portal/celery_worker.py
@@ -25,6 +25,8 @@ app.app_context().push()
 @celery.on_after_configure.connect
 def setup_periodic_tasks(sender, **kwargs):
     """Configure scheduled jobs in Celery"""
+    from .tasks import logger
+    logger.info("Starting ScheduledJob load...")
     # create test task if non-existent
     if not ScheduledJob.query.filter_by(name="__test_celery__").first():
         test_job = ScheduledJob(name="__test_celery__", task="test",
@@ -34,11 +36,17 @@ def setup_periodic_tasks(sender, **kwargs):
         test_job = db.session.merge(test_job)
 
     # add all tasks to Celery
-    for job in ScheduledJob.query.filter_by(active=True):
+    logger.info("ScheduledJobs found: {}".format(ScheduledJob.query.count()))
+    for job in ScheduledJob.query.all():
         task = getattr(tasks, job.task, None)
         if task:
+            logger.info("Adding task (id=`{}`, task=`{}` "
+                        "to CeleryBeat".format(job.id, job.task))
             args_in = job.args.split(',') if job.args else []
             kwargs_in = job.kwargs or {}
             kwargs_in['job_id'] = job.id
-            sender.add_periodic_task(job.crontab_schedule(),
-                                     task.s(*args_in, **kwargs_in))
+            try:
+                sender.add_periodic_task(job.crontab_schedule(),
+                                         task.s(*args_in, **kwargs_in))
+            except Exception as exc:
+                logger.info(exc)
--- a/portal/celery_worker.py
+++ b/portal/celery_worker.py
@@ -21,6 +21,7 @@ app = create_app()
 celery = create_celery(app)
 app.app_context().push()
 
+
 @celery.on_after_configure.connect
 def setup_periodic_tasks(sender, **kwargs):
     """Configure scheduled jobs in Celery"""
@@ -38,6 +39,6 @@ def setup_periodic_tasks(sender, **kwargs):
         if task:
             args_in = job.args.split(',') if job.args else []
             kwargs_in = job.kwargs or {}
+            kwargs_in['job_id'] = job.id
             sender.add_periodic_task(job.crontab_schedule(),
-                                     task.s(job.id, *args_in,
-                                            **kwargs_in))
+                                     task.s(*args_in, **kwargs_in))
--- a/portal/celery_worker.py
+++ b/portal/celery_worker.py
@@ -49,4 +49,4 @@ def setup_periodic_tasks(sender, **kwargs):
                 sender.add_periodic_task(job.crontab_schedule(),
                                          task.s(*args_in, **kwargs_in))
             except Exception as exc:
-                logger.info(exc)
+                logger.error(exc)
--- a/portal/templates/scheduled_jobs_list.html
+++ b/portal/templates/scheduled_jobs_list.html
@@ -3,9 +3,9 @@
 
 
 
-    <h1 class="tnth-headline">{{ _("Scheduled Jobs") }}</h1>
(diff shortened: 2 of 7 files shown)


portal/celery_worker.py
```diff
@@ -33,11 +36,17 @@ def setup_periodic_tasks(sender, **kwargs):
...
             kwargs_in = job.kwargs or {}
-            sender.add_periodic_task(job.crontab_schedule(),
-                                     task.s(job.id, *args_in,
-                                            **kwargs_in))
+            kwargs_in['job_id'] = job.id
+            try:
+                sender.add_periodic_task(job.crontab_schedule(),
+                                         task.s(*args_in, **kwargs_in))
+            except Exception as exc:
+                logger.info(exc)
```
ivan-c: Shouldn't this write to a different log level (`warn`/`error`)?

portal/templates/scheduled_jobs_list.html
```diff
@@ -10,20 +10,25 @@ <h4>{{ _("Current Jobs") }}</h4>
...
-                <br />{{ _("Active") }}: {{ job.active }}
-                <br />{{ _("Last Runtime") }}: {{ job.last_runtime }}
-                <br />{{ _("Last Status") }}: {{ job.last_status }}
-                <div id="triggerMessage_{{job.id}}" class="text-info"></div>
+                <div id="job_{{ job.id }}">
+                    <button title="Toggle Job" onclick="toggleJob(event, {{ job.id }})"><i id="activeIcon_{{job.id}}" class="{{'fa fa-toggle-on' if job.active else 'fa fa-toggle-off'}}" /></i></button>
+                    <button title="Run Job" onclick="triggerJob(event, {{ job.id }})"><i class="fa fa-play"></i></button>
+                    <b>{{ job.name }}</b>
+                    <div id="activeText_{{job.id}}" class="{{'text-info' if job.active else 'text-danger'}}">{{ "Active" if job.active else "Inactive" }}</div>
+                    {{ _("Task") }}: {{ job.task }}
```
ivan-c: I don't think we'll want to translate this content (notes that would only help a developer), at least for now. Will lead to a lot of questions from translators :).

There isn't any way to disable extraction, just from this file, is there? There's always the possibility they may change their minds...

portal/templates/scheduled_jobs_list.html
```diff
@@ -49,20 +54,82 @@ <h4>{{ _("Available Tasks") }}</h4>
 function triggerJob(event, jobId) {
```
ivan-c: Would it be reasonable to move this to a separate JS file? Trying to move away from inline JS so we can prevent XSS

portal/templates/scheduled_jobs_list.html
OptimusRhine: Well, right now it's just a few field labels getting translated. But I can just change those to straight strings (aka `Task` instead of `{{ _("Task") }}`) if we don't want those translated, sure.

portal/celery_worker.py
OptimusRhine: Good point, I'll update that.

portal/templates/scheduled_jobs_list.html
OptimusRhine: I'm fine with that, but I'm not sure what our standard approach here is (aka where exactly this code should live instead). Maybe @achen2401 can take a look?

portal/templates/scheduled_jobs_list.html
ivan-c: Thanks! The only ones I think they might balk at are `Kwargs` and `Args`

portal/templates/scheduled_jobs_list.html
achen2401: @OptimusRhine adam, the JS code here probably merits its own JS file since it only relates to schedule job.  I, however, don't have strong opinion here as since only admin staff has access to schedule job-related pages - and I think a check on user role should probably prevent any other user from accessing those pages, but  I am not an expert on XSS attack. I will leave it up to you whether to make the change here or not.

OptimusRhine: I think it's fine to leave as-is for now, as the page wouldn't be loaded in the first place unless the user has admin access.

pbugni: @OptimusRhine Sounds like a worthy goal to pull the JS out, if it's not too much to ask?

OptimusRhine: Thanks! And OK, since it sounds like we're going for consistency with all our JS, I'll go ahead and pull that out.

OptimusRhine:
@achen2401 I moved the JS out to a separate file - is this what we're looking for? Function-wise, it seems to work fine.

portal/templates/scheduled_jobs_list.html
```diff
@@ -1,69 +1,48 @@
...
+                    <br />Kwargs: {{ job.kwargs }}
+                    {% endif %}
+                    <br />Schedule: {{ job.schedule }}
+                    <br />Last Runtime:
+                    <div id="lastRuntime_{{job.id}}" class="text-info">{{ job.last_runtime }}</div>
+                    Last Status:
+                    <div id="lastStatus_{{job.id}}" class="text-info">{{ job.last_status }}</div>
+                    <br />
+                </div>
+                <script src="{{ url_for('static', filename='js/scheduledJobs.js') }}"></script>
```
achen2401: @OptimusRhine Adam, looks great, thank you very much for the changes.  One thing I noticed here, I think the inclusion of script here should be moved out of the for loop otherwise it will be included multiple times.

OptimusRhine: Ah, good catch :). Should be fixed now.

PR #1688, opened 2017-12-06
Feature/orgs ui fix 2

achen2401:
more fixes for TN-442:  https://jira.movember.com/browse/TN-442
re-indent text,
fix order of orgs


achen2401:
adding one more fix, close for now

PR #1694, opened 2017-12-06
TN-393 Filter DSP3p from intervention /reporting

OptimusRhine:
https://jira.movember.com/browse/TN-393

* added filter of 'Decision Support P3P' from 'interventions' reporting only (_not_ 'interventions doc' reporting)
* added more interventions reporting unit tests

### DIFF
--- a/portal/models/reporting.py
+++ b/portal/models/reporting.py
@@ -64,7 +64,8 @@ def get_reporting_stats():
                     desc = 'Community of Wellness'
                 if interv.quick_access_check(user):
                     stats['intervention_access'][desc] += 1
-                if interv in user.interventions:
+                if ((interv in user.interventions) and
+                        (interv.name != 'decision_support_p3p')):
                     stats['interventions'][desc] += 1
                 if (any(doc.intervention == interv for doc in user.documents)):
                     stats['intervention_reports'][desc] += 1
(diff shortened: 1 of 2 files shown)


portal/models/reporting.py
```diff
@@ -64,7 +64,8 @@ def get_reporting_stats():
                     desc = 'Community of Wellness'
                 if interv.quick_access_check(user):
                     stats['intervention_access'][desc] += 1
-                if interv in user.interventions:
+                if ((interv in user.interventions) and
+                        (interv.name != 'decision_support_p3p')):
                     stats['interventions'][desc] += 1
```
pbugni: Can you help me understand why we need this special case?  The code shouldn't in general be using configuration strings if we can avoid it.

OptimusRhine: Some discussion of this in https://jira.movember.com/browse/TN-393 , but in summary:

There are a some users throughout the instances (including prod) that have DSP3P assigned to their `user.interventions`. Justin believes these are artifacts of older, now nonexistent processes, but we can't really be sure that users still aren't getting assigned that intervention somehow. In any case, since it's not technically an intervention users should "belong" to, Justin asked if we could just filter it out from this specific reporting table (so analysts get more accurate data).

pbugni: Please modify this check to filter out ALL interventions with either (public_access == True or access_strategy count >0)

PR #1695, opened 2017-12-07
Bug fix for TN-485

pbugni:
Simplify url building for "Enter Manually" - several special cases no longer apply with the `/present-needed` API.

This should resolve symptom found in https://jira.movember.com/browse/TN-485

### DIFF
--- a/portal/templates/profile_macros.html
+++ b/portal/templates/profile_macros.html
@@ -1885,7 +1885,7 @@
                 }, 2000);
 
                 setTimeout(function() { window.location = winLocation;}, 100);
-                
+
             } else {
                 $("#manualEntryMessageContainer").html("{%trans%}The user does not have a valid assessment link.{%endtrans%}");
             };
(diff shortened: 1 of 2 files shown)


portal/templates/profile_macros.html
```diff
@@ -2050,9 +2046,11 @@ <h4 class="modal-title">{{ _("Enter questionnaire manually on patient's behalf")
 
             var method = $("input[name='entryMethod']:checked").val();
             var completionDate = $("#qCompletionDate").val();
+            var linkUrl = "{{url_for('assessment_engine_api.present_needed')}}" + "?subject_id={{person.id}}";
```
ivan-c: Would this be better?
```
{{url_for('assessment_engine_api.present_needed', subject_id=person.id)}}
```

achen2401: @ivan, ivan, thank u, should be fixed now.

ivan: When you type @ivan anywhere into GitHub, the wrong ivan gets an email

PR #1699, opened 2017-12-08
Handle QNR -> QB for the indefinite case

pbugni:
Previously, we were only matching up QNR submissions that fit within the "most_current_qb" - adding logic for the indefinite case.

### DIFF
--- a/portal/models/questionnaire_bank.py
+++ b/portal/models/questionnaire_bank.py
@@ -291,7 +291,7 @@ class QuestionnaireBank(db.Model):
         expired.
 
         NB the `indefinite` classification is outside the scope of this method,
-        and should be treated independently
+        and should be treated independently - see `indefinite_qb`
 
         """
         as_of_date = as_of_date or datetime.utcnow()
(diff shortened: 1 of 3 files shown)


portal/models/questionnaire_bank.py
```diff
@@ -322,6 +322,30 @@ def most_current_qb(user, as_of_date):
...
+        named tuple.
+
+        :returns QBD: with values only if the user has an indefinite qb
+
+        """
+        indefinite_qb = QuestionnaireBank.qbs_for_user(user, 'indefinite')
+        if not indefinite_qb:
+            return QBD(None, None, None, None)
+
+        if len(indefinite_qb) > 1:
```
OptimusRhine: Not a scenario we have to worry about right now (with only one indefinite QB in existence), but would it be better to just return the most recent (for example), and the log an error - similar to what we do for users belonging to more than one top-level org? Might help us avoid major crashes.

portal/views/assessment_engine.py
```diff
@@ -1312,6 +1312,14 @@ def assessment_add(patient_id):
         if (qb and qn and (qn.id in [qbq.questionnaire.id
                            for qbq in qb.questionnaires])):
             qnr_qb = qb
+        # if a valid qb wasn't found, try the indefinite option
```
OptimusRhine: My only worry with this is, what if the questionnaire is for the indefinite QB, but the same questionnaire type is in the user's `most_current_qb()`? Are we worried about the QNR getting assigned to the `most_current_qb()` instead of the indefinite? Or is this not a situation we're worried about (i.e. all the questionnaires for indefinite QBs, will never show up in non-indefinite QBs)?

portal/models/questionnaire_bank.py
pbugni: err, most recent what?  :)

portal/views/assessment_engine.py
pbugni: well, that situation would be tricky.  i'm comfortable waiting on that.

I would really like to see more aggressive validation here, but I'm afraid to break it at this hour.  Seems like we should not take submissions w/o a valid 'questionnaire' or a matching QB...

portal/models/questionnaire_bank.py
OptimusRhine: Mm, maybe just `indefinite_qb[0]`? I'm just slightly worried about how much would break for this hypothetical future user if an error was thrown here. But maybe it's not something we need to worry about for now.

PR #1701, opened 2017-12-08
TN-451 Modify consent audit logic

OptimusRhine:
https://jira.movember.com/browse/TN-451

* updated consent audit logic
  * existing consent audit comment changed from "Adding consent agreement" to "Consent agreement <consent_id> signed". Audit is otherwise unchanged (including timestamp)
  * new audit (NOT pointed at by the consent) created on consent create, with a timestamp of `datetime.utcnow()`, and a comment of "Consent agreement <consent_id> registered".
* created migration to change existing consent audit messages to match the new message
* added new and updated existing audit unit tests

### DIFF
--- a/portal/migrations/versions/59be0e4b2171_.py
+++ b/portal/migrations/versions/59be0e4b2171_.py
@@ -1,0 +1,49 @@
+from alembic import op
+import sqlalchemy as sa
+from sqlalchemy.orm import sessionmaker
+
+from portal.models.audit import Audit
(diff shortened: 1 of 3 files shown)


OptimusRhine:
(note about the pep8speaks comment above)

for `Object.query.filter( )` filter strings, `Object.field is not None` does not work the same way as `Object.field != None`. Mainly, the first still returns results where `Object.field == None`.

PR #1702, opened 2017-12-09
Log out of AE after assessment completion

ivan-c:
* Log current_user() out of Assessment Engine on survey completion (`/complete-assessment`)
* Also applies when "Take a Break" is used (follows same flow through `/complete-assessment`)

### DIFF
--- a/portal/models/auth.py
+++ b/portal/models/auth.py
@@ -249,7 +249,7 @@ class Token(db.Model):
         db.String(40), db.ForeignKey('clients.client_id', ondelete='CASCADE'),
         nullable=False,
     )
-    client = db.relationship('Client')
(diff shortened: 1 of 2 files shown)


ivan-c:
@pbugni What would be the "right" way to do it, without a for loop? 

I wasn't able to call any Query methods (eg `delete()`, `filter()`) on the new `backref` I created without adding `lazy='dynamic'`.

After adding `lazy='dynamic'` I got this error message:
`On relationship Token.client, 'dynamic' loaders cannot be used with many-to-one/one-to-one relationships and/or uselist=False`

pbugni:
Lets run with this implementation - sqla relationships are complicated!  I'd have to spend more time with it, and again, it's a tiny table - not worth it.

PR #1705, opened 2017-12-12
Postpone indefinite

pbugni:
Should fix latest symptom in https://jira.movember.com/browse/TN-434

Previously, suspending work on the `irondemog` assessment would cause an error on session ID lookup, as we were looking in the users "current_qb", and the indefinite case is *special*


pbugni:
Wrong target branch.  Will reopen as #1706

PR #1708, opened 2017-12-12
Feature/persistence extensions

pbugni:
Fix for https://jira.movember.com/browse/TN-313

Needed to modify behavior of the extension classes, so we can use them to clear out data as well as setting it - for example, when an organization previously named an extension, then it's removed from the site_persistence file.

### DIFF
--- a/portal/models/organization.py
+++ b/portal/models/organization.py
@@ -236,12 +236,18 @@ class Organization(db.Model):
             self.partOf_id = Reference.parse(data['partOf']).id
         for attr in ('use_specific_codings','race_codings',
                     'ethnicity_codings','indigenous_codings'):
-            if attr in data:
-                setattr(self, attr, data.get(attr))
-        if 'extension' in data:
-            for e in data['extension']:
-                instance = org_extension_map(self, e)
-                instance.apply_fhir()
+            # set regardless of presence in data - thus clearing
+            # old values if not currently mentioned
+            setattr(self, attr, data.get(attr))
+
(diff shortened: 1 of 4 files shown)


portal/models/organization.py
```diff
@@ -236,12 +236,18 @@ def update_from_fhir(self, data):
             self.partOf_id = Reference.parse(data['partOf']).id
         for attr in ('use_specific_codings','race_codings',
                     'ethnicity_codings','indigenous_codings'):
-            if attr in data:
-                setattr(self, attr, data.get(attr))
-        if 'extension' in data:
-            for e in data['extension']:
-                instance = org_extension_map(self, e)
-                instance.apply_fhir()
+            # set regardless of presence in data - thus clearing
```
OptimusRhine: A bit unclear on why we want to set these fields (`'use_specific_codings', 'race_codings', 'ethnicity_codings', 'indigenous_codings'`) even if they're not in the data. Should the behavior of these fields not be like the other org fields, where we only clear them out if they're included in the data with empty values?

(the extension logic change makes sense to me, this is only regarding those specific 4 fields above)

pbugni: that's exactly the point - we're going to clear them all, unless they're defined (exception is `id` as it's controlled by session, pg, etc.)  otherwise, we get into this mess, where data previously defined on an Organization doesn't go away on import.

pbugni:
Closing for now in light of partial PUT discussions.

PR #1711, opened 2017-12-13
Feature/persistence extensions

pbugni:
Another round, adhering to the partial PUT model.

Not crazy about needing to define empty extensions to clear out existing, but I believe that's what it takes to follow the suggested model.  For example, to clear out locale extensions for an organization, the persistence file must include:

```
      "extension": [
        {
          "url": "http://hl7.org/fhir/valueset/languages"
        }]
```

Partial fix for https://jira.movember.com/browse/TN-313 (also need updates in Organization.json)

### DIFF
--- a/portal/config/model_persistence.py
+++ b/portal/config/model_persistence.py
@@ -158,19 +158,27 @@ class ModelPersistence(object):
             serialize = 'as_json'
 
         merged = None
-        new_obj = from_method(new_data)
+
+        # Generate an empty but complete serialized form of the object
+        # so that regardless of shorthand in the persistence file (say
+        # ignoring empty fields), a valid representation is available.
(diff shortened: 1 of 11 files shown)


pbugni:
@OptimusRhine well, it's using stdlib dict.update, and i think there are several tests - anything in particular needing attention test wise?

OptimusRhine:
Oh yea, I mean as it is now, I'm confident it'll work as expected.

I was more thinking of adding in a small test (setting a field on the Org before the PUT, that's not referenced in the request JSON, and then after the PUT confirm that the field is still set), just in case someone down the road makes modifications to that view method, to make sure they don't call `from_fhir()` without first calling `as_fhir()`.

PR #1713, opened 2017-12-14
Special migration to clean up duplicate QNRs

pbugni:
NB: this only cleans up the identical ones.  Report will show
those needing additional attention on STDERR


ivan-c:
@pbugni Can you take another look at this when you get a chance?

PR #1714, opened 2017-12-14
TN-118 trigger new ToU agreement requests for existing users

OptimusRhine:
https://jira.movember.com/browse/TN-118

* added `ToU.active` field
  * defaults to `True`
* updated ToU checks (in `CoreData()` and in ToU [GET] API endpoints) to only return results where `ToU.active is True`
* added User method to deactivate all user's ToUs (optionally restricted to given `types` param)
* added unit tests
* added Celery task to call ToU deactivation on all Patient, Staff, or StaffAdmin users
* added ScheduledJob entry for above jobs to ePROMs and TrueNTH SP config files (with the jobs each being restricted to 'subject website consent' and 'website terms of use' types respectively, per request in Jira)


portal/models/user.py
```diff
@@ -867,6 +867,36 @@ def update_consents(self, consent_list, acting_user):
...
+        Called when the ToU agreement language is updated.
+
+        :param acting_user: user behind the request for permission checks
+        :param types: ToU types for which to invalide agreements (optional)
+
+        """
+        from .tou import ToU
+
+        for tou in ToU.query.join(Audit).filter(and_(
+                Audit.user_id == self.id,
```
pbugni: Shouldn't this be `Audit.subject_id == self.id`?

portal/tasks.py
```diff
@@ -315,3 +315,17 @@ def generate_and_send_summaries(cutoff_days, org_id):
...
+def deactivate_tous(**kwargs):
+    "Require users to re-consent to their initial consent"
+    types = kwargs.get('types')
+    sys = User.query.filter_by(email='__system__').first()
+
+    for user in User.query.filter(User.deleted_id.is_(None)):
+        if any((user.has_role(ROLE.PATIENT),
+                user.has_role(ROLE.STAFF),
+                user.has_role(ROLE.STAFF_ADMIN))):
+            user.deactivate_tous(acting_user=(sys or user), types=types)
```
pbugni: Seems like we should fail if there isn't a system user - giving this or option will leave audits that give the impression the user deactivated their own, right?

portal/views/tou.py
```diff
@@ -94,7 +94,9 @@ def get_tou(user_id):
         abort(404)
     current_user().check_role(permission='view', other_id=user_id)
 
-    tous = ToU.query.join(Audit).filter(Audit.user_id == user_id)
+    tous = ToU.query.join(Audit).filter(
+        and_(Audit.user_id == user_id,
```
pbugni: Again, I think we want `subject_id`, no?

portal/models/user.py
OptimusRhine: Ah, good point. Thing is, all the current ToU-setting and ToU-checking logic current checks against `Audit.user_id` (e.g. https://github.com/uwcirg/true_nth_usa_portal/blob/develop/portal/models/coredata.py#L373 ).

Should I update all ToU references to set/check `Audit.subject_id` instead, and also migrate all existing ToUs to set `Audit.subject_id = Audit.user_id`?

portal/tasks.py
OptimusRhine: Ok, I'll change it to throw an error instead.

portal/views/tou.py
OptimusRhine: (see similar comment above)

portal/models/user.py
pbugni: Lets clean it up!
Aren't those just checking the wrong value.  I mean, it's legit that user_id != subject_id in many flows - but it's only the subject_id that we'd care about when looking for a valid ToU, right?

OptimusRhine: Yea, it looks like we're setting the expected user to `Audit.subject_id` for those ToUs, we just never updated the relevant checks accordingly.

I'll update those (and my own checks here) to look at the correct field.

pbugni:
Looks solid - please merge when tests pass.

PR #1740, opened 2017-12-20
Expose translation state to flask CLI

ivan-c:
* Allow translation state to be passed when downloading translations
  * Auto-configures from `SYSTEM_STATE`
  * `production` will only downloaded published translations
* Add "translation download" step to deployment script
* Update translation files with all published translations


manage.py
```diff
@@ -167,12 +167,23 @@ def translation_upload():
...
-def translation_download(language):
+def translation_download(language, state):
     """Download .po file(s) from Smartling
 
     GETs the .po file for the specified language from Smartling via their API.
     If no language is specified, all available translations will be downloaded.
     After download, .po file(s) are compiled into .mo file(s) using pybabel
     """
-    smartling_download(language)
+
```
pbugni: would it be useful to capture some logging of this event?

ivan-c:
@pbugni Good point, would be an easy way to find a bunch of typos all at once

PR #1749, opened 2017-12-22
Fixup deploy script

ivan-c:
* Do not change branch unless specified (`-b "branch"` or `BRANCH="branch"`
* Reset local branch state to match remote branch when updating
  * **NB** Discards ALL local changes
* Add logging/comments

### DIFF
--- a/bin/deploy.sh
+++ b/bin/deploy.sh
@@ -19,14 +19,17 @@ update_repo(){
     git fetch origin
     git fetch --tags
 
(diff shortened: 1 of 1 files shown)


ivan-c:
@pbugni FYI I was able to confirm that the deploy script will [stop at the first command command that returns a non-zero exit code](https://www.gnu.org/software/bash/manual/html_node/The-Set-Builtin.html) (failure)
`git pull` must have been returning `0` as the exit code; I'll keep an eye out if any of our `flask` commands are doing the same when they are failing.

PR #1751, opened 2017-12-27
Feature/orphan identifiers

pbugni:
Fix for #TN-554 - included migration to remove obsolete cruft.


portal/migrations/versions/00b081faf417_.py
```diff
@@ -0,0 +1,52 @@
...
+# revision identifiers, used by Alembic.
+revision = '00b081faf417'
+down_revision = '8ecdd6381235'
+
+
+def upgrade():
+    conn = op.get_bind()
+    query = ("select identifiers.id, value, system from identifiers "
+        "where system = :identifier and identifiers.id "
+        "not in (select identifier_id from organization_identifiers)")
```
ivan-c: Why is this migration still necessary if you've added logic to delete "orphaned" (unspecified) identifiers when we read from config JSON? Just curious :)

pbugni: site persistence isn't doing this level of purge - it only deletes the ones that start out connected to current orgs but don't get mentioned in the JSON.  as many of these lost their org affiliation before site persistence was keeping track, they're out of reach.

ivan-c: Ahh, that makes sense, thanks!

PR #1754, opened 2017-12-28
Conditional Translation Download

ivan-c:
* Only download Smartling translations during deployment if configured
* Add `flask config` command for printing Flask configuration values

### DIFF
--- a/manage.py
+++ b/manage.py
@@ -5,6 +5,7 @@ FLASK_APP=manage.py flask --help
 """
 import os
 import click
+import json
 import redis
 
 import alembic.config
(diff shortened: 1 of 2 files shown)


manage.py
```diff
@@ -188,3 +189,20 @@ def translation_download(language, state):
...
+
+
+@click.option('--config_key', '-c', help='Return only a single config value')
+@app.cli.command()
+def config(config_key):
+    """List current flask configuration values in JSON"""
+
+    if config_key:
+        # Remap None values to an empty string
+        print(app.config.get(config_key, '') or '')
```
pbugni: the or here is unnecessary.  the second param to `get` is the default if not found.

manage.py
```diff
@@ -188,3 +189,20 @@ def translation_download(language, state):
...
+        return
+
+    print(json.dumps(
+        # Skip un-serializable values
+        {k: v for k, v in app.config.items() if isinstance(v, basestring)},
```
pbugni: Hmm, what types were you running into problems with?  This is silently omitting them, should we print out a string representation, or are you confident they aren't needed?

manage.py
ivan-c: If I don't have the `or`, config values that defined, but `None`(see below), will print `None` (instead of no printing)
```
config = {
    "key": None
}
```

manage.py
ivan-c: There are some config values that are objects (`SESSION_REDIS`); I don't think we're missing any important ones

PR #1759, opened 2018-01-04
TN-504 missing timezone fix

achen2401:
fix issue when user timezone is America/Chicago - not selected because it is missing from the timezone dropdown list
TODO - ask Paul about possible service for returning timezone list that frontend can make use of to dynamically populate the dropdown.

### DIFF
--- a/portal/static/js/main.js
+++ b/portal/static/js/main.js
@@ -1196,7 +1196,7 @@ var fillContent = {
     },
     "timezone": function(data) {
         data.extension.forEach(function(item, index) {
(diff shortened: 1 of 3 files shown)


achen2401:
@ivan-c Ivan, thank you very much!

pbugni:
@achen2401 I like that idea a good deal.  Is there already a story or sub-task for this?  I'll take this question to slack...

ivan-c:
@pbugni I think TN-557 is the subtask for that, (part of TN-555)

PR #1764, opened 2018-01-05
Fix Notification persistence issue

ivan-c:
* Do not set `Notification.created_at` in `as_json()` unless already saved in DB
  * `Notification.created_at` defaults to GMT now, and cannot be `null`
  * Returning `None` in `as_json()` when creating the empty/template JSON caused `null` to be saved to DB field


ivan-c:
@pbugni Please take a look when you get a chance, I want this to fit into your JSON import/export model

pbugni:
@ivan-c yes, we do a similar check in many a serialization method for the same reason.

PR #1768, opened 2018-01-08
TN-398 - follow up fix to check for existing notification before

achen2401:
follow-up fix to this story:  
https://jira.movember.com/browse/TN-553

- add check for existence of notification before deleting it - covering use case where multiple windows containing notification are open

### DIFF
--- a/portal/static/js/main.js
+++ b/portal/static/js/main.js
@@ -5154,16 +5154,19 @@ var tnthAjax = {
         }
     },
     "deleteNotification": function(userId, notificationId, params, callback) {
+
+        if (!callback) {
+            callback = function(data) {
+                return data;
+            }
+        }
+        
         if (!hasValue(userId)) {
-            if (callback) {
-                callback({"error": i18next.t("User Id is required")});
-            };
+            callback({"error": i18next.t("User Id is required")});
             return false;
         };
         if (!hasValue(notificationId)) {
-            if (callback) {
-                callback({"error": i18next.t("Notification id is required.")});
-            };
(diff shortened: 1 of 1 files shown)


portal/static/js/main.js
```diff
@@ -5166,16 +5166,36 @@ var tnthAjax = {
...
+                 */
+                var arrNotification = $.grep(data.notifications, function(notification) {
+                    return notification.id == notificationId;
+                });
+                if (arrNotification.length > 0) {
+                    /* 
+                     * delete notification only if it exists
+                     */
+                    self.sendRequest('/api/user/'+userId+'/notification/'+notificationId, 'DELETE', userId, {"sync":params.sync}, function(data) {
+                        if (data) {
```
ivan-c: Would it make sense to check `if (!callback) return;` once, at the beginning of this function? Then you wouldn't need to each time?

achen2401: @ivan-c Ivan, ah, you are right - I see multiple checks for the existence of callback - makes sense to just check for it once.  I did a slight modification of your suggestion - assigning callback to an anonymous function if it is not provided, in case the callee didn't provide it but still wants the function executed without callback.

ivan-c: Thanks for giving it a try! 
I've seen in the cPRO codebase that a pile of if/else only gets bigger over time. I try to avoid [deeply nesting if/else conditions](https://refactoring.com/catalog/replaceNestedConditionalWithGuardClauses.html) when I can, but I wasn't sure if it made sense here.

PR #1779, opened 2018-01-10
TN-536 - Updates to the patient notification when website consent has been updated (re-consent)

achen2401:
address: https://jira.movember.com/browse/TN-536

modify terms UI display when re-consenting of terms is required.

- hide pre-existing checked terms


portal/templates/initial_queries_macros.html
```diff
@@ -131,9 +131,9 @@ <h4 class="modal-title" id="termsReminderModalLabel">{{ _("To Continue") }}</h4>
     {{tou_agreement_url()}}
     <script>
       $(function() {
-          tnthAjax.getTerms({{user.id}}, false, false, function() {
+          tnthAjax.getTerms($("#iq_userId").val(), false, false, function() {
```
ivan-c: Would you be able to move this into `initialQueries.js` and check for `$('#iq_userId')` from there?

portal/static/js/main.js
```diff
@@ -4886,8 +4908,12 @@ var tnthAjax = {
             };
         });
     },
-    "getTerms": function(userId, type, sync, callback) {
-        this.sendRequest('/api/user/'+userId+'/tou'+(type && hasValue(type)?('/'+type):''), 'GET', userId, {sync:sync}, function(data) {
+    "getTerms": function(userId, type, sync, callback, params) {
+        if (!params) params = {};
+        var url = "/api/user/{userId}/tou{type}{all}".replace("{userId}", userId)
```
ivan-c: This seems kind of fragile (building QS params by concatenation). Are "all" and "type" mutually exclusive?
Would you be able to use `jQuery.param()` to build it instead?

portal/static/js/main.js
```diff
@@ -1263,18 +1269,33 @@ var fillContent = {
...
+                 */
+                if (arrReconsent.length > 0) {
+                    self.attr("data-reconsent", "true");
+                    if (!setReconsentDisplay) {
+                        $(this).closest("#termsCheckbox").attr("data-reconsent", "true");
+                        setReconsentDisplay = true;
+                    }
+                }
+
+                if (item_found !== 0 && (item_found == arrTypes.length)) {
```
ivan-c: Should that be `item_found > 0` ? I assume the type-checking will be handled by the second condition...

portal/static/js/main.js
achen2401: @ivan-c Ivan, "all" is a optional querystring parameer and "type" can be optional part of api path - so they can co-exist.  I am thinking we can't use jQuery.param() here - the serialized string from jQuery.param(), i.e. a=blah&b=blah, will not work in this instance, i.e. /api/user/[type][?all=true]

portal/static/js/initialQueries.js
```diff
@@ -57,6 +66,14 @@
     });
   };
 
+  FieldsChecker.prototype.initData = function(section) {
+    var tnthAjax = this.__getDependency("tnthAjax");
+    if (this.mainSections[section] && this.mainSections[section].initData) {
```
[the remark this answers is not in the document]
ivan-c: @achen2401 I finally looked up [this warning](https://blog.liftsecurity.io/2015/01/14/the-dangers-of-square-bracket-notation/). FYI It's only an issue when we assign a property and value from user input

```
field_name_from_user='constructor';
value_given_by_user='{}';

patient_object[field_name_from_user] = value_given_by_user;
```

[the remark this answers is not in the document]
achen2401: @ivan-c Ah, good to know, Ivan! Thank you very much for looking that up.  I was wondering about that warning.

PR #1799, opened 2018-01-17
Feature/re consent by study

pbugni:
Implementation for TN-535 and TN-561, adding finer controls to deactivating consents and tasks as requested.

### DIFF
--- a/portal/config/eproms/ScheduledJob.json
+++ b/portal/config/eproms/ScheduledJob.json
@@ -70,12 +70,77 @@
       "active": false,
       "args": null,
       "kwargs": {
-        "types": ["subject website consent"]
+        "types": ["privacy policy"],
+        "research_protocol": "IRONMAN v2"
       },
-      "name": "Deactivate subject website consent agreements",
+      "name": "Deactivate IRONMAN (v2) Privacy Policy",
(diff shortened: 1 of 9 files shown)


portal/config/eproms/ScheduledJob.json
```diff
@@ -70,12 +70,77 @@
...
-        "types": ["subject website consent"]
+        "types": ["privacy policy"],
+        "research_protocol": "IRONMAN v2"
       },
-      "name": "Deactivate subject website consent agreements",
+      "name": "Deactivate IRONMAN (v2) Privacy Policy",
       "resourceType": "ScheduledJob",
-      "schedule": "0 0 1 1 0",
-      "task": "deactivate_tous"
+      "schedule": "0 0 0 0 0",
```
OptimusRhine: So technically '0' isn't a valid value for the 3rd and 4th position with Cron schedules (see https://en.wikipedia.org/wiki/Cron).

I know the job probably won't ever be active, but the invalid values do cause errors when the job gets loaded into Celery, so it might be worth changing those to '1's in any case. Same with the other schedules below.

I'll put in a change shortly adding a comment about that in ScheduledJob, as well as updating the regex validation there accordingly.

PR #1801, opened 2018-01-19
Feature/website re-consent notifications fixes

achen2401:
addressing some issues found while testing TN-536:
https://jira.movember.com/browse/TN-536

- **subject website consent terms** are required of patients only, so notifications should only specify the role of patient
- change the job names to adhere to the naming convention for coredata
- styling fixes for notifications in banner
- fix duplicate entries issue in schedule jobs table when syncing


portal/migrations/versions/40b223faa7d4_.py
```diff
@@ -0,0 +1,62 @@
...
+
+# revision identifiers, used by Alembic.
+revision = '40b223faa7d4'
+down_revision = '2126386257ad'
+
+
+Session = sessionmaker()
+
+
+def update_job_name(session, old, new):
```
ivan-c: @pbugni Is a migration for a scheduled job necessary? I thought any unmentioned ones will be deleted?

achen2401: @ivan-c Ivan, yeah, I chatted with Adam about this.  It seemed *unmentioned ones* were not deleted when I re-ran the jobs.  That is why I asked Adam to help me out with writing the migration script.

OptimusRhine: Yea, I'm not sure where the disconnect here is, but both Amy and I tested this. Changing the names of the existing jobs in the ScheduledJob.json config file just leads to new jobs getting created with the new names, it doesn't remove the old ones.

We should definitely address that at some point so we don't need a migration for every job name change, but for now the migration should function as a work around.

ivan-c: Thanks for investigating! I [created an issue,](https://jira.movember.com/browse/TN-592) let me know if I missed anything.

PR #1809, opened 2018-01-24
TN-531 followup fixes - fix video layout, layout in general

achen2401:
followup fixes to TN-531: 
https://jira.movember.com/browse/TN-531

realized the absolute positioned sections in the layout of exercise and diet was making it difficult to position the video correctly so had to correct those, ahh..what a pain.

@ivan-c Ivan, this is needed for this release.


portal/gil/templates/gil/base.html
```diff
@@ -1,44 +1,44 @@
...
-    <meta name="msapplication-wide310x150logo" content="{{url_for('static', filename='img/widetile.png') }}">
-    <meta name="msapplication-square310x310logo" content="{{url_for('static', filename='img/largetile.png') }}">
+   <head>
+    {%- block head -%}
+      <meta charset="utf-8">
+      <meta http-equiv="x-ua-compatible" content="ie=edge">
+      <title>{%- block title -%}{{_("TrueNTH")}}{%- endblock -%}</title>
+      <meta name="description" content="">
+      <meta name="viewport" content="width=device-width, initial-scale=1">
+      <script src="https://cdnjs.cloudflare.com/ajax/libs/jquery/3.1.1/jquery.min.js"></script><script>window.jQuery || document.write('<script src="../scripts/jquery.min.js"><\/script>')</script>
```
ivan-c: Do we still need to bundle `../scripts/jquery.min.js` anymore?

achen2401: @ivan-c Ivan, I have removed the reference to the hard copy of jquery.min.js. I think they were adding it there for the situation when jquery is not successfully loaded from cdn site.

PR #1814, opened 2018-01-30
Feature/assessment responses report fixes

achen2401:
fix issues found in code for assessment responses report:
add check role check for added security
fix view name to conform to naming convention

### DIFF
--- a/portal/views/patients.py
+++ b/portal/views/patients.py
@@ -139,11 +139,11 @@ def patient_profile_create():
         leaf_organizations=leaf_organizations)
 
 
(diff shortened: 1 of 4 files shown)


portal/views/patients.py
```diff
@@ -139,9 +139,10 @@ def patient_profile_create():
         leaf_organizations=leaf_organizations)
 
 
-@patients.route('/sessionReport/<int:user_id>/<instrument_id>/<authored_date>')
+@patients.route('/session-report/<int:user_id>/<instrument_id>/<authored_date>')
```
ivan-c: Could you use `subject_id` instead of `user_id` where appropriate?

PR #1815, opened 2018-01-30
Fix for TN-602, TN-603 and TN-604  implement access_on_verify role

pbugni:
Fix for TN-602, TN-603 and TN-604 - namely, full implementation for MUSIC's needs of an access_on_verify role.

### DIFF
--- a/portal/migrations/versions/124cffb0fc6f_.py
+++ b/portal/migrations/versions/124cffb0fc6f_.py
@@ -1,0 +1,60 @@
+from alembic import op
+import sqlalchemy as sa
+
+
+"""Add `url_authenticated_and_verified` to encounter.auth_method enum
+
+Revision ID: 124cffb0fc6f
+Revises: 40b223faa7d4
+Create Date: 2018-01-29 16:35:25.374574
+
+"""
+
+# revision identifiers, used by Alembic.
+revision = '124cffb0fc6f'
+down_revision = '40b223faa7d4'
+
+old_auth_method_types = (
+    'password_authenticated', 'url_authenticated', 'staff_authenticated',
+    'staff_handed_to_patient', 'service_token_authenticated',
+    )
+new_auth_method_types = (
+    'password_authenticated', 'url_authenticated', 'staff_authenticated',
+    'staff_handed_to_patient', 'service_token_authenticated',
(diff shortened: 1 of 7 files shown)


portal/migrations/versions/124cffb0fc6f_.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+def downgrade():
+    # Create old type for restoration - need tmp for storage
+    tmp_type.create(op.get_bind(), checkfirst=False)
+    op.execute(
+        'ALTER TABLE encounters ALTER COLUMN auth_method TYPE _auth_methods'
+        ' USING auth_method::text::_auth_methods')
+    new_type.drop(op.get_bind(), checkfirst=False)
+
```
OptimusRhine: Do we need to convert the `auth_method` values of any Encounters using the new 'url_authenticated_and_verified' value to one of our old values (maybe 'url_authenticated') before fully converting the column back to the old Enum?

portal/views/user.py
```diff
@@ -372,9 +372,9 @@ def access_url(user_id):
     if not has.isdisjoint(not_allowed):
         abort(400, "Access URL not provided for privileged accounts")
 
-    if ROLE.WRITE_ONLY not in has:
+    if set((ROLE.ACCESS_ON_VERIFY, ROLE.WRITE_ONLY)).isdisjoint(has):
         # KEEP this restriction.  Weak authentication (which the
-        # returned URL provides) should only be available for WRITE_ONLY users
+        # returned URL provides) should only be available for these roles
         abort(400, "Account {} lacks WRITE_ONLY role".format(user_id))
```
OptimusRhine: Should this abort message reference the 'ACCESS_ON_VERIFY' role?

portal/migrations/versions/124cffb0fc6f_.py
pbugni: well - the encounter type didn't exist before.  i only would expect to rollback if a mistake was made, but I guess you're right, they could have come into existence, and then we noticed the need to rollback...

portal/views/user.py
pbugni: yup!

PR #1821, opened 2018-02-01
TN-615 allow empty study id && site id

achen2401:
address TN-615:
allow study id and site id to be changed back to empty value:

https://jira.movember.com/browse/TN-615

### DIFF
--- a/portal/static/js/main.js
+++ b/portal/static/js/main.js
@@ -4038,6 +4038,11 @@ var tnthAjax = {
         //don't think we want to translate message sent back to the server here
         params.message = "Error generated in JS - " + (hasValue(message) ? message : "no detail available");
 
+        if (window.console) {
+            console.log("Errors occurred.....");
+            console.log(params);
+        }
+
         $.ajax ({
             type: "GET",
             url: "/report-error",
--- a/portal/static/js/main.js
+++ b/portal/static/js/main.js
@@ -1894,15 +1894,20 @@ var assembleContent = {
                 };
             };
 
-            var studyId = $("#profileStudyId").val();
-            var siteId = $("#profileSiteId").val();
+            var studyIdField = $("#profileStudyId");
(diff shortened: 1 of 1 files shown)


portal/static/js/main.js
```diff
@@ -1894,30 +1894,37 @@ var assembleContent = {
                 };
             };
 
-            var studyId = $("#profileStudyId").val();
-            var siteId = $("#profileSiteId").val();
+            var studyIdField = $("#profileStudyId");
+            var siteIdField = $("#profileSiteId");
+            var hasStudyId = studyIdField.length > 0 && studyIdField.is(":visible");
```
ivan-c: Is the visibility of these fields triggered by something else you should be checking the state of instead?

portal/static/js/main.js
```diff
@@ -1894,30 +1894,37 @@ var assembleContent = {
...
+                                identifier.system != SYSTEM_IDENTIFIER_ENUM["practice_region"]) {
+                                identifiers.push(identifier);
+                            }
                         });
-                    };
-                }).fail(function() {
-                   // console.log("Problem retrieving data from server.");
+                    }
+                }).fail(function(xhr) {
+                   tnthAjax.reportError(userId, "api/demographics"+userId, xhr.responseText);
```
ivan-c: Does `tnthAjax.reportError` log to the frontend console as well? I can see error reporting being useful in both places :)

portal/static/js/main.js
achen2401: @ivan-c ivan, the fields are hidden by css in the stylesheet (e.g. study id is only visible in eproms, site id is only visible in truenthUSA) - that is why I am checking the visibility of each before updating it.

portal/static/js/main.js
achen2401: added, ivan!

PR #1822, opened 2018-02-01
TN-605 Practitioner module

OptimusRhine:
* added Practitioner class
  * stores first/last names, email, phone, and external identifiers (via 'practitioner_identifiers' table)
  * FHIR-compliant
* added Practitioner API endpoints (search, POST, PUT, GET)
  * for PUT and GET, objects can be accessed EITHER by their internal ID (`/api/practitioner/<id>`), OR by external identifier (`/api/practitioner/<value>?system=<system>`)
  * on PUT or POST, checks if identifiers in incoming JSON are already referenced by existing Practitioners (and if so, throws 409 error)
* added `practitioner_id` field to User model
  * read/writes to FHIR output/input via the 'careProvider' element, which now takes Organizations and/or a Practitioner
* added internal and external id References for Practitioner
  * internal works as normal (`[Pp]ractitioner/<id>`)
  * external expect `[Pp]ractitioner/<value>?system=<system>`
    * this logic is also now what Organization external identifier References use, instead of the previous `Organization/<system>/<value>`
* added relevant migrations
* added extensive unit tests for all the above, update some existing tests to match new logic


portal/models/practitioner.py
```diff
@@ -0,0 +1,115 @@
...
+        """Update the practitioner data from the given FHIR
+
+        If a field is defined, it is the final definition for the respective
+        field, resulting in a deletion of existing values in said field
+        that are not included.
+
+        :param fhir: JSON defining portions of the user demographics to change
+
+        """
+        def v_or_n(val):
```
pbugni: Lets move this to a better location for multiple use, as has become needed.  Not sure where though..

portal/models/practitioner.py
```diff
@@ -0,0 +1,115 @@
...
+            return val.rstrip() if val else None
+
+        if 'name' in fhir:
```
pbugni: They noticed that we didn't conform to the spec - `name` and the nested fields can be `0..*` so we're supporting accepting a list here now.  See `v_or_first()` usage in portal/models/user.py - that should also be refactored out like `v_or_n()` mentioned above.
(We have to support the list syntax as the third party FHIR lib being used packages things up by spec, so any `0..*` fields need to handle lists).

portal/views/practitioner.py
```diff
@@ -0,0 +1,311 @@
...
+        description:
+          if missing valid OAuth token or logged-in user lacks permission
+          to view requested patient
+
+    """
+    if (not request.json or 'resourceType' not in request.json or
+            request.json['resourceType'] != 'Practitioner'):
+        abort(400, "Requires FHIR resourceType of 'Practitioner'")
+    try:
+        practitioner = Practitioner.from_fhir(request.json)
```
pbugni: Won't this silently overwrite an existing one on a match?  I think we should force a PUT to update and raise if they POST on an existing, no?

OptimusRhine: This shouldn't overwrite anything existing, since this POST uses `from_fhir()` - which creates a new Practitioner and then updates it using `update_from_fhir()` - in contrast to the PUT, which actively looks up the Practitioner (using either ID or external identifier) and then just updates it using `update_from_fhir()`.

We could get _duplicates_ from this POST, but that's just because there aren't any fields we can place uniqueness constrictions on.

OptimusRhine: Per our slack convo, added checks around identifiers in PUT and POST endpoints

portal/models/practitioner.py
OptimusRhine: Makes sense; updated incoming/outgoing FHIR to match that naming convention

portal/models/practitioner.py
OptimusRhine: Moved `v_or_n()` and `v_or_first()` to `models.fhir.py` for now.

PR #1826, opened 2018-02-02
Feature/ui code for orgs fixes

achen2401:
refactor orgs ui code 
- remove global reference to orgTool object
- remove duplicate calls to populate orgs UI
- fix orgTool object self references

### DIFF
--- a/portal/templates/staff_profile_create.html
+++ b/portal/templates/staff_profile_create.html
@@ -62,6 +62,7 @@
 {% endblock %}
 {% block document_ready %}
     var orgList = {% if org_list %} {{org_list | safe}} {% else %} false {% endif %};
+    var OT = new OrgTool(); OT.init();
     //instantiation of account creation object, note, staff role is passed in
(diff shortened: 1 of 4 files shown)


portal/templates/staff_profile_create.html
```diff
@@ -62,6 +62,7 @@ <h4 class="tnth-headline left-indent-top">{{ _("New Staff") }}</h4>
 {% endblock %}
 {% block document_ready %}
     var orgList = {% if org_list %} {{org_list | safe}} {% else %} false {% endif %};
+    var OT = new OrgTool(); OT.init();
```
ivan-c: If you had an element you could check for the presence of (eg <`div id="orgList">`), could you move this initialization into `main.js`?

achen2401: @ivan-c Ivan, yeah, I see this and  some inline JS still lingering in some template files.  My next step is to clean those up as well.  Expecting a PR soon on this.

PR #1831, opened 2018-02-05
Fix practitioner PUT issue

OptimusRhine:
* resolved issue on practitioner PUT, where updating with the same external identifier info was returning a duplicate identifier error (now works as expected)
* added unit test for the above scenario

### DIFF
--- a/tests/test_practitioner.py
+++ b/tests/test_practitioner.py
@@ -270,3 +270,9 @@ class TestPractitioner(TestCase):
                                data=json.dumps(data),
                                content_type='application/json')
         self.assertEqual(resp.status_code, 409)
+
+        # test updating with same external identifier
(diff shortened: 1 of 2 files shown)


tests/test_practitioner.py
```diff
@@ -270,3 +270,9 @@ def test_practitioner_put(self):
                                data=json.dumps(data),
                                content_type='application/json')
         self.assertEqual(resp.status_code, 409)
+
+        # test updating with same external identifier
+        resp = self.client.put('/api/practitioner/{}'.format(pract2_id),
```
pbugni: Wouldn't it be best to test using the external identifier in the URL?

OptimusRhine: Sure. Those are really two different sections of the PUT code, and the external identifier query part is tested above, but I can also test it here as well

PR #1834, opened 2018-02-05
TN-650 Another exception needed to properly catch bad data in FHIR_da…

pbugni:
…tetime.parse()

This resolves a 500 raised by MUSIC when sending in another bad date format.


OptimusRhine:
Looks good to me. Not sure what kind of data they're sending to hit this, but it's definitely a situation we should be properly handling.

PR #1835, opened 2018-02-06
Fix for TN-659, which should resolve parent issue TN-535

pbugni:
To clear up confusion and simplify process when new ToUs are available, we'll now combine the `deactivate` and `notify` steps into a single scheduled job.

See the applicable list of scheduled jobs in https://jira.movember.com/browse/TN-535?focusedCommentId=134356&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-134356

### DIFF
--- a/portal/config/eproms/ScheduledJob.json
+++ b/portal/config/eproms/ScheduledJob.json
@@ -71,9 +71,11 @@
       "args": null,
       "kwargs": {
         "types": ["privacy policy"],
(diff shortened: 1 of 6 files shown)


pbugni:
@OptimusRhine i'm fine with digging it out of the history here if necessary - just assume avoid keeping code around just in case - Diana appears to be on board, but this may change.

PR #1837, opened 2018-02-06
TN-631 Track Role & Org data in piwik

OptimusRhine:
* added user role (codified/simplified to 'test', 'staff', or 'patient', or nothing) + org (in case of multiple orgs, just the first one found) to piwik push, using CustomDimensions

### DIFF
--- a/portal/gil/templates/gil/base.html
+++ b/portal/gil/templates/gil/base.html
@@ -319,7 +319,17 @@
             {% endif %}
 
             {% if user %}
-            _paq.push(["setUserId", "{{ user.id }}"]);
+                _paq.push(["setUserId", "{{ user.id }}"]);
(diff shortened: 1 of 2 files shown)


OptimusRhine:
I think that's covered by the portal_wrapper. At least, those are the only two places (gil/base and portal_wrapper) where we were sending piwik data previously, I only added the new CustomVariable fields to the existing pushes.

pbugni:
@achen2401 can you weigh in here?  Adam extended the existing piwik code as you'll see above.  Don't we need the same gil/base work in eproms/base to accurately capture eproms usage?

achen2401:
@pbugni paul, if I remember this correctly, I think adam's right - eproms/base is covered by the piwik code in portal wrapper.  we needed to do this for gil/base because gil pages don't currently pull in portal wrapper.

PR #1843, opened 2018-02-07
TN-631 Fix piwik user org push

OptimusRhine:
* fixed check/push for User Organizations in piwik data push (now use `user.organizations.first()`)


OptimusRhine:
Pushin this through to prevent further errors in stage

PR #1844, opened 2018-02-08
TN-659 take 3

pbugni:
One more round of task renaming and refactoring to fit the spec.  See description in https://jira.movember.com/browse/TN-659 (which I've been vetting w/ Diana as of late).


pbugni:
@OptimusRhine I had the same thoughts, and rather wondered why this had been done as a scheduled job initially.  My assumption was that the task does (worse case) take up to 10 seconds to complete, and the scheduled jobs interface provided the plumbing to handle that.

I think we should roll this out as is.  A more complete admin panel is in our future, but is likely to keep changing in the short run.

PR #1848, opened 2018-02-08
Wrap extensions in QNRs

ivan-c:
* Wrap extensions in outer list if need be, for QNR answers

### DIFF
--- a/portal/views/assessment_engine.py
+++ b/portal/views/assessment_engine.py
@@ -600,7 +600,19 @@ def assessment(patient_id, instrument_id):
             ].astext.endswith(instrument_id)
         )
 
-    documents = [qnr.document for qnr in questionnaire_responses]
+    documents = []
+    for qnr in questionnaire_responses:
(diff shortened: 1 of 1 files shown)


portal/views/assessment_engine.py
```diff
@@ -600,7 +600,19 @@ def assessment(patient_id, instrument_id):
...
+    documents = []
+    for qnr in questionnaire_responses:
+        for question in qnr.document['group']['question']:
+            for answer in question['answer']:
+                # Hack: Extensions should be a list, correct in-place if need be
+                # todo: migrate towards FHIR spec in persisted data
+                if (
+                    'valueCoding' in answer and
+                    'extension' in answer['valueCoding'] and
+                    isinstance(answer['valueCoding']['extension'], dict)
```
pbugni: I think the reverse check may be safer - perhaps it's always a dict, but what if it's a string, or None, or an int...
how about `... and not isinstance(answer['valueCoding']['extension'], (tuple, list))`

PR #1851, opened 2018-02-09
followup styling fixes of orgs ui in truenthusa

achen2401:
followup fix to styling of orgs ui in truenth usa


achen2401:
merging this styling fix for now

PR #1852, opened 2018-02-09
TN-666 Include status in observation handling for all APIs. 

pbugni:
Previously, the shortcuts were overlooking this detail, and although the value
quantity would return the set value, the status value of `unknown`
needs to be respected for when the user's value isn't yet known.

Was previously only working when setting the observation by id.

Extensive test coverage.  Will be further tested by CofW integration folks on the test system once rolled out.

### DIFF
--- a/portal/models/user.py
+++ b/portal/models/user.py
@@ -621,7 +621,8 @@ class User(db.Model, UserMixin):
 
         issued = fhir.get('issued') and\
             parser.parse(fhir.get('issued')) or None
-        observation = self.save_observation(cc, vq, audit, issued)
+        status = fhir.get('status')
+        observation = self.save_observation(cc, vq, audit, status, issued)
         if 'performer' in fhir:
             for p in fhir['performer']:
                 performer = Performer.from_fhir(p)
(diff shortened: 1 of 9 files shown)


portal/views/clinical.py
```diff
@@ -207,6 +210,9 @@ def pca_diag_set(patient_id):
             value:
               type: boolean
               description: the patient's PCa diagnosis
+            status:
```
ivan-c: might be worth making a schema for this to enumerate all the values we accept, but it's a pain to get just right

portal/models/user.py
```diff
@@ -676,13 +677,30 @@ def add_service_account(self):
...
-        return [obs.value_quantity for obs in self.observations if
-                obs.codeable_concept_id == codeable_concept.id]
+        matching_obs = [
+            obs for obs in self.observations if
+            obs.codeable_concept_id == codeable_concept.id]
+        if not matching_obs:
+            return None, None
+        if len(matching_obs) > 1:
+            raise ValueError(
+                "multiple observations for {} on constrianed {}".format(
```
ivan-c: >multiple observations for {} on *constrianed*

found my one change ;)

PR #1868, opened 2018-02-14
TN-617 Granular assessment status

ivan-c:
* Return lists of due assessments; partial and full

### DIFF
--- a/portal/views/assessment_engine.py
+++ b/portal/views/assessment_engine.py
@@ -1729,26 +1729,25 @@ def patient_assessment_status(patient_id):
 
     """
     patient = get_user(patient_id)
-    if patient:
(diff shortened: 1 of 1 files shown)


portal/views/assessment_engine.py
```diff
@@ -1719,7 +1719,7 @@ def patient_assessment_status(patient_id):
       - application/json
     responses:
       200:
-        description: return current overall assessment status of given patient
+        description: return current assessment status of given patient
```
pbugni: Not necessary, but it would be nice to detail what we're returning - obviously this comment only covers the generic idea.

ivan-c: thanks, I'll have to add an enum for the different "status" values, and a references for the Q/NR values

PR #1869, opened 2018-02-14
Fix TN-678 issue with card_html generation when no top_org found

OptimusRhine:
* resolved an issue with `card_html` generation when `AssessmentStatus.top_organization == None`
  * this is only an issue now because of the recent changes to the `most_current_qb()` logic, where a user with an open QNR for a QB can stay on that QB even after it's no longer referenced by an org

### DIFF
--- a/portal/models/intervention_strategies.py
+++ b/portal/models/intervention_strategies.py
@@ -234,7 +234,7 @@ def update_card_html_on_completion():
         def thank_you_block(name, registry):
             greeting = _(u"Thank you, {}.").format(name)
             confirm = _(
-                u"You've completed the {} questionnaire"
+                u"You've completed the {}questionnaire"
                 ".").format(registry)
             reminder = _(
                 u"You will be notified when the next "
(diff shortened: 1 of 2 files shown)


OptimusRhine:
This resolves the issue we're currently seeing in stg-eproms:
```
  File "/srv/www/stg-eproms.us.truenth.org/portal/portal/views/patients.py", line 169, in patient_profile
    display = intervention.display_for_user(patient)
  File "/srv/www/stg-eproms.us.truenth.org/portal/portal/models/intervention.py", line 135, in display_for_user
    if func(intervention=self, user=user):
  File "/srv/www/stg-eproms.us.truenth.org/portal/portal/models/intervention_strategies.py", line 404, in update_user_card_html
    intro=intro_html(assessment_status), header=header,
  File "/srv/www/stg-eproms.us.truenth.org/portal/portal/models/intervention_strategies.py", line 289, in intro_html
    assessment_status.top_organization.name,
AttributeError: 'NoneType' object has no attribute 'name'
```

@pbugni this is definitely the surface-level fix; if you can see a cleaner way of resolving within the AssessmentStatus model itself, free free to close this PR.

portal/models/intervention_strategies.py
```diff
@@ -286,14 +286,14 @@ def intro_html(assessment_status):
                         u"Please complete your {} "
```
OptimusRhine: Tiny detail here, but if no `assigning_authority` is found, won't these strings read as "Please complete your  questionnaire" (in other words, have two consecutive spaces)?

PR #1876, opened 2018-02-15
Part of TN-440 testing, bump only Eastern Health to IRONMAN v3

pbugni:


mcjustin:
@ivan-c  good that this is merged in for the demo release, but we do NOT want this change included in the prod update. I also need to change the USA org to NOT be protocol v3 , so I'll create a new PR off the release to accomodate both of these changes, which we can merge in after today's demo update. @pbugni  fyi

PR #1880, opened 2018-02-16
Reverts IRONMAN org/protocol changes which were intended for testing …

mcjustin:
…only (with this change, all are at protocol v2 again)


mcjustin:
Per chat, this shouldn't be merged in until *after* we've updated the demo system. This PR needs to be merged in before our prod update.

PR #1881, opened 2018-02-16
TN-120 styling fixes tou ui

achen2401:
styling fixes, including mobile view of tou, in both eproms and truenthUSA
see TN-120:
https://jira.movember.com/browse/TN-120

**For next sprint (sprint 7)**


portal/templates/initial_queries_macros.html
```diff
@@ -98,7 +98,7 @@ <h4 class="tnth-headline">{{ _("Terms of Use") }}</h4>
                         data-url="/privacy">
                           <i class="fa fa-square-o fa-lg terms-tick-box"></i>
                           {% trans privacy_url="/privacy?disableLinks=true", terms_url="/terms?disableLinks=true" %}
```
ivan-c: Could you rewrite `privacy_url` and `terms_url` to use `url_for()`? (I think you can pass querystring params as arguments to url_for())

achen2401: @ivan-c ivan, was about to fix this but then realized that this one is tricky -- privacy and terms url endpoints are different between TNUSA and EPROMS (either gil.privacy and gil.terms or eproms.privacy and eproms.terms) so this can't easily be re-written using url_for.  I think I will leave this one as is for now.

ivan-c: Fair enough :)

ivan-c: @pbugni Just curious, do you know of a clever way to do this with our blueprint setup?

PR #1883, opened 2018-02-16
fix code for task that translates po to json file

achen2401:
found issue with the gulp task that translates po to json file:
- Corrected the issue so that each po file will have its corresponding json file and consequently all json files can then be combined into one (i.e. resulted by running the combineTranslationJsons task) json file for the frontend to consume.
- re-run the tasks to make sure they are working as intended
*For next sprint (sprint 7)*


achen2401:
@ivan-c Ivan, thanks, I thought I deleted those - thank you for noticing those. I re-ran the tasks just to make sure.  Those files were leftover files when I was testing my changes.

PR #1890, opened 2018-02-21
TN-717 fix to avoid duplicate qb by intervention

pbugni:
This slipped through testing mostly due to alert fatigue.  Removed check to only log error on prod so this is seen earlier - in fact if it happens in testing, we'll now get an exception and therefore a failed test.


pbugni:
btw - I fully expect we'll start seeing error emails on staging systems when this rolls out - but we need to fix the problem, not the symptom - that is, remove organizations from accounts that have multiple.  just hides other errors as we've seen lately.

PR #1892, opened 2018-02-22
TN-713 service token management

pbugni:
New scheduled job (on both systems) to run a daily check for expired tokens (deletes stale) and nearly expired service tokens (emails sponsor).

### DIFF
--- a/portal/models/auth.py
+++ b/portal/models/auth.py
@@ -4,14 +4,18 @@ import hashlib
 import hmac
 import json
 import time
-from flask import abort, current_app
+from flask import abort, current_app, url_for
 from datetime import datetime, timedelta
+from smtplib import SMTPRecipientsRefused
 from sqlalchemy.dialects.postgresql import ENUM
 from urlparse import urlparse
 
 from ..database import db
 from ..extensions import oauth
-from .relationship import RELATIONSHIP
+from .message import EmailMessage
+from .role import Role, ROLE
+from .relationship import Relationship, RELATIONSHIP
+from .user import User, UserRoles, UserRelationship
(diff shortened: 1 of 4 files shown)


portal/models/auth.py
```diff
@@ -401,3 +405,68 @@ def create_service_token(client, user):
...
+        # Lookup the sponsor of the service account, and send em an email
+        sponsor = UserRelationship.query.join(
+            Relationship,
+            Relationship.id == UserRelationship.relationship_id).filter(
+            Relationship.name == RELATIONSHIP.SPONSOR).filter(
+            UserRelationship.other_user_id == user_id).with_entities(
+            UserRelationship.user_id)
+        if sponsor.count() != 1:
+            raise ValueError(
+                'expiring service token, cannot locate sponsor for {}'.format(
```
ivan-c: I might remove "for" to make it clear it's a user_id (not token_id)

portal/models/auth.py
```diff
@@ -401,3 +405,68 @@ def create_service_token(client, user):
...
+            "The service token in use at {app} expires {expires}.  "
+            "Please renew at {client_url}".format(
+                app=current_app.config.get('USER_APP_NAME'),
+                expires=expires,
+                client_url=url_for(
+                    'auth.client_edit', client_id=client_id, _external=True)))
+        current_app.logger.warn(body)
+        em = EmailMessage(
+            recipients=sponsor_email,
+            sender=current_app.config['MAIL_DEFAULT_SENDER'],
```
ivan-c: Might be nice to move that default into the model at some point...

PR #1893, opened 2018-02-22
clean up profile and initial queries js

achen2401:
followup clean-up of profile and initial queries js - 
remove references to global variables
retrieve config variables via ajax 
fix code indentations
fix references to this

### DIFF
--- a/portal/static/js/main.js
+++ b/portal/static/js/main.js
@@ -86,15 +86,66 @@ var ConsentUIHelper = function(consentItems, userId) {
      * relevant variables currently defined in profile_macro.html, but can be defined by consumer
      *
      */
-    this.ctop = (typeof CONSENT_WITH_TOP_LEVEL_ORG != "undefined") && CONSENT_WITH_TOP_LEVEL_ORG;
-    this.isAdmin = typeof isAdmin != "undefined" && isAdmin ? true: false;
-    this.editable = (typeof consentEditable != "undefined" && consentEditable == true) ? true : false;
-    this.consentDateEditable = this.editable && (typeof isTestPatient != "undefined" && isTestPatient);
+    this.ctop = false;
+    this.isAdmin = false;
+    this.editable = String($("#profileConsentEditable").val()).toLowerCase() === "true" ? true : false;
+    this.consentDateEditable = this.editable && (String($("#profileIsTestPatient").val()).toLowerCase() === "true");
     this.touObj = [];
     this.showInitialConsentTerms = false;
     this.hasConsent = false;
     this.hasHistory = false;
     this.orgTool = null;
+    this.initChecks = [];
+    this.userRoles = [];
+
+
+    const CONSENT_WITH_TOP_LEVEL_ORG = "CONSENT_WITH_TOP_LEVEL_ORG";
+
+
+    this.init = function(callback) {
+        var self = this;
+        self.initStartTime = new Date();
+
+        self.getOrgTool();
+        
+        /*
+         * check whether user is admin, can also check for other roles if need be
+         */
+        tnthAjax.getRoles($("#profileCurrentUserId").val(), false, function(data) {
+            if (data && data.roles) {
+                data.roles.forEach(function(role) {
+                    if (role.name.toLowerCase() === "admin") {
+                        self.isAdmin = true;
+                    }
+                    self.userRoles.push(role.name);
+                });
+            }
+            self.initChecks.push({"roleCheckDone": true});
+        });
+
+         /*
+          * get config variable, only consent with top level variable for now, but can be used to get others
+          */
+        tnthAjax.getConfiguration(CONSENT_WITH_TOP_LEVEL_ORG, self.userId, false, function(data) {
+            if (data && data.hasOwnProperty(CONSENT_WITH_TOP_LEVEL_ORG)) {
+                self.ctop = data.CONSENT_WITH_TOP_LEVEL_ORG;
+            }
+            self.initChecks.push({"configCheckDone": true});
+        }, true);
+
+        /*
+         * wait for ajax calls to finish
+         */
+        var consentVar = setInterval(function(){
+            self.initEndTime = new Date();
+            var elapsedTime = self.initEndTime - self.initStartTime;
+            elapsedTime /= 1000;
+            if (self.initChecks.length === 2 || elapsedTime >= 3) {
+                callback();
+                clearInterval(consentVar);
+            };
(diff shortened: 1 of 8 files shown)


portal/templates/profile/profile_base.html
```diff
@@ -4,6 +4,12 @@
 <form id="profileForm" class="form tnth-form to-validate" data-toggle="validator">
   <input type="hidden" id="profileUserId" value="{{user.id if user}}" />
   <input type="hidden" id="profileCurrentUserId" value="{{current_user.id if current_user}}" />
+  {% if (current_user.has_role(ROLE.STAFF) and ROLE.STAFF in config.CONSENT_EDIT_PERMISSIBLE_ROLES) or (current_user.has_role(ROLE.PATIENT) and ROLE.PATIENT in config.CONSENT_EDIT_PERMISSIBLE_ROLES) or current_user.has_role(ROLE.ADMIN) %}
+    <input type="hidden" id="profileConsentEditable" value="true" />
+  {% endif %}
+  {% if user.has_role(ROLE.PATIENT) and (current_user and not current_user.has_role(ROLE.PATIENT)) and (config.get("SYSTEM_TYPE").upper() != "PRODUCTION") %}
```
ivan-c: It would probably be more straightforward and reusable to cache the current user's roles in localStorage

portal/static/js/main.js
```diff
@@ -86,15 +86,66 @@ var ConsentUIHelper = function(consentItems, userId) {
...
+    this.init = function(callback) {
+        var self = this;
+        self.initStartTime = new Date();
+
+        self.getOrgTool();
+        
+        /*
+         * check whether user is admin, can also check for other roles if need be
+         */
+        tnthAjax.getRoles($("#profileCurrentUserId").val(), false, function(data) {
```
ivan-c: Would you be able to cache the roles in localStorage (and only make the request if necessary)?

portal/static/js/main.js
```diff
@@ -86,15 +86,66 @@ var ConsentUIHelper = function(consentItems, userId) {
      * relevant variables currently defined in profile_macro.html, but can be defined by consumer
      *
      */
-    this.ctop = (typeof CONSENT_WITH_TOP_LEVEL_ORG != "undefined") && CONSENT_WITH_TOP_LEVEL_ORG;
-    this.isAdmin = typeof isAdmin != "undefined" && isAdmin ? true: false;
-    this.editable = (typeof consentEditable != "undefined" && consentEditable == true) ? true : false;
-    this.consentDateEditable = this.editable && (typeof isTestPatient != "undefined" && isTestPatient);
+    this.ctop = false;
+    this.isAdmin = false;
```
ivan-c: It might make more sense for this to get its value from `tnthAjax.getRoles` (see below), rather than `tnthAjax.getRoles` setting `this.isAdmin`.
```
this.isAdmin = (tnthAjax.getRoles().indexOf("admin") > -1);
```

portal/static/js/main.js
```diff
@@ -5550,6 +5621,31 @@ var tnthAjax = {
...
+        }
+    },
+    "getConfiguration": function(configVar, userId, params, callback, setConfigInUI) {
+        callback = callback || function() {};
+        if (!configVar) {
+            callback({"error": "a configuration variable name is required"});
+            return false;
+        }
+        var self = this;
+        this.sendRequest("/api/settings/"+ configVar, "GET", userId, (params||{}), function(data) {
```
ivan-c: Would you be able to cache the config settings in localStorage (and only make the request if necessary)?

portal/static/js/main.js
```diff
@@ -3061,101 +3125,100 @@ var Profile = function(subjectId, currentUserId) {
...
+                        tnthAjax.postClinical(self.subjectId, toCall, "false", $(this).attr("data-status"), $(this));
+                        ["pca_diag", "pca_localized"].forEach(function(fieldName) {
+                            $("input[name='" + fieldName + "']").each(function() {
+                                $(this).prop("checked", false);
+                            });
+                        });
+                        if ($("input[name='pca_diag']").length > 0) {
+                            tnthAjax.putClinical(self.subjectId,"pca_diag","false", $(this));
+                        };
+                        if ($("input[name='pca_localized']").length > 0) {
```
ivan-c: I *think* this `if` block and its matching one in `else if (String(toCall) === "pca_diag")` could be moved to the `else` block that is containing them both (removing one copy).

achen2401:
@ivan-c ivan - I think I addressed all the issues you raised.  let me know if you see any issue with my latest changes.

portal/static/js/main.js
achen2401: @ivan-c Ivan, I think it was purposefully coded that way because we also needed to consider start treatment question - in which case we won't be setting the localized (i.e. pca_diag) variable at all.

portal/views/portal.py
```diff
@@ -696,18 +696,32 @@ def settings():
...
-    # handing out consent related keys e.g. consent_with_top_level_org
-    elif key.startswith('CONSENT_'):
-        return jsonify({key: current_app.config.get(key)})
+    if config_key:
+        key = config_key.upper()
+        # handing out LifeRay keys at this time
+        if key.startswith('LR_'):
+            return jsonify({key: current_app.config.get(key)})
+        # handing out consent related keys e.g. consent_with_top_level_org
+        elif key.startswith('CONSENT_'):
```
ivan-c: The next time you have to add another prefix/key, you might try something like this:
```
config_prefix_whitelist = (
    'REQUIRED_CORE_DATA',
    'SYSTEM',
)
if not any(key.startswith(prefix) for prefix in config_prefix_whitelist):
    abort(400, "Configuration key '{}' not available".format(key))
return jsonify({key: current_app.config.get(key)})
```

portal/static/js/main.js
```diff
@@ -5550,6 +5679,71 @@ var tnthAjax = {
...
+        } else {
+            var self = this;
+            this.sendRequest("/api/settings/"+configVar, "GET", userId, (params||{}), function(data) {
+                if (data) {
+                    callback(data);
+                    if (data.hasOwnProperty(configVar)) {
+                        if (setConfigInUI) {
+                            self.setConfigurationUI(configVar, data[configVar]+"");
+                        }
+                        sessionStorage.setItem(sessionConfigKey, JSON.stringify(data));
```
ivan-c: Is sessionStorage being cleared at logout (for security)?

PR #1898, opened 2018-02-26
TN-743 hotfix to add a swiss org (alongside existing placeholder swis…

mcjustin:
…s org)


mcjustin:
@ivan-c I'm not aware of demo usage this PM, but they will expect that we can do this w/ minimal downtime. If it was prod we'd use the /settings org updater UI. But that does seem less clean than a full update, so I'll defer to you.

PR #1903, opened 2018-03-02
TN-529 Resources Page for Eproms staff

achen2401:
address TN-529:  
https://jira.movember.com/browse/TN-529

new resources page for eproms staff


portal/eproms/views.py
```diff
@@ -255,3 +257,88 @@ def website_consent_script(patient_id):
...
+
+
+def get_any_tag_data(anyTags):
+    """
+        query LR based on any tags
+        this is an OR condition
+        will match any tag specified
+        e.g. anyTag=tag1&anyTag=tag2
+    """
+    tags_list = []
```
ivan-c: This could a little cleaner using [some functions](https://docs.python.org/3.4/library/urllib.parse.html#urllib.parse.urlencode) from the `requests` module:
```
liferay_qs_params = {
    'anyTags': anyTags,
    'sort': 'true',
    'sortType': 'DESC',
}
liferay_url = "".join(
    current_app.config["LR_ORIGIN"],
    "/c/portal/truenth/asset/query?",
    requests.compat.urlencode(
        lr_qs_params,
        doseq=True,
    ),
)
```

portal/eproms/views.py
```diff
@@ -255,3 +257,88 @@ def website_consent_script(patient_id):
...
+        resources_data = get_any_tag_data(['{} work instruction'.format(org.name.lower())])
+        results = json.JSONDecoder().decode(resources_data)['results']
+        if (len(results) > 0):
+            video_content = []
+            for asset in results:
+                if 'video' in asset['tags']:
+                    video_content.append(get_asset(asset['uuid']))
+            return render_template('eproms/resources.html',
+                                   results=results, video_content=video_content)
+        else:
```
ivan-c: If you move the "exception handling" to the top, you can un-indent all of your code, making it a little easier to understand.
eg:
```
if not org:
    abort(400, 'user must belong to an organization')
```

portal/eproms/templates/eproms/resources.html
```diff
@@ -0,0 +1,37 @@
...
+                            </li>
+                        {% endif %}
+                    {% endfor %}
+                </ul>
+            </div>
+        </div>
+    </section>
+</article>
+<script src="{{url_for('.static', filename='js/resources.js')}}"></script>
+{% endblock %}
```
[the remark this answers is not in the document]
achen2401: @ivan-c Ivan, I did make the change to add a new line at the end here..

achen2401:
@ivan-c Ivan, I think I address all the issues you raised - but let me know if you see any issue with my latest changes.  A note about passing multiple tags to LR url - query parameter for tag(s) is **anyTags**=tag1&**anyTags**=tag2 instead of the expected **anyTags**=tag1,tag2,...making reformatting a bit challenging - I am not sure using requests.compat.urlencode to serialize the parameters in that case will work.  I added comment in the code - hopefully that will make sense.

ivan-c:
@achen2401 Sorry, that example was kinda broken. The way LR works with the tag querystring parameters is pretty standard and our preferred way of passing multiple values for a single key ([apparently there's no standard consensus](https://stackoverflow.com/questions/1746507/authoritative-position-of-duplicate-http-get-query-keys)), and should be supported with `urlencode`, as long as you pass `doseq=True`.

Give this a try, or let me know if I misunderstood something, otherwise I'll merge
```
import requests
liferay_qs_params = {
    'anyTags': [
        'tag1',
        'tag2',
        'tag3',
    ],
    'sort': 'true',
    'sortType': 'DESC',
}
requests.compat.urlencode(
    liferay_qs_params,
    doseq=True,
)
```
```
>>> requests.compat.urlencode(
...     liferay_qs_params,
...     doseq=True,
... )
'sort=true&sortType=DESC&anyTags=tag1&anyTags=tag2&anyTags=tag3'
```

achen2401:
@ivan-c ivan, thank you very much.  that makes sense.  I will give it a try.

achen2401:
@ivan-c Ivan, I hope this looks better now.  I think I misunderstood what you meant before.  Let me know if you see any issue with my latest changes.

PR #1908, opened 2018-03-05
TN-761 Adding reporting views.  

pbugni:
Refactored reporting methods out of tasks module.
Added view for custom one off code previously being run by developers anytime we needed to verify counts.  Also added view to see email content generated by scheduled job.

Log in as admin to generate cvs data previously generated by hand:

  `GET /admin/overdue-numbers`  # generates a csv download - slow!

Admin, staff or staff admin can see email content - will restrict output
counts to user's orgs:
  `GET /admin/overdue-table?org_id=20000`  # for IRONMAN
  `GET /admin/overdue-table?org_id=10000`  # for TNGR

### DIFF
--- a/portal/models/reporting.py
+++ b/portal/models/reporting.py
@@ -1,19 +1,23 @@
 """Reporting statistics and data module"""
 from collections import defaultdict
 from datetime import datetime
-from flask import current_app, render_template
-from flask_babel import gettext as _
+from flask import current_app
+from smtplib import SMTPRecipientsRefused
 
+from .app_text import app_text, MailResource, SiteSummaryEmail_ATMA
 from .assessment_status import AssessmentStatus
+from ..audit import auditable_event
+from .communication import load_template_args
 from ..dogpile_cache import dogpile_cache
 from .fhir import CC
 from .intervention import Intervention
-from .organization import OrgTree
+from .message import EmailMessage
+from .organization import Organization, OrgTree
 from .procedure_codes import known_treatment_started
 from .procedure_codes import known_treatment_not_started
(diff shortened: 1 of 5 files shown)


portal/views/reporting.py
```diff
@@ -0,0 +1,141 @@
...
+            qb = a_s.qb_data.qb.name if a_s.qb_data.qb else None
+            for org in user.organizations:
+                top = ot.find_top_level_org([org])
+                org_name = "{}: {}".format(
+                    top[0].name, org.name) if top else org.name
+                cw.writerow((
+                    user.id, email, qb, a_s.overall_status, od, org_name))
+
+    output = make_response(results.getvalue())
+    output.headers['Content-Disposition'] = "attachment; filename=overdue-numbers.csv"
```
ivan-c: Might be convenient to add a timestamp to the filename

portal/models/reporting.py
```diff
@@ -119,49 +125,46 @@ def overdue_stats_by_org():
...
-
-    return render_template('site_overdue_table.html',
-                           ranges=day_ranges, rows=rows)
+    top_org = Organization.query.get(org_id)
+    if not top_org:
+        raise ValueError("No org with ID {} found.".format(org_id))
+    name_key = SiteSummaryEmail_ATMA.name_key(org=top_org.name)
+
+    for user in User.query.filter_by(deleted_id=None).all():
+        if (user.has_role(ROLE.STAFF) and user.email and (u'@' in user.email)
```
ivan-c: I like to invert these conditions and `continue` if the necessary conditions aren't met, that way you don't have all of your code indented under an `if`, making it a little easier to read

PR #1911, opened 2018-03-05
TN-751 add missing flash message html code in GIL

achen2401:
address TN-751:
https://jira.movember.com/browse/TN-751
password reset modal not appearing in GIL page because flash message html code is missing from template

### DIFF
--- a/portal/gil/templates/gil/base.html
+++ b/portal/gil/templates/gil/base.html
@@ -278,6 +278,40 @@
       </div>
     </div>
 {% endblock %}
+{# One-time system messages called Flash messages #}
+{% block flash_messages %}
+  {%- with messages = get_flashed_messages(with_categories=true) -%}
+      {% if messages %}
+          {# Hack to prevent modal if it's only successful login #}
+          {% set showMessages = true %}
+          {% if messages|first|last == "You have signed in successfully." %}
+              {% set showMessages = false %}
+          {% endif %}
+          {% if showMessages %}
+              {# Modal dialog for alerts #}
+              <div class="modal fade" tabindex="-1" role="dialog" id="alertModal">
+                <div class="modal-dialog">
+                  <div class="modal-content box-modal">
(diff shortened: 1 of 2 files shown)


portal/gil/templates/gil/base.html
```diff
@@ -278,6 +278,40 @@ <h4 class="box-modal__title">{{_("Session Timed Out")}}</h4><p class="box-modal_
       </div>
     </div>
 {% endblock %}
+{# One-time system messages called Flash messages #}
+{% block flash_messages %}
+  {%- with messages = get_flashed_messages(with_categories=true) -%}
+      {% if messages %}
+          {# Hack to prevent modal if it's only successful login #}
+          {% set showMessages = true %}
+          {% if messages|first|last == "You have signed in successfully." %}
```
pbugni: Will this work if viewing non-english?

achen2401: @pbugni paul, i think it will work, but not sure.. messages variable were assigned the text (e.g. "You have signed in successfully.") before it is translated later.  @ivan-c Ivan, am I correct here?

ivan-c: It should work as long as `messages` are translated beforehand, but I can't find where `flash(_("You have signed in successfully."))` is being called. Where is that string coming from @achen2401 ?

[See](http://flask.pocoo.org/docs/0.12/patterns/flashing/)

achen2401: @ivan, I think it is coming from part of flask user library - see /truenth/env/src/flask-user/flask_user/views.py, _do_login_user function

pbugni: Well -  let's not hold it up for now, sounds like you believe this will work in either case.  We should simply test in another language.

@ivan-c that line comes from https://github.com/lingthio/Flask-User/blob/v0.6/flask_user/views.py#L682  (do we need to consider external libraries in building .pot files?)

PR #1912, opened 2018-03-06
tn-762 2 new ironman orgs

mcjustin:
@ivan-c  Mo wants the orgs that we recently merged into master to go live on prod by mid this week (before next Monday's update). I asked them to look into the future a bit to include as many other orgs as possible with this update, thus this PR.


mcjustin:
@ivan-c  adding to my last comment here, let's hold off on merging this until tomorrow, as we may know the proper ID's for these orgs by then.

PR #1914, opened 2018-03-06
Feature/improve get user

pbugni:
Expanded `get_user` to include a number of error handling and validity checks in the new
`get_user_or_abort` function.

This refactor did reveal a number of inconsistent or missing checks for invalid user / patient IDs.

Also includes a fair amount of pep8 cleanup.

### DIFF
--- a/portal/views/patients.py
+++ b/portal/views/patients.py
@@ -1,6 +1,6 @@
 """Patient view functions (i.e. not part of the API or auth)"""
-from flask import abort, Blueprint, jsonify, render_template, request
-from flask import current_app, url_for
+from flask import abort, Blueprint, jsonify, render_template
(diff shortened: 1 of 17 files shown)


portal/views/patients.py
```diff
@@ -177,19 +173,20 @@ def patient_profile(patient_id):
...
+
 @patients.route('/treatment-options')
 @oauth.require_oauth()
 def treatment_options():
     code_list = current_app.config.get('TREATMENT_OPTIONS')
     if code_list:
         treatment_options = []
         for item in code_list:
-            code, system = item;
+            code, system = item
```
ivan-c: I think you could assign these directly in the for loop:
```
 for code, system in code_list:
```

PR #1916, opened 2018-03-06
pass as-of-date/author parameter to API regardless of entry method

achen2401:
address server errors that are generated when as-of-date/authored date is not passed to api/present-needed API

- fix is to pass as-of-date regardless of entry method


ivan-c:
@achen2401 @pbugni Fixed this in #1917, let me know if this is still applicable though

PR #1918, opened 2018-03-06
Remove card HTML, description from ePROMs AE card

ivan-c:


ivan-c:
@mcjustin I think it's probably fine to keep, at least for now, since we're not using the AE "card" in TNUSA (see `"public_access": false,`)

PR #1924, opened 2018-03-06
TN-529 use title to list resource items

achen2401:
use title of LR item instead of tag to display the name for the resource item
@mcjustin Justin, as per today's demo, we use **title** of LR item to identify each item on the resources page.

Example json returned from Liferay for resources items:

https://stg-lr7.us.truenth.org//c/portal/truenth/asset/query?anyTags=ironman%20work%20instruction


achen2401:
shoot, meant for the release, will close and recreate

PR #1931, opened 2018-03-08
Frontend identifiers saving fix

achen2401:
re-worked front-end logic so that empty strings won't be saved for identifiers (e.g. study id, site id
**for this release**

### DIFF
--- a/portal/static/js/main.js
+++ b/portal/static/js/main.js
@@ -2120,7 +2120,7 @@ var assembleContent = {
             var siteId = siteIdField.val();
 
 
-            var identifiers = null;
+            var identifiers = [];
             /*
              * get current identifier(s)
              */
@@ -2142,14 +2142,6 @@ var assembleContent = {
             /*
              * NOTE: this will save study Id or site Id only if each has a value
              * otherwise if each is empty, it will be purged from the identifiers that had older value of each
-             */
-
-            if (!identifiers) {
-                identifiers = [];
-            }
-            /*
-             * NOTE: filter out existing study id/site id as we do not want to save duplicates
-             * ALSO, we should purge if the corresponding field value is empty
              */
             identifiers = $.grep(identifiers, function(identifier) {
                 return identifier.system !== SYSTEM_IDENTIFIER_ENUM["external_study_id"] &&
--- a/portal/static/js/main.js
+++ b/portal/static/js/main.js
@@ -2139,7 +2139,7 @@ var assembleContent = {
 
             /*
              * NOTE: this will save study Id or site Id only if each has a value
-             * otherwise it will be purged from the identifiers
+             * otherwise if each is empty, it will be purged from the identifiers
              */
 
             if (hasStudyId || hasSiteId) {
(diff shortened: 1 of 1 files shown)


portal/static/js/main.js
```diff
@@ -2118,60 +2118,93 @@ var assembleContent = {
...
+            }).fail(function(xhr) {
+               tnthAjax.reportError(userId, "api/demographics"+userId, xhr.responseText);
+            });
 
+            /*
+             * NOTE: this will save study Id or site Id only if each has a value
+             * otherwise if each is empty, it will be purged from the identifiers
+             */
+
+            if (hasStudyId || hasSiteId) {
```
ivan-c: Is this condition still necessary? You check `hasStudyId` and `hasSiteId` individually inside of it (I may have misread the indentation level though...)

achen2401: @ivan-c ivan, renamed the variables so it is more clear - study Id field only exists in eproms wherease site id field only exists in truenth - so need to check each field's presence before saving the value for each

portal/static/js/main.js
```diff
@@ -2112,8 +2112,12 @@ var assembleContent = {
 
             var studyIdField = $("#profileStudyId");
             var siteIdField = $("#profileSiteId");
-            var hasStudyId = studyIdField.length > 0 && studyIdField.is(":visible");
-            var hasSiteId = siteIdField.length > 0 && siteIdField.is(":visible");
+            /*
+             * siteId field is only present in Truenth
+             * studyId field is only present in Eproms
+             */
+            var hasStudyIdField = studyIdField.length > 0 && studyIdField.is(":visible");
```
ivan-c: Would you be able to just check `studyId`, `siteId` and remove `hasStudyIdField`, `hasSiteIdField`? It doesn't seem like you the latter variables later.

portal/static/js/main.js
```diff
@@ -2112,66 +2112,103 @@ var assembleContent = {
...
+                                return identifier.system !== SYSTEM_IDENTIFIER_ENUM["external_study_id"];
+                            });
+                            identifiers.push(studyIdObj);
+                        } else {
+                            identifiers = [studyIdObj];
+                        }
                     } else {
-                        identifiers = [studyIdObj];
-                    };
+                        //filter out study ID since it is now empty
```
ivan-c: Do you still need to filter these out after Paul's backend changes to remove them?

portal/static/js/main.js
```diff
@@ -2112,66 +2112,103 @@ var assembleContent = {
...
+               tnthAjax.reportError(userId, "api/demographics"+userId, xhr.responseText);
+            });
 
-                if (hasStudyId) {
+            /*
+             * NOTE: this will save study Id or site Id only if each has a value
+             * otherwise if each is empty, it will be purged from the identifiers
+             */
+
+            if (hasStudyIdField || hasSiteIdField) {
```
ivan-c: It might be less work to create an array and push `studyIdObj`, `siteIdObj` if `studyId` and `siteId` have good values, then merge that array to `identifiers`. 

I think with @pbugni changes to the demographics API you can always expect good identifiers (no empty values), so you shouldn't have to do any filtering.

portal/static/js/main.js
achen2401: @ivan-c ivan, I think the issue was if the user had a studyid/siteid and later emptied it, we need to purge it so the older studyid/siteid won't be retained - current identifiers can hold the older value of each.

ivan-c: Ah, I was wondering if that was the case. I think it might be more appropriate to fix up the bad data with a DB migration than having your code handle every case.
@pbugni 
Do we need a migration for this?

achen2401:
@ivan-c ivan,  I cleaned up the code a bit - hopefully these changes will take - my code gets quite verbose at time :-O same as my thinking I think

ivan-c:
@achen2401 
No worries, thanks so much for making the effort! :)

PR #1933, opened 2018-03-09
TN-805 MUSIC org rename

mcjustin:


mcjustin:
right @ivan-c , not part of the RC.

PR #1937, opened 2018-03-12
TN-810 fhir datetime output includes time zone

pbugni:
Although the [FHIR spec](https://www.hl7.org/fhir/datatypes.html#dateTime) seems to imply the timezone portion of a datetime string is optional, this is giving integrators (such as MUSIC) trouble.

We'll now always include time zone info, it'll always be in UTC, and it'll always be in ISO FORMAT.

NB - this will affect ALL APIs producing datetime strings in their output, and this change should clearly be communicated to all API consumers.

This should resolve https://jira.movember.com/browse/TN-810

### DIFF
--- a/portal/date_tools.py
+++ b/portal/date_tools.py
@@ -16,13 +16,17 @@ def as_fhir(obj):
     if hasattr(obj, 'as_fhir'):
         return obj.as_fhir()
     if isinstance(obj, datetime):
-        # Make SURE we only communicate unaware or UTC timezones
+        # Make SURE we only communicate UTC timezone aware objects
         tz = getattr(obj, 'tzinfo', None)
         if tz and tz != pytz.utc:
             current_app.logger.error("Datetime export of NON-UTC timezone")
-        return obj.strftime("%Y-%m-%dT%H:%M:%S%z")
+        if not tz:
+            utc_included = obj.replace(tzinfo=pytz.UTC)
+        else:
+            utc_included = obj
+        return utc_included.isoformat()
     if isinstance(obj, date):
-        return obj.strftime('%Y-%m-%d')
+        return obj.isoformat()
 
 
 class FHIR_datetime(object):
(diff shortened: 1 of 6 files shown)


tests/test_clinical.py
```diff
@@ -119,7 +119,7 @@ def test_clinicalPUT(self):
         self.assert200(rv)
         clinical_data = json.loads(rv.data)
         self.assertEquals(clinical_data['status'], 'unknown')
-        self.assertEquals(clinical_data['issued'], new_issued)
+        self.assertEquals(clinical_data['issued'][:-6], new_issued)  # chop tz
```
ivan-c: Might be better to test that `clinical_data['issued'] == new_issued + "+00:00"`, in case we somehow aren't using UTC as the default TZ

portal/date_tools.py
```diff
@@ -16,13 +16,17 @@ def as_fhir(obj):
     if hasattr(obj, 'as_fhir'):
         return obj.as_fhir()
     if isinstance(obj, datetime):
-        # Make SURE we only communicate unaware or UTC timezones
+        # Make SURE we only communicate UTC timezone aware objects
         tz = getattr(obj, 'tzinfo', None)
         if tz and tz != pytz.utc:
             current_app.logger.error("Datetime export of NON-UTC timezone")
-        return obj.strftime("%Y-%m-%dT%H:%M:%S%z")
+        if not tz:
```
ivan-c: Is it dangerous to presume UTC if it hasn't been explicitly passed?

pbugni: We convert to UTC if given a different timezone.  We're documented to expect UTC.  I think it'd raise problems to enforce a required tz input, but that could be done.

ivan-c: >We convert to UTC if given a different timezone

Sounds good, does it make sense for that to generate an error (log) message though?

PR #1939, opened 2018-03-12
Drop microseconds from FHIR datetime output

pbugni:
ArborMetrix is using a parser that can't handle more than 3 digits of accuracy, and w/o valid need for this level of precision, we opted to remove from display alone - microseconds continue to be persisted if provided.

### DIFF
--- a/tests/test_fhir.py
+++ b/tests/test_fhir.py
@@ -124,28 +124,40 @@ class TestFHIR(TestCase):
         self.assertIn(str(qr.subject_id), qr_str)
         self.assertIn(str(qr.status), qr_str)
 
-    def test_tz_aware_conversion(self):
-        eastern = pytz.timezone('US/Eastern')
-        aware = datetime(2016, 7, 15, 9, 20, 37, 0, eastern)
-        parsed = FHIR_datetime.parse(aware.strftime("%Y-%m-%dT%H:%M:%S%z"))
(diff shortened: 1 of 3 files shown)


tests/test_fhir.py
```diff
@@ -124,28 +124,40 @@ def test_qr_format(self):
...
+    isostring = FHIR_datetime.as_fhir(unaware)
+    assert isostring[-6:] == '+00:00'
+
+
+def test_dt_rounding():
+    """Microseconds should be rounded in FHIR output"""
+    sample = datetime.utcnow()
+    sample = sample.replace(tzinfo=pytz.utc)
+
+    assert sample.microsecond != 0
```
ivan-c: I hope this test doesn't fire at the wrong time ;D, might better to make it deterministic

tests/test_fhir.py
```diff
@@ -124,28 +124,40 @@ def test_qr_format(self):
...
+def test_tz_aware_output():
+    """FHIR_datetime.as_fhir() should always be tz aware (UTC)"""
+    unaware = datetime(2016, 7, 15, 9, 20, 37, 0)
+
+    # generate fhir - expect UTC tz info at end
+    isostring = FHIR_datetime.as_fhir(unaware)
+    assert isostring[-6:] == '+00:00'
+
+
+def test_dt_rounding():
```
ivan-c: This is actually truncating (not rounding), isn't it?

PR #1940, opened 2018-03-13
MUSIC config settings belong in Truenth (GIL) config, not eproms.

pbugni:
Obviously, these belong in the GIL config - moving over.

@mcjustin Sorry for this oversight - I believe I had run my tests from an eproms config (and therefore hit MUSIC config values), but this would mean we were not accessing the correct email content on stg.us.


mcjustin:
@pbugni looks good to me

PR #1947, opened 2018-03-14
TN-823 handle bogus email in invite API

pbugni:


portal/models/user.py
```diff
@@ -201,6 +201,23 @@ def default_email(context=None):
     return value
 
 
+def validate_email(email):
+    """Not done at model level, as there are exceptions
+
+    We allow for placeholders and masks on email, so not all emails are valid.
```
ivan-c: Wouldn't it make more sense to check for the masking/placeholder pattern, if this is to specifically check that (not whether an email is generally valid or not)?

pbugni: this is only being used when we need a valid email - which is why it isn't plugged in at the model class

PR #1949, opened 2018-03-14
Tn 824 music sub orgs

mcjustin:


portal/config/gil/Organization.json
```diff
@@ -1946,6 +1946,120 @@
         }
       ],
       "use_specific_codings": false
+    },
+    {
+      "address": [],
```
pbugni: We don't need to include empty fields.  Should only define desired values.

portal/config/gil/Organization.json
```diff
@@ -1946,6 +1946,120 @@
...
+        }
+      ],
+      "indigenous_codings": true,
+      "name": "Michigan Medicine - Department of Urology",
+      "partOf": {
+        "reference": "api/organization/40001"
+      },
+      "race_codings": true,
+      "resourceType": "Organization",
+      "telecom": [
```
pbugni: Can skip this whole `telecom` block.

pbugni:
I'm rolling this out before the tests finish - as MUSIC is waiting on it.

PR #1954, opened 2018-03-16
Feature/music send email ui changes

achen2401:
changes based on specifications in this story: 
https://jira.movember.com/browse/TN-609


portal/static/js/main.js
```diff
@@ -2592,7 +2592,53 @@ var Profile = function(subjectId, currentUserId) {
...
+                    $("#assessmentStatusContainer .assessment-label")
+                    .html(hasReports?i18next.t("complete"):i18next.t("incomplete"))
+                    .addClass(hasReports?"text-success":"text-warning")
+                    .removeClass(!hasReports?"text-success": "text-warning");
+                }
+            });
+        }
+
+        if ($("#registrationStatusContainer").length > 0) {
+            tnthAjax.getRoles(self.subjectId, false, function(data) {
```
ivan-c: It might be convenient to implement a function (or extend `getRoles()`) so you can pass a specific role and get a true/false return instead of having to filter the results every time.

portal/static/js/main.js
```diff
@@ -2592,7 +2592,53 @@ var Profile = function(subjectId, currentUserId) {
...
+                if (!data.error) {
+                    var pcpReports = data["user_documents"]? $.grep(data["user_documents"], function(document) {
+                        return /P3P/gi.test(document.filename);
+                    }): false;
+                    var hasReports = pcpReports && pcpReports.length > 0;
+                    /*
+                     * adjust select/send email UI only if email is available
+                     */
+                    if (hasValue($("#email").val())) {
+                        if (!hasReports) {
```
ivan-c: It might make sense to have a single if/else block on `hasReports`, then do these specific checks (eg `hasValue($("#email").val())`) inside that block, rather than checking `hasReports` multiple times.

portal/templates/profile/profile_macros.html
```diff
@@ -240,29 +240,36 @@ <h4 id="patientDetailLoc" class="profile-item-title index-item">
         <div class="help-block with-errors" id="erroremail"></div>
     </div>
 {%- endmacro %}
+{% macro render_emailSelector(prefix='registration', status_title='Registration Status:', label='not registered') -%}
+    <div id="{{prefix}}StatusContainer" class="{{prefix}}-status-container">
+        <span class="text-muted">{{_(status_title)}}</span>
```
ivan-c: I think you'll need to move the `_()` call to the macro definition line (see below) for those strings to be properly extracted
`{% macro render_emailSelector(prefix='registration', status_title=_('Registration Status'), label=_('not registered')) -%}`

portal/templates/profile/profile_macros.html
```diff
@@ -240,29 +240,36 @@ <h4 id="patientDetailLoc" class="profile-item-title index-item">
...
-                    {{_("Send email to user")}}
-                {%-endif-%}
-            </label>
-            <select  id="profileEmailSelect" class="form-control bd-element" aria-labelledby="sendEmailLabel">
-                <option value="">{{ _("Select Email...") }}</option>
-                <option value="invite" data-url="{{url_for('portal.patient_invite_email', user_id=person.id)}}" {%if not person.has_role('write_only')%}class="inactive"{%endif%}>{{_("TrueNTH Invite")}}</option>
-                <option value="reminder" data-url="{{url_for('portal.patient_reminder_email', user_id=person.id)}}">{{_("TrueNTH Reminder")}}</option>
-            </select>
+            {% if person.first_top_organization() and (person.first_top_organization().name == config.ACCEPT_TERMS_ON_NEXT_ORG) %}
+                {% call render_emailSelector(prefix='assessment', status_title='P3P Assessment Status:', label='not available') %}
```
ivan-c: Same issue as above, `'P3P Assessment Status:'` should be wrapped with `_()`

PR #1956, opened 2018-03-20
Fix TN-848 - handle integer input for boolean value quantities

pbugni:


pbugni:
FWIW, did also test from test intervention - passing a "1" as MUSIC intends to do correctly sets the respective clinical observation value True.

PR #1964, opened 2018-03-22
Remove outer "questionnaire" wrap as it's breaking MUSIC's parser.

pbugni:


pbugni:
Pushing out for live testing by MUSIC.

PR #1969, opened 2018-03-22
Add en_GB as a clone of en_AU (as needed by Southampton)

pbugni:
@ivan-c more to come in this PR - but here's the addition of the language code for now, for your review.


pbugni:
This should be ready to go - I'm working on automated tests that read in the config, but it is believed to be code complete.

PR #1977, opened 2018-03-26
TN-543 ironman language options

mcjustin:


portal/config/eproms/Organization.json
```diff
@@ -443,6 +443,11 @@
                 "code": "en_US",
                 "display": "American English",
                 "system": "urn:ietf:bcp:47"
+              },
+              {
+                "code": "es_MX",
```
ivan-c: Could you change that to `es_US`? `es_US` is what's configured in Smartling

mcjustin: done

PR #1997, opened 2018-04-02
TN-914 fix.  CLI for manual password reset.

pbugni:


pbugni:
@mcjustin the one mention of `go` you referenced seems about as clear as I can make it.  How would this be better?  Do you want a link somewhere else?

PR #2002, opened 2018-04-04
Fix TN-947 - don't include observations with null issued when looking…

pbugni:
… for trigger date.


mcjustin:
@pbugni  I was just commenting - seeing that the function fetch_datetime_for_concept() is only called from one place, this seems safely isolated to me...

PR #2016, opened 2018-04-10
TN-944 fix registration status display on UI

achen2401:
fix registration status display on profile page
add check for registration status on whether to display reset password UI
remove unused/unnecessary code

### DIFF
--- a/portal/models/user.py
+++ b/portal/models/user.py
@@ -359,13 +359,10 @@ class User(db.Model, UserMixin):
                         self, non_registered_roles))
 
         # Still here implies not yet registered, enforce role presence
-        if not disjoint:
+        if not current_roles:
             return False
-        else:
-            if not current_roles:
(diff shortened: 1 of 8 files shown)


portal/models/user.py
```diff
@@ -362,6 +362,8 @@ def is_registered(self):
         if not disjoint:
             return False
         else:
+            if not current_roles:
```
pbugni: No, I don't believe we want this clause.  The point is, if we created a non-registered user somehow, they should have one of the `PRE_REGISTERED_ROLES`.
Am I overlooking a legitimate use case @achen2401 ?

pbugni:
@achen2401 I see - yes, I introduced that test breakage by not altering the default test user to appear as a registered user.  Just added a commit to this branch - will merge once the tests pass.  Thanks.

PR #2018, opened 2018-04-11
Expose config vars as environment variables

ivan-c:
* Expose LR config variables
* Remove default `SECRET_KEY`
  * NB: existing checkouts that do not have `SECRET_KEY` will error until set
* Cleanup config vars

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,7 @@ env:
     - TOXENV=py27
   global:
     - FLASK_APP="$(readlink -f manage.py)"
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -17,6 +17,7 @@ env:
     - TOXENV=py27
   global:
     - FLASK_APP="$(readlink -f manage.py)"
+    - SECRET_KEY="$(openssl rand -base64 32)"
```
pbugni: oh i *love* this approach!

PR #2026, opened 2018-04-13
TN-905 offer registration given user_document for `access_on_verify`

pbugni:


portal/templates/flask_user/register.html
```diff
@@ -4,7 +4,28 @@
 {% from "flask_user/_macros.html" import render_field, render_submit_field, back_btn %}
 <form id="regForm" action="" method="POST" novalidate class="form to-validate" style="margin-left: 1em">
     <div class="row">
-        <div class="col-md-push-3 col-md-5 col-xs-11">
+        {% set custom_invite_check = (not form.email.errors) and session.get('registration.invite_post_p3p', '') %}
```
ivan-c: I think the empty string default (`''`) is unnecessary here since `None` will evaluate to `False`, as `''` does

PR #2030, opened 2018-04-16
Don't upgrade system-installed version of pip

ivan-c:
* Workaround current [issue with pip >= 10](https://github.com/pypa/pip/issues/5240)

### DIFF
--- a/docker/Dockerfile.build
+++ b/docker/Dockerfile.build
@@ -25,7 +25,7 @@ ENV \
     ARTIFACT_DIR="${ARTIFACT_DIR:-/tmp/artifacts}"
 
 RUN \
-    pip install --upgrade pip setuptools && \
+    pip install --upgrade setuptools && \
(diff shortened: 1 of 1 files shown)


pbugni:
I believe this was a temporary upstream problem with `pip`.  I'd rather see us continue to use the upgraded pip unless you have reason to believe this is a bad choice, @ivan-c ?

ivan-c:
@pbugni From [this comment](https://github.com/pypa/pip/issues/5240#issuecomment-381673334), it sounds like the `pip` installed from debian package (system version of `pip`) shouldn't be modified, or doesn't expect to be run as root. 

I think it's against recommended best practice to run `pip` as root to install python modules system-wide , since python modules come from an untrusted source. This is happening isolated inside a docker image, but is still swimming upstream, so I relented since the default `pip` inside debian stretch docker images should be good enough (9.0.1)

PR #2033, opened 2018-04-17
TN-979 endpoint to register un-registered accounts

pbugni:


portal/views/user.py
```diff
@@ -1067,6 +1076,50 @@ def relationships(user_id):
...
+@user_api.route('/user/register-now')
+@oauth.require_oauth()
+def register_now():
+    """Target for triggering registration of account
+
+    Some flows generate accounts that are not yet ``registered``,
+    such as when given the ``access_on_verify`` role.  When
+    such a user is "logged in" (that is, given access after
+    verifying basic information), redirect to this endpoint
+    to trigger promotion to a registered account.
```
mcjustin: I'd prob change this from 'When
+    such a user is "logged in" (that is, given access after
+    verifying basic information), redirect to this endpoint
+    to trigger promotion to a registered account.'
to 
"When it's desirable to promote the user to a registered account (eg when they've completed a flow like MUSIC P3P, where stakeholders wanted to avoid the potential disruption of registration), redirect to this endpoint to trigger promotion to a registered account.''

pbugni: @mcjustin thanks - commit made with those changes.

PR #2035, opened 2018-04-18
TN-979 Must log user out before sending to user.register

pbugni:


mcjustin:
Looks good to me @pbugni

PR #2036, opened 2018-04-18
TN-979 take 3

pbugni:
Few bugs fixed:
1. include email on redirect to user.register, so it's correctly pre populated
2. exclude ALL the special roles when promoting the account.


pbugni:
@ivan-c sorry for the haste, but we need this on stg, not waiting for tests.

PR #2037, opened 2018-04-18
TN-983 fix locale selection for i18next config

achen2401:
fix default locale selection for i18next
linting some JS issues (e.g. unnecessary comma, etc., single quotes)
move document ready block to bottom


achen2401:
s

PR #2050, opened 2018-04-19
TN-996 translate strings used in variable substitution for email

pbugni:


portal/models/communication.py
```diff
@@ -67,14 +68,15 @@ def _lookup_assessment_button():
         return make_button(_lookup_assessment_link())
 
     def _lookup_assessment_link():
+        label = _(u'Compete Questionnaire')
```
ivan-c: minor typo

PR #2060, opened 2018-04-21
fix self reference error

achen2401:
fixed undefined js error raised when choosing an org.
**for this release**

### DIFF
--- a/portal/static/js/main.js
+++ b/portal/static/js/main.js
@@ -5856,6 +5856,7 @@ var tnthAjax = {
     },
     "getConfigurationByKey": function(configVar, userId, params, callback, setConfigInUI) {
         callback = callback || function() {};
+        var self = this;
         if (!userId) {
             callback({"error": i18next.t("User id is required.")});
             return false;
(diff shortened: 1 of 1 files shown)


portal/static/js/main.js
```diff
@@ -5856,6 +5856,7 @@ var tnthAjax = {
     },
     "getConfigurationByKey": function(configVar, userId, params, callback, setConfigInUI) {
         callback = callback || function() {};
+        var self = this;
```
pbugni: I wouldn't know, but isn't the `self` convention a python thing?  Why not just use `this` in place of `self` below?

I also see [this Stack Overflow explanation](https://stackoverflow.com/a/16876159), suggesting `self` has a built in meaning (unless overwritten as is done here).

achen2401: @pbugni  paul - I need to preserve  **this**  because the execution context has changed -  **this** in the anonymous function callback now no longer reference the TnthAjax object that initiated the call.   The undefined error was henceforth raised.   http://davidshariff.com/blog/javascript-this-keyword/  has a discussion about it.

PR #2067, opened 2018-04-24
TN-1020 Fix - error text now includes `languageId`:

pbugni:
"Unexpected exception in `send_queued_communications` on 5 : https://stg-lr7.us.truenth.org/c/portal/truenth/asset/mail?version=2.4&uuid=2c837970-71e3-d736-9d2f-626f417bbbad&languageId=de_CH contains unknown variables: First_name,Last_name"

@ivan-c I don't believe we need this as a hotfix, but it will certainly be useful in tracking down bad variables in localized content.

### DIFF
--- a/portal/models/communication.py
+++ b/portal/models/communication.py
@@ -253,7 +253,7 @@ class Communication(db.Model):
         if missing:
             raise ValueError(
                 "{} contains unknown varables: {}".format(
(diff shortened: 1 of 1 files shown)


ivan-c:
@pbugni switched to hotfix, see #2070

PR #2077, opened 2018-04-26
TN-1036 GET/POST identifier API

pbugni:
As a workaround to PUT /api/demographics, this PR enables POSTing additional identifiers on user.

### DIFF
--- a/portal/views/identifier.py
+++ b/portal/views/identifier.py
@@ -1,0 +1,140 @@
+"""Identifier API"""
+from flask import abort, Blueprint, jsonify, request
+
+from ..database import db
+from ..extensions import oauth
+from ..models.identifier import Identifier
+from ..models.user import current_user, get_user_or_abort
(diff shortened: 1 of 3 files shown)


portal/views/identifier.py
```diff
@@ -0,0 +1,140 @@
...
+                  - value
+                properties:
+                  system:
+                    type: string
+                    description: uri namespace for the identifier value
+                  value:
+                    type: string
+                    description: unique value for given system namespace
+                  use:
+                    type: string
```
ivan-c: [`enum`](https://swagger.io/docs/specification/data-models/enums/) might be better here, unless we really do accept any value

pbugni: there is an enum used at the ORM level to constrain this field (see models.identifier `identifier_use`) - the error will bubble up if used incorrectly.  i don't think it's worth the effort of keeping them in sync to define a second enum at the UI level.

PR #2097, opened 2018-05-04
TN-893, TN-934 remove sql from audit log on save observation

pbugni:


portal/models/codeable_concept.py
```diff
@@ -26,8 +26,8 @@ class CodeableConcept(db.Model):
 
     def __str__(self):
         """Print friendly format for logging, etc."""
-        summary = "CodeableConcept {} [".format(
-            self.text if self.text else '')
+        text = "{} ".format(self.text) if self.text else ''
+        summary = "CodeableConcept {}[".format(text)
```
ivan-c: Is that `[` errant?

pbugni: No, you'll see a matching one just below - used to bracket the `codings` connected to the CodableConcept.  Perhaps they should be parens...?  Anyhow, that is how it has been till now.

PR #2108, opened 2018-05-08
TN-1070 handle message localization needs outside of request context

pbugni:


portal/models/communication.py
```diff
@@ -235,35 +235,40 @@ def __str__(self):
             ' of {0.communication_request.name}'.format(self))
 
     def generate_message(self):
-        "Collate message details into EmailMessage"
+        """Collate message details into EmailMessage"""
+
+        def localize_message():
```
ivan-c: Why aren't you passing the variables this function depends on as arguments?

PR #2119, opened 2018-05-12
Pin pip version when building debian package

ivan-c:
* Don't use newest version of pip when building `portal` debian package
  * Newer versions of pip (10+) prefer installing portal as wheel over egg (see below)
* `pkginfo` reads version metadata from egg (`portal.egg-info/PKG-INFO`)
  * `pkginfo` can't read wheel package metadata (`dist-info` directory?)

Todo: 
* Figure out wheel packaging
  * `python setup.py bdist_wheel`

[Pip 10.0 changelog](https://pip.pypa.io/en/stable/news/#id5)
>Installing from a local directory or a VCS URL now builds a wheel to install, rather than running setup.py install. Wheels from these sources are not cached. (#4501)

### DIFF
--- a/debian/rules
+++ b/debian/rules
@@ -2,10 +2,8 @@
 
 DH_VIRTUALENV_ARGS = --preinstall='pip>=8,<10.0' --preinstall='setuptools>=12'
 
-
 distclean:
 	rm -Rf build debian/portal/ env
-
 
 override_dh_virtualenv:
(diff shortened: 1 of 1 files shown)


ivan-c:
Might just be an issue with using `setuptools_scm` incorrectly, [todo try setting version with `__version__`](https://github.com/pypa/setuptools_scm#setuppy-usage)

PR #2123, opened 2018-05-15
TN-1108 refinements

achen2401:
- add translation fixes to tables in profile
- fix display and un-complicate event logics in clinical questions section

### DIFF
--- a/portal/static/js/profile.js
+++ b/portal/static/js/profile.js
@@ -172,6 +172,15 @@
                     $(".pagination-detail .pagination-info").html(rowInfo);
                     return rowInfo;
                 },
+                formatAllRows: function() {
+                    return i18next.t("All rows");
+                },
+                formatSearch: function() {
+                    return i18next.t("Search");
+                },
+                formatNoMatches: function() {
+                    return i18next.t("No matching records found");
+                },
                 formatRecordsPerPage: function(pageNumber) {
                     return i18next.t("{pageNumber} records per page").replace("{pageNumber}", pageNumber);
                 },
@@ -671,9 +680,6 @@
                     break;
                 case "auditlog":
                     this.initAuditLogSection();
-                    break;
-                case "biopsy":
-                    this.initBiopsySection();
                     break;
                 }
             },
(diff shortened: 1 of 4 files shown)


portal/static/js/profile.js
```diff
@@ -1873,15 +1879,18 @@
                 });
             },
             updateClinicalSection: function(data) {
+                if (!data) { return false; }
```
ivan-c: I love this pattern of exiting ASAP!

portal/static/js/profile.js
```diff
@@ -1939,157 +1948,126 @@
...
                 if (treatmentCode) {
-                    if (String(treatmentCode) === String(this.modules.SYSTEM_IDENTIFIER_ENUM.CANCER_TREATMENT_CODE)) {
-                        $("#tx_yes").prop("checked", true);
+                    var hasCancerTreatment = String(treatmentCode) === String(this.modules.SYSTEM_IDENTIFIER_ENUM.CANCER_TREATMENT_CODE);
+                    $("#tx_yes").prop("checked", hasCancerTreatment);
+                    $("#tx_no").prop("checked", !hasCancerTreatment);
+                }
+            },
+            onBeforeInitClinicalQuestionsSection: function() {
+                if (this.mode === "profile") {
```
ivan-c: If you're not going to use an `else`, it might be simpler to exit. Then you wouldn't need to indent the whole next section
```
if (this.mode !== "profile") { return; }
```

portal/static/js/profile.js
```diff
@@ -1939,157 +1948,126 @@
...
-                                                return true;
-                                            } else {
-                                                self.modules.tnthAjax.postClinical(self.subjectId, toCall, toSend, "", targetContainer, {
-                                                    "issuedDate": $("#biopsyDate").val()
-                                                });
-                                            }
-                                        }
-                                        thisItem.parents(".pat-q").nextAll().fadeIn();
+                        self.onBeforeInitClinicalQuestionsSection();
+                        if (self.subjectId) {
```
ivan-c: Might be able to do similarly here, the diff makes it hard to be sure though...

portal/static/js/profile.js
```diff
@@ -1939,157 +1948,126 @@
...
-                                    if (!$("#patientQ input[type='radio']").is(":checked")) {
-                                        $("#patientQContainer").append("<span class='text-muted'>" + i18next.t("no answers provided") + "</span>");
-                                    }
+                            });
+                            self.__convertToNumericField($("#biopsy_day, #biopsy_year"));
+                            $("#biopsy_day, #biopsy_month, #biopsy_year").on("change", function() {
+                                var d = $("#biopsy_day"), m = $("#biopsy_month"), y = $("#biopsy_year");
+                                var isValid = self.modules.tnthDates.validateDateInputFields(m, d, y, "biopsyDateError");
+                                if (isValid) {
+                                    $("#biopsyDate").val(y.val()+"-"+m.val()+"-"+d.val());
```
ivan-c: Might be nice to use a datetime library for formatting, can you use [`toLocaleDateString()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleDateString)?

achen2401: @ivan-c Ivan, I think so. I did try but LocaleDateString() is not returning datestring in ios short format (yyyy-mm-dd), at least it is not in chrome or firefox - a discussion about the inconsistencies:  https://stackoverflow.com/questions/27939773/tolocaledatestring-short-format.  Think want to leave it unchanged for now just to be safe.

ivan-c: @achen2401 Thanks, it looks like there's not an easy way to get the formatting you want :(
I found this while looking, but it's only *slightly* better :/
```
var date = Date();
date.toISOString().split('T')[0];
```

PR #2141, opened 2018-05-18
portal wrapper - timeout cookie read fix

achen2401:
found issue with reading cookie(s) from document.cookie in portal wrapper - it is not returning all cookies, due to this I think: https://stackoverflow.com/questions/1022112/why-doesnt-document-cookie-show-all-the-cookie-for-the-site

add the workaround so the cookie value will be used when present

### DIFF
--- a/portal/templates/portal_wrapper.html
+++ b/portal/templates/portal_wrapper.html
@@ -144,7 +144,7 @@
         </div>
     </div>
 </div>
-<input type="hidden" id="sessionMonitorProps" data-baseurl="{{PORTAL}}" data-crsftoken="{{csrf_token()}}" data-sessionlifetime="{{config.PERMANENT_SESSION_LIFETIME if config.PERMANENT_SESSION_LIFETIME}}" />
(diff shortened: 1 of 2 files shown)


portal/templates/portal_wrapper.html
```diff
@@ -144,7 +144,7 @@
         </div>
     </div>
 </div>
-<input type="hidden" id="sessionMonitorProps" data-baseurl="{{PORTAL}}" data-crsftoken="{{csrf_token()}}" data-sessionlifetime="{{config.PERMANENT_SESSION_LIFETIME if config.PERMANENT_SESSION_LIFETIME}}" />
+<input type="hidden" id="sessionMonitorProps" data-baseurl="{{PORTAL}}" data-crsftoken="{{csrf_token()}}" data-sessionlifetime="{{config.PERMANENT_SESSION_LIFETIME if config.PERMANENT_SESSION_LIFETIME}}" data-cookie-timeout="{{request.cookies.get('SS_TIMEOUT')}}"/>
```
ivan-c: I might change those config lookups to fail if they're not configured, since those config values are required ie:
```
{{request.cookies['SS_TIMEOUT']}}
{{config.PERMANENT_SESSION_LIFETIME}}
```

PR #2148, opened 2018-05-21
TN-1150 Add user method to determine if they are `email_ready`

pbugni:


tests/test_user.py
```diff
@@ -1204,3 +1204,16 @@ def test_invite(self):
...
+        # with valid email and `NO_CHALLENGE_WO_DATA` set, should pass
+        self.assertTrue(self.app.config.get("NO_CHALLENGE_WO_DATA"))
+        self.assertTrue(self.test_user.email_ready())
+
+        # alter config and expect failure w/o all DOB, first & last name
+        self.app.config['NO_CHALLENGE_WO_DATA'] = False
+        self.assertFalse(self.test_user.email_ready())
+
+        # set birthday and expect a pass
+        self.test_user.birthdate = '1976-04-01'
```
ivan-c: Would it make sense to explicitly set the other fields required to validate `email_ready()`? 
I think we're relying on the user to already be setup as such before this test

pbugni: oh, sure.  too many assumptions in there already

PR #2149, opened 2018-05-22
initial queries bug fix and streamline code

achen2401:
found issue while testing -
clearing session storage when performing login-as as a switch of user has occurred
streamline code in initial queries - also use web worker if possible to initialize data


achen2401:
found issue with the fix - will reopen when resolved.

PR #2150, opened 2018-05-22
TN-1157 guard /api/user/register-now from invalid use

pbugni:
Now confirming target user is `email_ready()` before sending them down a registration path they can't complete.


pbugni:
@ivan-c merging now as i want to extend this in other PRs.  Curious if we need to do anything extra to push the new translated strings added to `models/user.py`

PR #2152, opened 2018-05-23
Feature/js refactor 6

achen2401:
initial queries  - bug fix (i.e., remove settimeout code for initializing sections (scope issue)) and streamline code
main.js - streamline code
procedures.js - bug fix (i.e. remove eventloader when no procedure is found), use web worker if possible for loading treatment options
profile.js - bug fix, don't send birthday data if empty'
remove unnecessary parameter from profileOrgs template code


portal/templates/profile/my_profile.html
```diff
@@ -42,7 +42,7 @@
                     <div class="profile-item-container {% if user and user.has_role(ROLE.ADMIN) %}editable{% endif %}" data-sections="org">
                         {{profile_macro.profileEditButton()}}
                         {% call profile_macro.titleSection() -%}{{ _("My Clinic") }}{%- endcall %}
-                        {{profile_macro.profileOrg(user,None,consent_agreements, current_user)}}
+                        {{profile_macro.profileOrg(user,consent_agreements, current_user)}}
```
ivan-c: Can you use [keyword arguments](https://docs.python.org/3/glossary.html#term-argument) for these function calls? 

They're self-documenting and don't require you to pass `None` if you don't want to include an argument
eg:
```
{{profile_macro.profileOrg(person=user, consent_agreements=consent_agreements, current_user=current_user)}}
```

PR #2155, opened 2018-05-23
TN-1107 disable send email button(s) for not email-ready users

achen2401:
TN-1107:  https://jira.movember.com/browse/TN-1107

### DIFF
--- a/portal/templates/profile/profile_macros.html
+++ b/portal/templates/profile/profile_macros.html
@@ -301,7 +301,7 @@
                 </select>
             </div>
         </div>
-        <button id="btnProfileSend{{prefix}}Email" class="btn-send-email btn btn-tnth-primary" type="button" v-bind:class="{disabled:!demo.data.email}" v-bind:disabled="!demo.data.email">{{ _("Send email") }}</button>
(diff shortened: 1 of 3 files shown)


portal/templates/profile/profile_macros.html
```diff
@@ -456,7 +456,7 @@ <h4 id="communicationsLoc" class="profile-item-title index-item">{{ _("Communica
 
 {% macro resetPassword(person) %}
```
ivan-c: Just curious, would it be possible to move any VueJS-only template code to the `static/` directory with the rest of the frontend files, or does the HTML still need to be served by Flask?

I think we might be able reduce the complexity of jinja/vue/jquery if they only worked as VueJS templates, but I don't think it would be worth the effort if the complexity of loading them is the same.

achen2401: @ivan-c ivan, yeah, I think that is possible.  But the snippet is part of the jinja template right now so harder to convert/move- can possibly convert the entire macro to a Vue component  (https://vuejs.org/v2/guide/components.html) but that takes some planning.

ivan-c: @achen2401 Cool, thanks for the info! Not a high priority, but any progress we can make to separating the front and backend will make them both easier to maintain :)

PR #2160, opened 2018-05-24
TN-1107 email ready message styling fix

achen2401:
- fix styling of message display when user does not have email ready status
- move code to macro


portal/templates/profile/profile_macros.html
```diff
@@ -24,6 +24,9 @@ <h4 id="{%-if id -%}{{id}}{%- endif -%}" class="profile-item-title index-item {{
 {% macro saveLoaderDiv(container_id="") -%}
     <div class="load-container save-loader-wrapper"><i id="{{container_id}}_load" class="fa fa-spinner fa-spin load-icon fa-lg save-info" aria-hidden="true"></i><i id="{{container_id}}_success" class="fa fa-check success-icon save-info" aria-hidden="true">{{_("Updated")}}</i><i id="{{container_id}}_error" class="fa fa-times error-icon save-info" aria-hidden="true">'{{_("Unable to Update. System error.")}}</i></div>
 {%- endmacro %}
+{% macro emailReadyMessage() -%}
+    <email-ready-message v-bind:message="userEmailReadyMessage"></email-ready-message>
```
ivan-c: Is that a custom HTML element? 
Naive question, but why would you use this over something like `<details>`?

ivan-c: Cool! Sounds like a really nice way to re-use components :)

PR #2161, opened 2018-05-24
TN-1170 fix problems with marking patient undead with pre-existing deceasedDateTime

pbugni:


portal/models/user.py
```diff
@@ -1156,6 +1156,23 @@ def update_birthdate(self, fhir):
                        "'%Y-%m-%d'".format(fhir['birthDate']))
 
     def update_deceased(self, fhir):
+        # As the update process starts with a complete record of the
+        # current patient and then merges in given fields, it's possible
+        # to land here with conflicting data.  Process boolean first
+        # as it may be clearing a previously set datetime.
+
+        if fhir.get('deceasedBoolean', None) is False and self.deceased_id:
```
ivan-c: I think that `None` param isn't required
```
>>> type({}.get('foo'))
<type 'NoneType'>
```

tests/test_demographics.py
```diff
@@ -428,3 +428,24 @@ def test_alt_phone_removal(self):
...
+        with SessionScope(db):
+            db.session.add(d_audit)
+            db.session.commit()
+        self.test_user, d_audit = map(db.session.merge, (self.test_user, d_audit))
+        self.test_user.deceased = d_audit
+        self.login()
+        data = {"resourceType": "Patient",
+                'deceasedBoolean': False}
+        rv = self.client.put(
+            '/api/demographics/%s' % TEST_USER_ID,
```
ivan-c: just curious, would it make sense to use `url_for()` in tests...?

pbugni: alters the setup - one could certainly make the case for it, but we're currently doing it this way.  /shrug

PR #2162, opened 2018-05-24
Feature/linting i18next config

achen2401:
linting i18next config JS file
also found that it wasn't consuming session storage data as it should (making ajax call every page load)
don't make ajax call to load translation json if the language is the default en-US
Ivan, let me know if you see any issue with my changes.

### DIFF
--- a/portal/static/js/i18next-config.js
+++ b/portal/static/js/i18next-config.js
@@ -21,110 +21,108 @@ var __i18next = window.__i18next = (function() {
                 window.localStorage.removeItem("i18nextLng");
             }
         }
-        options = options || {};
-        if (options.lng) {
-            options.lng = options.lng.replace("_", "-");
-        }
-        /*
-         * consuming translation json from each corresponding locale
-         */
-        var source = options.loadPath ? options.loadPath : "/static/files/locales/{{lng}}/translation.json";
-
-        i18next.use(i18nextXHRBackend) /* global i18next i18nextXHRBackend */
-            .init({
-                fallbackLng: options.fallbackLng ? options.fallbackLng.replace("_", "-") : "en-US",
-                lng: options.lng ? options.lng : "en-US",
-                preload: options.lng ? [options.lng] : false,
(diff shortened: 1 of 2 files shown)


portal/static/js/i18next-config.js
```diff
@@ -21,110 +21,114 @@ var __i18next = window.__i18next = (function() {
...
-                keySeparator: "----",
-                nsSeparator: "----",
-                load: "currentOnly", //this reads language code in en-US, en-AU format
-                returnEmptyString: false,
-                returnNull: false,
-                saveMissing: true,
-                missingKeyHandler: function(lng, ns, key, fallbackValue) {
-                    if (options.missingKeyHandler) {
-                        options.missingKeyHandler(lng, ns, key, fallbackValue);
+        var source = options.loadPath ? options.loadPath : "/static/files/locales/{{lng}}/translation.json"; //consuming translation json from each corresponding locale
```
ivan-c: Are you able to use `||` here? A little easier to read than a ternary, but not a big deal :)
eg
```
var source = options.loadPath || "/static/files/locales/{{lng}}/translation.json";
```

portal/static/js/i18next-config.js
```diff
@@ -21,110 +21,114 @@ var __i18next = window.__i18next = (function() {
...
+                    var sessionItemKey = "i18nextData_" + options.language;
+                    if (data && (typeof data === "undefined" ? "undefined" : _typeof(data)) === "object") { /*global _typeof */
+                        if (!cache) { data["_t"] = new Date();}
+                        // URL encoded form data must be in querystring format
+                        data = addQueryString("", data).slice(1); /* global addQueryString */
+                    }
+                    if (options.queryStringParams) {
+                        url = addQueryString(url, options.queryStringParams); /* global addQueryString */
+                    }
+                    try {
```
ivan-c: Are you able to move some of this code outside of the `try` block (unless all of it could throw an exception)?

PR #2163, opened 2018-05-24
put try-catch around loading portal wrapper call

achen2401:
Spinning wheel loading display is displayed before the call to portal wrapper. In the event that the call to get the portal wrapper failed, due to some errors, the call to hide the spinning wheel was never reached.  Made the fix to hide the spinning wheel on request error  and also adding try-catch block around the call to try to catch any runtime error.


ivan-c:
@achen2401 This looks like a really good solution for the portal wrapper (especially debugging inside an intervention).

I think we might still need something to cover a more fundamental error in their browser though, like a *very* old browser, or an error loading an external library. It would likely have to be entirely in native javascript, and loaded independently of the normal libraries, to prevent an error in the normal code from masking the error-reporting code. There won't be a perfect solution, just wanted to get some thoughts down :)

PR #2173, opened 2018-05-29
Feature/js error handling

achen2401:
general JS error handling:
add error handling for Vue instances
add check for presence of JQuery
add check for whether Javascript is enabled
display message if user is using older IE browser(s)
bug fix - fix missing social icons on GIL home page
**it will be good for this PR to be in this sprint but if it turns out that a lot of changes are required upon review, it can wait until next sprint then.**

### DIFF
--- a/portal/views/portal.py
+++ b/portal/views/portal.py
@@ -63,7 +63,7 @@ def favicon():
 
 @portal.route('/no-script')
 def no_script():
-    return  make_response(_(u"This application requires Javascript enabled.\
+    return make_response(_(u"This application requires Javascript enabled.\
                             Please check your browser setting."))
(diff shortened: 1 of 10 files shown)


portal/views/portal.py
```diff
@@ -61,6 +61,12 @@ def favicon():
     return redirect(url_for('static', filename='img/favicon.ico'), code=302)
 
 
+@portal.route('/no-script')
+def no_script():
+    return make_response(_(u"This application requires Javascript enabled.\
+                            Please check your browser setting."))
```
ivan-c: >Please check your browser setting.

should probably be setting*s*

PR #2175, opened 2018-05-30
TN-1187, TN-188 Created workaround to non-associative relativedelta month math

pbugni:
Resolve the inconsistencies of using a non discrete metric (month) in an associative manner.

The tests module now include `associative_backdate` to adjust the "now" value such that the tests will perform in a deterministic manner when run at end dates which don't yield associative results.

### DIFF
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -50,6 +50,35 @@ from portal.models.communication_request import CommunicationRequest
 CommunicationRequest
 
 
+def associative_backdate(now, backdate):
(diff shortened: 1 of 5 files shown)


tests/__init__.py
```diff
@@ -50,6 +50,35 @@
 CommunicationRequest
 
 
+def associative_backdate(now, backdate):
```
ivan-c: Would [commutative](https://en.wikipedia.org/wiki/Commutative_property) be more appropriate jargon? :)

PR #2185, opened 2018-05-30
Feature/development gulpfile code cleanup

achen2401:
code cleanup - use pipe callback instead of settimeout function for executing callback for gulp less tasks.
@ivan-c Ivan, this is good for this sprint but probably can wait until next sprint also.


portal/development_gulpfile.js
```diff
@@ -157,14 +157,12 @@ gulp.task("gilLess", () => {
         .pipe(postCSS())
         .pipe(rename(GIL + ".css"))
         .pipe(sourcemaps.write("../../../"+mapPath)) //path relative to the source file
-        .pipe(gulp.dest("gil/" + cssPath));
-
-    setTimeout(function() {
-        replaceStd(GIL + ".css.map");
-    }, 500);
+        .pipe(gulp.dest("gil/" + cssPath))
```
ivan-c: If you need to access something outside `rootPath` (ie `static`) you should probably change `rootPath` to a higher directory. 

Can you rewrite `rootPath` or `cssPath` as necessary so everything can be relative (subdirectory) to `rootPath`?

Decent example of using using `process.cwd()` to set something similar to `rootPath` [here](https://stackoverflow.com/questions/30594308/how-can-i-get-the-path-from-where-gulp-was-run/30595657#30595657)

achen2401: @ivan-c ivan, it is a bit complicated - dealing with multiple entry points here:  ./static/, ./gil/static, ./eproms/static,  to each of the destination that the gulp task generates the css file. I rewrote to have a central rootPath, "./" to but still need to diverge to each (EPROMsPath and GILPath) for the generated css file.

ivan-c:
Sorry for the nitpicking, I was hoping to make it easier to run `npm`/`gulp` from any directory, but it seems `gulp` will `cd` to the same directory when you specify `--gulpfile`, reducing the need to use absolute paths everywhere
>--gulpfile, -f Manually set path of gulpfile. Useful if you have multiple gulpfiles. This will set the CWD to the gulpfile directory as well.

PR #2186, opened 2018-05-30
TN-1195 tx trigger date older than consent

pbugni:
With a treatment date much older than the consent date, we hit an error due to a well hidden bug.
Now checking the trigger_date for each questionnaire bank as it's being processed, as they don't all use the same trigger_date logic (baseline vs. recurring, for example).

### DIFF
--- a/tests/test_assessment_status.py
+++ b/tests/test_assessment_status.py
@@ -21,6 +21,7 @@ from portal.models.recur import Recur
 from portal.models.research_protocol import ResearchProtocol
 from portal.models.role import ROLE
 from portal.models.user import get_user
+from portal.system_uri import ICHOM
 from tests import associative_backdate, TestCase, TEST_USER_ID
 
 now = datetime.utcnow()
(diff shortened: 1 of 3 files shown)


tests/test_assessment_status.py
```diff
@@ -655,6 +656,24 @@ def test_boundary_in_progress_expired(self):
         a_s = AssessmentStatus(user=self.test_user, as_of_date=nowish)
         self.assertEquals(a_s.overall_status, 'Partially Completed')
 
+    def test_all_expired_old_tx(self):
+        self.login()
+        # backdate outside of baseline window (which uses consent date)
+        backdate, nowish = associative_backdate(
+            now=now, backdate=relativedelta(months=4, hours=1))
```
ivan-c: Would you be able to use the current datetime instead of `now`?

pbugni: `now` is the effective current datetime, cached at the module level to avoid the indeterminate problems of test run time.

ivan-c: That would make individual tests run consistently relative to eachother, but I don't think it would cover test runs on different days, like the situation that allowed us to discover the `dateutil.relativedelta` forward/back issue

PR #2191, opened 2018-06-01
TN-1198 copyright translation fix

achen2401:
address TN-1198 https://jira.movember.com/browse/TN-1198
fix untranslated copyright text before authentication

### DIFF
--- a/portal/static/js/main.js
+++ b/portal/static/js/main.js
@@ -1897,44 +1897,43 @@ var tnthDates = {
         var locale = "";
         if (sessionLocale) {
             return sessionLocale;
-        } else {
-            if (!checkJQuery()) { /*global checkJQuery */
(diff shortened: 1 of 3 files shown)


portal/static/js/main.js
```diff
@@ -1891,17 +1891,21 @@ var tnthDates = {
...
     },
-    "getUserLocale": function(force) {
+    "getUserLocale": function() {
         var sessionKey = this.localeSessionKey;
         var sessionLocale = sessionStorage.getItem(sessionKey);
         var locale = "";
-        if (!force && sessionLocale) {
+        if (sessionLocale) {
             return sessionLocale;
         } else {
```
ivan-c: I don't *think* you need that else anymore since you're `return`ing

PR #2197, opened 2018-06-04
TN-1162 pre-flight check for ExclusionPersistence

pbugni:


pbugni:
@ivan-c I just rebased due to conflicts, which was a significant pain due to all the 2to3 changes.  It'd be nice to get this merged soon so we can avoid doing that again...

PR #2199, opened 2018-06-04
TN-1176 role display name

pbugni:


pbugni:
@achen2401 could you modify the front end to use the `display_name` field you'll now find in the roles APIs in place of the front end code doing the same (Title Case and white space for underscores trick)?

PR #2204, opened 2018-06-05
TN-1222 Initial 2to3 run

ivan-c:
* Run `2to3` as follows below
  * Do not use fixes that add unnecessary `list()` wrapping
  * Do not run `print` fixes, some already exist as functions
* Todo: fix flask-dogpilecache

```
2to3 \ 
    --fix all \ 
    --nofix dict \ 
    --nofix print \ 
    --nofix filter \ 
    --nofix map \ 
    --nofix xrange \ 
    --nofix zip \ 
    --write \ 
    --nobackups \ 
    --processes=2 .
```

### DIFF
--- a/docker/remap_envvars.py
+++ b/docker/remap_envvars.py
@@ -2,7 +2,7 @@
 # Parses and remaps DB URLs into standard psql environment variables
 # https://www.postgresql.org/docs/9.6/static/libpq-envars.html
 
-from urlparse import urlparse
+from urllib.parse import urlparse
 from os import environ
 
 
--- a/portal/audit.py
+++ b/portal/audit.py
@@ -31,7 +31,7 @@ def auditable_event(message, user_id, subject_id, context="other"):
     subject_id: The user id upon which the action was performed
 
     """
-    text = u"performed by {0} on {1}: {2}: {3}".format(
+    text = "performed by {0} on {1}: {2}: {3}".format(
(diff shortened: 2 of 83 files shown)


docker/remap_envvars.py
```diff
@@ -2,7 +2,7 @@
 # Parses and remaps DB URLs into standard psql environment variables
 # https://www.postgresql.org/docs/9.6/static/libpq-envars.html
 
-from urlparse import urlparse
+from urllib.parse import urlparse
```
pbugni: this will break on python2 - do we need a shim?

portal/audit.py
```diff
@@ -31,7 +31,7 @@ def auditable_event(message, user_id, subject_id, context="other"):
     subject_id: The user id upon which the action was performed
 
     """
-    text = u"performed by {0} on {1}: {2}: {3}".format(
+    text = "performed by {0} on {1}: {2}: {3}".format(
```
pbugni: Won't this potentially create coercion errors if say the message contains unicode while we're still on py2?

portal/models/communication.py
```diff
@@ -20,6 +20,7 @@
 from .questionnaire_bank import QuestionnaireBank
 from ..trace import dump_trace, establish_trace, trace
 from .user import User
+import collections
```
pbugni: should be above local imports

PR #2206, opened 2018-06-05
TN-1163 Added new tests for dict_tools.py

John0Randolph:
- added tests for dict_tools.py

### DIFF
--- a/portal/dict_tools.py
+++ b/portal/dict_tools.py
@@ -30,7 +30,7 @@ def dict_match(newd, oldd, diff_stream):
 
     """
     added, removed, modified, same = dict_compare(newd, oldd)
-    if len(same) == len(list(newd.keys())) and len(list(newd.keys())) == len(list(oldd.keys())):
+    if len(same) == len(newd.keys()) and len(newd.keys()) == len(oldd.keys()):
         assert not added
         assert not removed
         assert not modified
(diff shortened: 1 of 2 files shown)


portal/dict_tools.py
```diff
@@ -30,7 +30,7 @@ def dict_match(newd, oldd, diff_stream):
 
     """
     added, removed, modified, same = dict_compare(newd, oldd)
-    if len(same) == len(newd.keys()) and len(newd.keys()) == len(oldd.keys()):
+    if len(same) == len(list(newd.keys())) and len(list(newd.keys())) == len(list(oldd.keys())):
```
pbugni: I imagine the list(*.keys()) came from http://python-future.org/compatible_idioms.html#dict-keys-values-items-as-a-list, but I don't believe that's necessary to check the `len`gth?  (That is, we don't need to force it to be of type list, just comparing the length of the two here)

portal/dict_tools.py
```diff
@@ -43,7 +43,7 @@ def dict_match(newd, oldd, diff_stream):
             diff_stream.write(
                 "removed {}\n".format({k:oldd.get(k) for k in removed}))
         if modified:
-            for k in modified.keys():
+            for k in list(modified.keys()):
                 diff_stream.write(
```
pbugni: same as above, we just need to iterate over the keys here - no need to force to be a `list` type

portal/dict_tools.py
John0Randolph: Looks like you're right. I added the list(*.keys()) with 2to3 but I see now that they're not necessary in this case.

PR #2209, opened 2018-06-06
System maintenace UI changes

achen2401:
@mcjustin @pbugni @ivan-c 
I added the UI **placeholder element** for system maintenance message in the landing pages for EPROMs and TruenTH USA - and stylings for each.  As I was working on it, I became unsure as to what config variable (e.g. message, time windows) would be best for this.  You are free to fill in the message/conditional.


achen2401:
should be against master, will re-open.

PR #2214, opened 2018-06-06
Fixup/sort imports

ivan-c:
* Add [isort](https://github.com/timothycrosley/isort) config
* Sort imports for consistency

### DIFF
--- a/.isort.cfg
+++ b/.isort.cfg
@@ -1,0 +1,5 @@
+# isort configuration file
+# https://github.com/timothycrosley/isort
+[settings]
+line_length=120
+multi_line_output=3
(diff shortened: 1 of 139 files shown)


.isort.cfg
```diff
@@ -0,0 +1,5 @@
+# isort configuration file
+# https://github.com/timothycrosley/isort
+[settings]
+line_length=120
```
pbugni: I'd rather stick w/ 79 as per https://www.python.org/dev/peps/pep-0008/#maximum-line-length

ivan-c: Sorry, we had been using [100 characters for pep8speaks](https://github.com/uwcirg/true_nth_usa_portal/blob/develop/.pep8speaks.yml#L5-L6).
I'll match what we're using there.

PR #2217, opened 2018-06-06
Fixup/2to3 print

ivan-c:
* Run 2to3 [`print` ruleset](https://docs.python.org/2/library/2to3.html#2to3fixer-print)

### DIFF
--- a/manage.py
+++ b/manage.py
@@ -289,11 +289,11 @@ def config(config_key):
 
     if config_key:
         # Remap None values to an empty string
-        print(app.config.get(config_key, '') or '')
+        print((app.config.get(config_key, '') or ''))
         return
 
-    print(json.dumps(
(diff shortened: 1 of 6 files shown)


manage.py
```diff
@@ -289,11 +289,11 @@ def config(config_key):
 
     if config_key:
         # Remap None values to an empty string
-        print(app.config.get(config_key, '') or '')
+        print((app.config.get(config_key, '') or ''))
```
pbugni: isn't that a redundant `or ''`, as you provide that as a default to `get()`?

ivan-c: the default to `get()` will work if it's not in `app.config`, but there are some config values that are set to `None`. The `or ''` is to catch config values set as `None` and return `''` instead.

PR #2227, opened 2018-06-08
Feature/psa tracker ui enhancements

achen2401:
address: https://jira.movember.com/browse/TN-932
PSA Tracker UI enhancements:
- ran linting on psaTracker code
- add text label and tooltip to data points on graph
- add filtering options 
- fix Y results axis to use log scale
- add treatment information (and to graph)
- bug fix for date validation


portal/development_gulpfile.js
```diff
@@ -182,6 +183,20 @@ gulp.task("topnavLess", function() {
...
     return true;
 });
+
+gulp.task("psaTrackerLess", function() {
+    gulp.src(lessPath + "/" + PSATRACKER + ".less")
+        .pipe(sourcemaps.init())
+        .pipe(less({
+            plugins: [cleancss]
+        }))
+        .pipe(sourcemaps.write("../maps"))
```
ivan-c: Can you use one of the path variables instead of a relative path?

portal/static/js/main.js
```diff
@@ -1722,7 +1722,7 @@ var tnthDates = {
         if (!date || date.length < 10) { return false; }
         var dArray = $.trim(date).split(" ");
         if (dArray.length < 3) { return false; }
-        var day = dArray[0], month = dArray[1], year = dArray[2];
+        var day = parseInt(dArray[0])+"", month = dArray[1], year = dArray[2];
```
ivan-c: Just curious, why do you need to cast to an int, then back to a string?

portal/static/js/psaTracker.js
```diff
@@ -1,423 +1,726 @@
...
-                    var self = this;
-                    var d3 = self.d3;
-                    var i18next = self.i18next;
-                    const WIDTH = 550, HEIGHT = 430, TOP = 50, RIGHT = 20, BOTTOM = 110, LEFT = 60, TIME_FORMAT = "%d %b %Y";
-
-                    // Set the dimensions of the canvas / graph
-                    var margin = {top: TOP, right: RIGHT, bottom: BOTTOM, left: LEFT},
+                }
+            },
+            drawGraph: function() { //using d3 library to draw graph
```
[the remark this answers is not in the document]
ivan-c: Would you be able to break this function into smaller pieces?
I'm not sure the best way, but if some of the helper functions (eg `logTickFormat()`) are reusable, maybe they could be moved outside `drawGraph()`?

portal/static/js/psaTracker.js
```diff
@@ -1,423 +1,726 @@
...
-                    date:"",
-                    edit: false
-                },
-                headers: [
-                    this.i18next.t("PSA (ng/ml)"),
-                    this.i18next.t("Date")
-                ],
+    var psaApp = window.psaApp = new Vue({ /*global Vue*/
+        el: "#mainPsaApp",
+        components: {
```
ivan-c: Would anything make sense as a separate component (eg adding a PSA result)?

PR #2235, opened 2018-06-11
Resolve error introduced in 2to3 urllib.parse.urlunsplit

pbugni:
In 2to3 we picked up a buggy version of urlunsplit, that compares the type of the first arg with each successive arg in urlunsplit.  Work around this bug by using empty strings (same type as args returned from urlsplit) rather than None for clearing scheme and hostname from a URL.

### DIFF
--- a/portal/views/patch_flask_user.py
+++ b/portal/views/patch_flask_user.py
@@ -12,7 +12,7 @@ from flask_user.views import _endpoint_url
 def patch_make_safe_url(url):
     """Patch flask_user.make_safe_url() to include '?'
 
-    Turns an usafe absolute URL into a safe relative URL by removing
+    Turns an unsafe absolute URL into a safe relative URL by removing
     the scheme and the hostname
     Example:
         make_safe_url('http://hostname/path1/path2?q1=v1&q2=v2#fragment')
@@ -20,8 +20,9 @@ def patch_make_safe_url(url):
 
     """
     parts = urlsplit(url)
+    no_scheme, no_hostname = u'', u''
     safe_url = urlunsplit(
-        (None, None, parts.path, parts.query, parts.fragment))
+        (no_scheme, no_hostname, parts.path, parts.query, parts.fragment))
     return safe_url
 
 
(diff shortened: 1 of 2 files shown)


portal/views/patch_flask_user.py
```diff
@@ -12,16 +12,17 @@
...
-    Turns an usafe absolute URL into a safe relative URL by removing
+    Turns an unsafe absolute URL into a safe relative URL by removing
     the scheme and the hostname
     Example:
         make_safe_url('http://hostname/path1/path2?q1=v1&q2=v2#fragment')
         returns: '/path1/path2?q1=v1&q2=v2#fragment
 
     """
     parts = urlsplit(url)
+    no_scheme, no_hostname = u'', u''
```
ivan-c: I think you'll have to remove the `u`, the attributes of `parts` are all `str`, not `unicode`

```
>>> url='http://hostname/path1/path2?q1=v1&q2=v2#fragment'
>>> parts = urlsplit(url)
>>> no_scheme, no_hostname = u'', u''
>>> safe_url = urlunsplit((
...     no_scheme,
...     no_hostname,
...     parts.path,
...     parts.query,
...     parts.fragment,
... ))
  File "<stdin>", line 6, in <module>
  File "/home/ivanc/src/portal/shared_services/env/local/lib/python2.7/site-packages/future/backports/urllib/parse.py", line 399, in urlunsplit
    _coerce_args(*components))
  File "/home/ivanc/src/portal/shared_services/env/local/lib/python2.7/site-packages/future/backports/urllib/parse.py", line 115, in _coerce_args
    raise TypeError("Cannot mix str and non-str arguments")
TypeError: Cannot mix str and non-str arguments
>>> type(parts.path)
<type 'str'>
>>> type(no_scheme)
<type 'unicode'>
>>>
```

tests/test_patch_flask_user.py
```diff
@@ -6,32 +6,32 @@
 class TestPathFlaskUser(TestCase):
```
ivan-c: Does this need to inherit from `TestCase`?

pbugni: Great point - too bad I didn't think about that or see this prior to merge.  This is probably a good next step for @John0Randolph as this test class can be easily ported to pytest.

PR #2240, opened 2018-06-12
TN-936 psa-tracker strategy

pbugni:


pbugni:
@mcjustin not just a typo in `test_diag_changed_strategy` - found during manual testing, that the strategy function `observation_check` was not correctly handling a history of observations.  Added this test to confirm the fix there sticks.

PR #2242, opened 2018-06-12
TN-1207 MedidataRave integration UI changes

achen2401:
https://jira.movember.com/browse/TN-1207
UI changes for MedidataRave integration - disable MedidataRave fields based on config variables (i.e. MEDIDATA_RAVE_FIELDS and MEDIDATA_RAVE_ORG)  **note**: fields won't be disabled yet until the config variables are set


portal/static/js/profile.js
```diff
@@ -228,17 +228,31 @@
                     return {"done": true};
                 });
             },
-            setTopLevelOrgs: function(data) {
-                if (data && data.careProvider) {
-                    var orgTool = this.getOrgTool();
+            setTopLevelOrgs: function(callback) {
+                callback = callback || {};
```
ivan-c: Should that default be an anonymous function (`function()`) instead of an object (`{}`)?

portal/static/js/profile.js
```diff
@@ -433,6 +437,15 @@
             isAdmin: function() {
                 return this.currentUserRoles.indexOf("admin") !== -1;
             },
+            isPatient: function() {
+                if (this.mode === "createPatientAccount" || $("#profileMainContent").hasClass("patient-view")) {
+                    return true;
+                } 
+                if (this.userRoles.length === 0) { //this is a blocking call if not cached, so avoid it if possible
```
ivan-c: Shouldn't the role data be cached in localStorage by a service worker at this point?

achen2401: @ivan-c ivan, you are right, it will be if the call to get role data is made, but it is not always made (e.g. as in account creation page)

ivan-c: Ah, good to know.

PR #2243, opened 2018-06-12
TN-1163 date_tools.py, access_on_verify, app_text py2 to py3

John0Randolph:
Some py2 to py3 updates
Updated README with PosgreSQL Windows installation guide
more tests in test_date_tools.py


README.rst
```diff
@@ -397,3 +397,85 @@ the ``docs/build/html/index.html`` file
...
+.. code:: bash
+
+    $ pip install -r requirements.txt
+
+Configuration files
+^^^^^^^^^^^^^^^^^^^
+
+In $PATH\\data\pg_hba.conf, change the bottom few lines to read:
+
+# TYPE  DATABASE        USER            ADDRESS                 METHOD
```
ivan-c: Can you put the config file contents in a [literal block](http://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#literal-blocks)?

README.rst
```diff
@@ -397,3 +397,85 @@ the ``docs/build/html/index.html`` file
...
+Install requirements by running:
+
+.. code:: bash
+
+    $ pip install -r requirements.txt
+
+Configuration files
+^^^^^^^^^^^^^^^^^^^
+
+In $PATH\\data\pg_hba.conf, change the bottom few lines to read:
```
ivan-c: Can you wrap the file path in literal marks? eg:
```
In ``$PATH\\data\pg_hba.conf``, change ...
```

README.rst
```diff
@@ -397,3 +397,85 @@ the ``docs/build/html/index.html`` file
...
+
+host    all             all             ::1/128                 trust
+
+Copy the default configuration file to the named configuration file
+
+.. code:: bash
+
+    $ copy $PROJECT_HOME/instance/application.cfg.default $PROJECT_HOME/instance/application.cfg
+
+In application.cfg, (below), fill in the values on line 9 for user, password, localhost, portnum, and dbname
```
ivan-c: Can you remove the reference to line number and refer to the config name (ie `SQLALCHEMY_DATABASE_URI`) instead?

portal/date_tools.py
```diff
@@ -137,5 +137,5 @@ def localize_datetime(dt, user):
         best = local.astimezone(tz)
     else:
         best = dt
-    d, m, y = best.strftime('%-d %b %Y').split()
+    d, m, y = best.strftime('%d %b %Y').split()
```
ivan-c: Was this intentional? I think this will remove any leading `0` on the day of the month:
```
>>> f
datetime.datetime(2018, 6, 9, 16, 28, 1, 964588)
>>> f.strftime('%d %b %Y')
'09 Jun 2018'
>>> f.strftime('%-d %b %Y')
'9 Jun 2018'
```

portal/models/fhir.py
```diff
@@ -25,6 +25,7 @@
 from .organization import OrgTree
 from .performer import Performer
 from .reference import Reference
+from past.builtins import basestring
```
ivan-c: Can you move this import to the 1st block of imports, after the `flask` import line, to keep it alphabetical?


Imports are typically separated into 3 sorted blocks:
1. Built-in modules (eg `os`)
2. 3rd party modules (typically installed from PyPI)
3. Submodules in the same project

We're trying a tool out ([isort](https://github.com/timothycrosley/isort)) if you want to give it a try too

tests/test_date_tools.py
```diff
@@ -28,3 +27,11 @@ def test_int_date(self):
         with self.assertRaises(BadRequest) as e:
             dt = FHIR_datetime.parse(acceptance_date, 'acceptance date')
         self.assertTrue('acceptance date' in str(e.exception))
+
+    def test_localize_datetime_none(self):
+        self.assertEqual(localize_datetime(None, None), '')
+
+    def test_localize_datetime_no_user(self):
+        input_date = datetime.strptime('Jun 11 2012', '%b %d %Y')
```
ivan-c: Can you use a day of the month that has a leading `0` for your test case?

portal/models/fhir.py
```diff
@@ -660,7 +661,7 @@ def fetch_HL7_V3_Namespace(valueSet):
 def fetch_local_valueset(valueSet):
     """Pull and parse the named valueSet from our local definition"""
     response = valueset_nhhd_291036()
-    data = json.loads(response.data)
+    data = json.loads(response.data.decode("utf-8"))
```
ivan-c: @pbugni Is there a reason we can't use `response.json` to get the JSON directly?

ivan-c:
@John0Randolph 
You'll want to add the `test_date_tools` module to [tox.ini](https://github.com/uwcirg/true_nth_usa_portal/blob/develop/tox.ini#L18) so that we'll run your newly ported tests in python3 as well. eg:
```
    py34: nosetests -v tests.test_dict_tools tests.test_date_tools []

```

portal/models/fhir.py
pbugni: excellent point, @ivan-c 

@John0Randolph you can kill off the `data` variable, and use `response.json()['codeSystem']['concept']` in the call to `parse_concepts`  - note a similar patch should be applied to `fetch_HL7_V3_Namespace()` just above.

John0Randolph: This fix works in fetch_HL7_V3_Namespace(), but when I put it in fetch_local_valueset(valueSet), it throws an error (although I might be implementing it wrong). Code:
    response = valueset_nhhd_291036()
    #data = json.loads(response.data.decode("utf-8"))
    return parse_concepts(response.json()['codeSystem']['concept'],
                          system='{}/{}'.format(TRUENTH_VALUESET, valueSet))

Error:
TypeError: 'dict' object is not callable

portal/date_tools.py
John0Randolph: Running it with "%-d', I'm getting a ValueError, but with '%d' it's not removing any leading 0.

ivan-c: It looks like this is [a quirk of windows](https://stackoverflow.com/questions/28894172/why-does-d-or-e-remove-the-leading-space-or-zero/28916983#28916983). Please revert 
>Python datetime.strftime() delegates to C strftime() function that is platform-dependent:

[See also](https://github.com/facebook/prophet/issues/99)

[Possible solution here](https://stackoverflow.com/questions/904928/python-strftime-date-without-leading-0/16097385#16097385)

portal/models/fhir.py
ivan-c: json is a property, not a method :)

`return parse_concepts(response.json`~~()~~`['codeSystem']['concept']`

tests/test_date_tools.py
```diff
@@ -28,3 +27,11 @@ def test_int_date(self):
...
             dt = FHIR_datetime.parse(acceptance_date, 'acceptance date')
         self.assertTrue('acceptance date' in str(e.exception))
+
+    def test_localize_datetime_none(self):
+        self.assertEqual(localize_datetime(None, None), '')
+
+    def test_localize_datetime_no_user(self):
+        input_date = datetime.strptime('Jun 01 2012', '%b %d %Y')
+        expected = '1 Jun 2012'
+        self.assertEqual(localize_datetime(input_date, None), expected)
```
ivan-c: More keyword arguments :)

portal/models/identifier.py
```diff
@@ -27,7 +27,7 @@ def value(self):
     @value.setter
     def value(self, value):
         # Force to text
-        self._value = unicode(value)
+        self._value = str(value)
```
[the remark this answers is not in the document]
John0Randolph: So removing str() from this file still allows the tests to pass, but removing the str() calls in app_text causes the tests to throw errors in both py2.7 and py3.4. I'm not exactly sure why but there must be a good reason that in app_text some variables were being recast as unicode and I believe str() is the equivalent of that call for both py2.7 and py3.4 compatability.

tests/test_app_text.py
```diff
@@ -114,7 +113,9 @@ def test_asset_variable_replacement(self):
                                        asset=invalid_asset,
                                        variables=test_vars)
         error_key = resource.asset.split()[-1]
-        self.assertEqual(error_key, "u'variable'")
+        if (error_key.startswith("u")):
```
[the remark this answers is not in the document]
John0Randolph: I believe I changed that because of the way unicode strings are handled in py2 vs py3. In py2, the result of this test is as "u'variable'" whereas in py3, it is "'variable'" (but they are both unicode strings for 'variable'), so I added that line to account for both. It's probably not the best way to do that, though - is there a cleaner way?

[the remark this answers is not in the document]
ivan-c: [This looks like a better way](https://stackoverflow.com/questions/9079036/detect-python-version-at-runtime/9079062#9079062), thanks for the explanation!

PR #2244, opened 2018-06-12
TN-1236 Port use of `type()` to python enum.Enum for ROLE and RELATIONSHIP

pbugni:
Not only a more modern and better solution, this resolved problems in porting to python3, the use of native strings in `type()`.


portal/static/js/vendor.js
```diff
@@ -2954,8 +2954,8 @@ var Dropdown = (function ($) {
     BACKDROP: '.dropdown-backdrop',
     DATA_TOGGLE: '[data-toggle="dropdown"]',
     FORM_CHILD: '.dropdown form',
-    ROLE_MENU: '[role="menu"]',
```
ivan-c: `sed` false positives in this file?

PR #2250, opened 2018-06-13
Feature/orgs profile display fix

achen2401:
https://jira.movember.com/browse/TN-1241

fix display of organizations in profile - to exclude text (e.g. practitioner) other than orgs

### DIFF
--- a/portal/static/js/profile.js
+++ b/portal/static/js/profile.js
@@ -1640,7 +1640,8 @@
                 if (!this.demo.data.careProvider || this.demo.data.careProvider.length === 0) {
                     return this.getNoOrgDisplay();
                 }
-                /* example return from api demographics: { display: Duke, reference: "api/organization/1301"} */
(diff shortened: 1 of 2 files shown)


portal/static/js/profile.js
```diff
@@ -1633,6 +1633,25 @@
...
+            },
+            getOrgsDisplay: function() {
+                if (!this.demo.data.careProvider || this.demo.data.careProvider.length === 0) {
+                    return this.getNoOrgDisplay();
+                }
+                /* example return from api demographics: [{ display: Duke, reference: "api/organization/1301"}, {"display":"Arvin George","reference":"api/practitioner/1851648521?system=http://hl7.org/fhir/sid/us-npi"}]
+                 * NOTE: need to exclude displays other than organization */
+                var self = this;
+                var arrDisplay = this.demo.data.careProvider.map(function(item) {
+                    if (item.reference.match(/^api\/organization\/0$/gi)) { //organization id = 0
```
ivan-c: If you're doing an exact match for that string (`api/organization/0`) do you still need a regular expression?

PR #2251, opened 2018-06-13
TN-1243 Adding CommunicationRequest records for IRONMAN 9 mo assessment

mcjustin:
@pbugni  before you merge... noting that we should really go ahead and add all these at this point:
9 | 12 | 15 | 18 | 21 | 24 | 30 | 36
... per https://docs.google.com/spreadsheets/d/1oJ8HKfMHOdXkSshjRlr8lFXxT4aUHX5ntxnKMgf50wE/edit#gid=1339608238
And, that as it stands this will make this file quite verbose...

Are you adding QB's for the above at this point?

### DIFF
--- a/portal/config/eproms/CommunicationRequest.json
+++ b/portal/config/eproms/CommunicationRequest.json
@@ -543,6 +543,176 @@
         {
           "system": "http://us.truenth.org/identity-codes/communicationrequest/name",
           "use": "usual",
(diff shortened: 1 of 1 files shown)


pbugni:
FYI, delaying merging this until we've resolved TN-1237, as people will get reminders on which they can't act.

PR #2252, opened 2018-06-13
ALL_BUT_WRITE_ONLY should simply be a list of role strings.

pbugni:
Correct error from PR #2244 

@ivan-c this was noticed on stg by @achen2401 - we're currently locked out of any view decorated with `@roles_required(ALL_BUT_WRITE_ONLY)`

### DIFF
--- a/portal/models/role.py
+++ b/portal/models/role.py
@@ -86,10 +86,7 @@ STATIC_ROLES = {
 
 
 ROLE = Enum('ROLE', {r.upper(): r for r in STATIC_ROLES})
-ALL_BUT_WRITE_ONLY = Enum(
-    'ALL_BUT_WRITE_ONLY',
-    {r.upper(): r for r in STATIC_ROLES if r != 'write_only'})
-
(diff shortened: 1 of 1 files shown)


portal/models/role.py
```diff
@@ -86,10 +86,7 @@ def display_name(self):
 
 
 ROLE = Enum('ROLE', {r.upper(): r for r in STATIC_ROLES})
-ALL_BUT_WRITE_ONLY = Enum(
-    'ALL_BUT_WRITE_ONLY',
-    {r.upper(): r for r in STATIC_ROLES if r != 'write_only'})
-
+ALL_BUT_WRITE_ONLY = [r.value for r in ROLE if r.value != 'write_only']
```
ivan-c: Why not an immutable type eg set/tuple?

pbugni: Unfortunate design of the flask-user `@roles.required` API into which this variable is passed.  It interprets a `tuple` as a list for which the user must have *all* of the roles, and a `list` as a list of options for which the user must have at least one.  We need the later, so it must be a list.

PR #2259, opened 2018-06-14
Limit celery-worker CPU usage

ivan-c:
* Limit celery-worker CPU usage to .3 CPUs
  * Prevent CPU spike and system alarm after deployment


ivan-c:
>correct the puppet roles to name celeryd -> celeryworker(d)

@pbugni that threw me off too today. would be a good opportunity to switch to [systemd unit files ](https://github.com/celery/celery/issues/4304#issuecomment-353168370)for those going forward

PR #2265, opened 2018-06-15
TN-1163 Made fewer tests in test_date_tools.py depend on TestCase

John0Randolph:


tests/test_date_tools.py
```diff
@@ -5,34 +5,38 @@
 
 from portal.date_tools import FHIR_datetime, RelativeDelta, localize_datetime
 from tests import TestCase
+import pytest
```
ivan-c: Can you put this 3rd party import in its own block, above the `portal` import line?

PR #2267, opened 2018-06-15
in patients list - allow user with admin role to create account

achen2401:
- allow user with admin role in patients list to have access to create account button

### DIFF
--- a/portal/static/js/admin.js
+++ b/portal/static/js/admin.js
@@ -285,7 +285,7 @@
             },
             setUserRoles: function(callback) {
                 callback = callback || function() {};
-                var self = this;
+                var self = this, tnthAjax = this.getDependency("tnthAjax");
                 tnthAjax.getRoles(this.userId, function(data) {
                     if (!data || data.error) {
                         callback({"error": i18next.t("Error occurred setting user roles")});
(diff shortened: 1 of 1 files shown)


portal/static/js/admin.js
```diff
@@ -45,7 +45,9 @@
             configured: false,
             initIntervalId: 0,
             sortFilterEnabled: false,
+            isAdmin: false,
             userId: null,
+            userRoles: [],
```
ivan-c: Do you still need to save the roles here if they're already cached in localStorage?

portal/static/js/admin.js
```diff
@@ -261,7 +263,47 @@
...
+                if (this.userRoles.length > 0) {
+                    callback(this.userRoles);
+                    return;
+                }
+                this.setUserRoles(callback);
+            },
+            setUserRoles: function(callback) {
+                callback = callback || function() {};
+                var self = this, tnthAjax = this.getDependency("tnthAjax");
+                tnthAjax.getRoles(this.userId, function(data) {
```
ivan-c: Naive question, but couldn't you just check the returned list from `tnthAjax.getRoles()` instead of assigning them to `userRoles` and checking that?

portal/static/js/admin.js
achen2401: @ivan-c Ivan, yea, it will be, it is just that having a property here allows other methods in the Admin Vue instance to have access to it.

ivan-c:
@achen2401 Thanks for the explanations!

PR #2272, opened 2018-06-17
Feature/psa tracker date validation fix

achen2401:
https://jira.movember.com/browse/TN-932
follow up fixes:
date validation
x and y axis grid lines
styling fixes


portal/static/js/psaTracker.js
```diff
@@ -156,13 +156,16 @@
...
                 if(!isValid) {
-                    this.addErrorMessage = this.i18next.t("Result must be a number.");
+                    this.addErrorMessage = this.i18next.t("Result must be a number and within valid range (less than 9999).");
                 } else {
                     this.addErrorMessage = "";
                 }
                 return isValid;
             },
             validateDate: function(date) {
+                if (!date || date.length <= 9) { //don't show validation error until enough characters are entered, just return false
```
ivan-c: Do you still need? Doesn't the validation error only appears after the field loses focus?

achen2401: @ivan-c ivan yup, you are right, it does.  I am putting a check here to actually not show validation error until the user has entered sufficient characters.  i was seeing it even when i didn't enter any character.

PR #2277, opened 2018-06-18
psa tracker follow y axis label rendering fix

achen2401:
followup fixes to my last PR:

found that y axis labels were not displayed correctly, made fix for that, also linting/simply code


portal/static/js/psaTracker.js
```diff
@@ -472,7 +473,7 @@
             },
             getRange: function getRange(size, startAt, step) {
                 var arr = []; size=size||10; startAt=startAt||0; step = step||1;
-                for (var index=startAt; index < size; index++) {
+                for (var index=parseInt(startAt); index < size; index++) {
```
ivan-c: Do you still need to cast from string to int here? It looks like it should already be an int (`-1`) below...

PR #2280, opened 2018-06-18
TN-1163 updated test_assessment_engine.py for py3 compatability

John0Randolph:


tests/test_assessment_engine.py
```diff
@@ -227,5 +228,10 @@ def test_assessments_csv(self):
             '/api/patient/assessment?format=csv&instrument_id={}'.format(instrument_id),
         )
         csv_string = download_response.data
```
ivan-c: Try the [`get_data()` method](http://werkzeug.pocoo.org/docs/0.14/wrappers/#werkzeug.wrappers.BaseResponse.get_data) to get a unicode string, instead of a bytestring, then you shouldn't need the version-conditional code.
```
csv_string  = download_response.get_data(as_text=True)
```

portal/models/identifier.py
```diff
@@ -48,6 +48,9 @@ def from_fhir(cls, data):
     def __str__(self):
         return 'Identifier {0.use} {0.system} {0.value}'.format(self)
 
+    def __hash__(self):
```
ivan-c: Need to read more about this change before I can comment ([this seems like a good reference](https://hynek.me/articles/hashes-and-equality/))

ivan-c: @pbugni Thoughts?

tests/test_date_tools.py
```diff
@@ -4,6 +4,8 @@
 import pytest
 from werkzeug.exceptions import BadRequest
 
+import pytest
```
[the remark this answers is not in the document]
John0Randolph: Where is this already imported? When I remove it, I get an error on line 32 `with pytest.raises(ValueError)`

tests/__init__.py
```diff
@@ -328,7 +328,7 @@ def shallow_org_tree(self):
         org_102 = Organization(id=102, name='102')
         org_1001 = Organization(id=1001, name='1001', partOf_id=101)
         with SessionScope(db):
-            map(db.session.add, (org_101, org_102, org_1001))
+            list(map(db.session.add, [org_101, org_102, org_1001]))
```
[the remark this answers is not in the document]
John0Randolph: Yeah, when I revert back to the previous line I get the following error:

`Traceback (most recent call last):
  File "C:\Users\JohnR17\Desktop\true_nth_usa_portal\tests\test_assessment_engine.py", line 194, in test_assessments_bundle
    self.bless_with_basics()
  File "C:\Users\JohnR17\Desktop\true_nth_usa_portal\tests\__init__.py", line 290, in bless_with_basics
    assert org
nose.proxy.AssertionError:`

tests/test_date_tools.py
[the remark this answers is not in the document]
ivan-c: Github is telling me it's already [on line 4 of `test_date_tools.py`](https://github.com/uwcirg/true_nth_usa_portal/pull/2280/files#diff-fb8ef56b50ef6f1b0429da28c462c10bR4).

The [Files Changed tab](https://github.com/uwcirg/true_nth_usa_portal/pull/2280/files) in the PR tells me the overall changes from your PR, maybe you're viewing an earlier commit?

tests/__init__.py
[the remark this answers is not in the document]
ivan-c: I think you missed a line, the failed `assert` should be below that line from what I can tell...

portal/models/identifier.py
ivan-c: Did some reading and we should only need that method defined if we use `Identifier` objects in dicts and sets (we likely do).

It's unlikely the `system` and `value` will ever change, but I think you might want to prefer the primary key (a suggestion in the above article), if available, for a attribute that's less likely to change. eg:

```
return hash(self.id) if self.id else hash(self.system)*hash(self.value)
```

ivan-c:
Oh yeah, don't forget to add `test_assessment_engine.py` to [`tox.ini`](https://github.com/uwcirg/true_nth_usa_portal/blob/develop/tox.ini#L18)!

Just a heads up, I might switch the way we call `py.test` in `tox.ini` so the files that are tested are [conditionally skipped](https://docs.pytest.org/en/latest/skipping.html#id1), instead of explicitly included

portal/models/identifier.py
John0Randolph: Yes, I was getting an error because Identifiers are used in sets somewhere. Ok, I'll change to that.

tests/test_assessment_engine.py
```diff
@@ -226,6 +227,6 @@ def test_assessments_csv(self):
         download_response = self.client.get(
             '/api/patient/assessment?format=csv&instrument_id={}'.format(instrument_id),
         )
-        csv_string = download_response.data
-        self.assertGreater(csv_string.split("\n"), 1)
+        csv_string = download_response.get_data(as_text=True)
+        self.assertGreater(len(csv_string.split("\n")), 1)
```
ivan-c: Nice, glad it worked! Can you change this to a py.test-style assert? eg
```
assert len(csv_string.split("\n")) > 1
```

tests/__init__.py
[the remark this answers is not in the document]
pbugni: Given the new behavior of map - in that it returns a lazy-evaluation and we just want it done, lets skip the call to map and simply use the list comprehension forced by the list() cast:
`[db.session_add(org) for org in (org_101, org_102, org_103)]`

portal/models/identifier.py
pbugni: I like this suggested approach - let's add a `__key__` and use it in both equality and hash:  https://stackoverflow.com/a/2909119

portal/models/identifier.py
```diff
@@ -48,9 +48,15 @@ def from_fhir(cls, data):
     def __str__(self):
         return 'Identifier {0.use} {0.system} {0.value}'.format(self)
 
+    def __key(self):
+        # Only use (system, value), as per unique constraint
+        return (self.system, self.value)
```
ivan-c: Can you add `self.id` to the beginning of that tuple?

tests/test_assessment_engine.py
```diff
@@ -1,5 +1,6 @@
 """Unit test module for Assessment Engine API"""
 import json
+import sys
```
ivan-c: I think this is leftover from when you were checking python versions, doesn't look like you need it anymore

portal/models/identifier.py
John0Randolph: Previously `__eq__` used only `self.system` and `self.value` and had that comment `Use only (system, value) as per unique constraint` - which makes me think that either id can change or there's another reason to not use it in `__eq__`, and therefore to not use it in `__key()`

tests/test_assessment_engine.py
John0Randolph: Good catch

portal/models/identifier.py
ivan-c: Good point!
@pbugni Maybe we should only check `self.id` in `__hash__()`? eg:

```
def __hash__(self):
  return hash((self.id,) + self.__key())
```

pbugni: no no - don't go looking for `id` - as that is a primary key set *after* `db.session.commit()`.
We use this, for example, when looking up identifiers from site persistence, they may be generated on the fly from a JSON file, and not yet committed.  
@John0Randolph 's impl of `__key__` looks spot on to me.

portal/models/identifier.py
```diff
@@ -48,9 +48,15 @@ def from_fhir(cls, data):
     def __str__(self):
         return 'Identifier {0.use} {0.system} {0.value}'.format(self)
 
+    def __key(self):
```
pbugni: I would suggest you follow the python `dunder` pattern used for "magic methods" - that is, double underscores on both sides of key, `def __key__(self): ...`

John0Randolph: Got it, will do.

PR #2286, opened 2018-06-19
TN-1264 Fix email config issue

ivan-c:
* Only use SSL when configured
  * Default email settings use local email server, without SSL
  * Mail relay requires SSL


ivan-c:
@pbugni I made this a hotfix so we can apply it to production without needing a release first. tested and seems to be working so I'll back-merge into develop

PR #2289, opened 2018-06-19
TN-1201 Respect browser locale preferences

ivan-c:
* Add browser fallback for locale configuration
  * If no explicit locale preferred (set in /profile)
  * If not logged in

### DIFF
--- a/portal/models/i18n.py
+++ b/portal/models/i18n.py
@@ -17,7 +17,9 @@ from polib import pofile
 import requests
 
 from ..extensions import babel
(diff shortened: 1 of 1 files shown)


ivan-c:
@pbugni Thanks! I found it in a discussion [here](https://github.com/pallets/werkzeug/issues/450).

Under the covers [it's just using a dict](https://github.com/python-babel/babel/blob/master/babel/core.py#L80-L89) to map languages to which is there is only one reasonable locale. ie `sv` (language) to `sv_SE` (locale)

PR #2297, opened 2018-06-20
Corrected PSA strategy as per TN-1262

pbugni:
PSA Strategy now:
if (TNG Patients AND localized PCa AND Tx Started) OR (TNG Patients AND w/o positive PCa dx)


mcjustin:
Just one thing, I think the commit msg should be 'TN-1262 Corrected PSA strategy', for jira integration

PR #2304, opened 2018-06-21
TN-1275 extend user/org api to copy org's timezone, language to user

pbugni:
As needed by Medidata/RAVE, extend `POST /api/user/<user_id>/organization` to enable mapping an organization's defaults for timezone and language to a user.

### DIFF
--- a/portal/views/organization.py
+++ b/portal/views/organization.py
@@ -8,12 +8,13 @@ from sqlalchemy import and_, exc
 from ..audit import auditable_event
 from ..database import db
 from ..extensions import oauth
+from ..models.coding import Coding
 from ..models.identifier import Identifier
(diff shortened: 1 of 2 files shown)


portal/views/organization.py
```diff
@@ -496,8 +508,34 @@ def add_user_organizations(user_id):
     if not request.json or 'organizations' not in request.json:
         abort(400, "Requires `organizations` list")
 
+    # validate input - don't allow multiple orgs with `apply_to_user`
+    # attributes for any given field
+    applied_fields = []
+
+    def apply_defaults(item, org, applied):
+        for field in 'timezone', 'language':
+            if field in item:
```
ivan-c: I might invert this condition and `continue` to de-indent the meat of the code a level

PR #2307, opened 2018-06-21
TN-1163 updated test_auth.py for py2 & py3 compatability

John0Randolph:


tests/test_auth.py
```diff
@@ -81,7 +82,7 @@ def test_client_bad_add(self):
         self.login()
         rv = self.client.post('/client',
                 data=dict(application_origins="bad data in"))
-        self.assertTrue("Invalid URL" in rv.data)
+        self.assertTrue("Invalid URL" in rv.data.decode('utf-8'))
```
ivan-c: Can you rename the `rv` variable? 

@pbugni `rv` is a command to pdb, if you use it as a variable name it's difficult to evaluate inside pdb :)

ivan-c: Can you use py.test-style `assert`?

ivan-c: I think you can get a unicode string from the response using `get_data(as_text=True)`, so you don't need to explicitly decode the string as unicode

tests/test_auth.py
```diff
@@ -106,7 +107,7 @@ def test_client_edit(self):
         # 200 response, because page is reloaded with validation errors
         self.assert200(rv2)
         error_text = 'URL host must match a provided Application Origin URL'
-        self.assertTrue(error_text in rv2.data)
+        self.assertTrue(error_text in rv2.data.decode('utf-8'))
```
ivan-c: See above

tests/test_auth.py
John0Randolph: What's a good name to rename rv to? Also, rv is in a handful of testing files - should I rename everywhere?

John0Randolph: And should I change everything I can to py.test-style `assert`?

ivan-c: >What's a good name to rename rv to? Also, rv is in a handful of testing files - should I rename everywhere?

It's difficult to tell in this context, but `rv` is a http response object (try `type(rv)`), my goto is usually just `response`

>And should I change everything I can in this file to py.test-style assert?

Entirely up to you :). The focus here is py3 compliance, but I'm just taking the opportunity to improve the code you're already changing.

tests/test_auth.py
```diff
@@ -127,20 +129,20 @@ def test_unicode_name(self):
         new_user = add_authomatic_user(authomatic_user, None)
 
         user = User.query.filter_by(email='test@test.org').first()
-        self.assertEqual(user.last_name, u'Bugn\xed')
-        self.assertEqual(new_user, user)
+        assert user.last_name == u'Bugn\xed'
```
ivan-c: You shouldn't need the `u` prefix since you imported `unicode_literals`

tests/test_auth.py
```diff
@@ -127,20 +129,20 @@ def test_unicode_name(self):
...
     def test_callback_validation(self):
         """Confirm only valid urls can be set"""
         client = self.add_client()
         self.login()
-        rv = self.client.post('/client/{0}'.format(client.client_id),
+        response = self.client.post('/client/{0}'.format(client.client_id),
                 data=dict(callback_url='badprotocol.com',
                     application_origins=client.application_origins))
-        self.assertEqual(200, rv.status_code)
+        assert 200 == response.status_code
```
ivan-c: I see Paul is a fan of [Yoda conditions](https://en.wikipedia.org/wiki/Yoda_conditions) :)

tests/test_auth.py
```diff
@@ -45,10 +47,10 @@ def test_local_user_add(self):
                 'retype_password': 'one2Three',
                 'email': 'otu@example.com',
                }
-        rv = self.client.post('/user/register', data=data)
-        self.assertEqual(rv.status_code, 302)
+        response = self.client.post('/user/register', data=data)
+        assert response.status_code == 302
         new_user = User.query.filter_by(username=data['email']).first()
-        self.assertEqual(new_user.active, True)
+        assert new_user.active == True
```
ivan-c: That `==` should be `is`

tests/test_auth.py
```diff
@@ -127,20 +129,20 @@ def test_unicode_name(self):
...
-        rv = self.client.post('/client/{0}'.format(client.client_id),
+        response = self.client.post('/client/{0}'.format(client.client_id),
                 data=dict(callback_url='badprotocol.com',
                     application_origins=client.application_origins))
-        self.assertEqual(200, rv.status_code)
+        assert 200 == response.status_code
 
         client = Client.query.get('test_client')
-        self.assertEqual(client.callback_url, None)
+        assert client.callback_url == None
```
ivan-c: That `==` should be `is`

PR #2309, opened 2018-06-21
TN-1282 Catch and report organization reference errors

pbugni:


tests/test_organization.py
```diff
@@ -621,6 +621,21 @@ def test_user_org_post(self):
         self.assert200(rv)
         self.assertEqual(len(rv.json['organizations']), 2)
 
+    def test_user_org_bogus_identifier(self):
+        self.shallow_org_tree()
```
ivan-c: @John0Randolph Just FYI these are the type of setUp-type functions that you might wanna try implementing as a [py.test fixture](https://docs.pytest.org/en/latest/fixture.html)

tests/test_organization.py
```diff
@@ -621,6 +621,21 @@ def test_user_org_post(self):
...
+        self.prep_org_w_identifier()
+        data = {'organizations': [
+            {'reference':
+                 'api/organization/123-45?system={}'.format(US_NPI[:-1])}
+        ]}
+        self.login()
+        rv = self.client.post(
```
ivan-c: Could you rename `rv` to `response`? `rv` is command inside pdb (return value) and although I could `print rv` instead it's not exactly the same and I'm lazy :)

pbugni: that's a rather common pattern in our tests - i have no problem with you replacing if you'd like - probably originally copied from some 3rd party test example.
meanwhile, don't want to wait on another test run - lets merge this.

PR #2313, opened 2018-06-22
TN-1280 followup fixes - visual fixes for expired questionnaire card

achen2401:
Followup fix for: https://jira.movember.com/browse/TN-1125

fix visual for expired questionnaire card


portal/eproms/templates/eproms/portal.html
```diff
@@ -65,7 +65,13 @@ <h2 class="tnth-subhead">{{ _("Welcome to TrueNTH") }}</h2>
                 </div>
             {%- else -%}
                 <div class="button-container">
-                    <a class="btn-lg btn-tnth-primary" href="{{ display.link_url }}">{{ display.link_label }}</a>
+                    <a class="btn-lg btn-tnth-primary" href="{{ display.link_url }}">
+                    	{%-if display.link_label %}
```
ivan-c: You might be able to condense this into one line with the [jinja `default` filter](http://jinja.pocoo.org/docs/2.10/templates/#default). 
I think you'll have to pass `True` since `display.link_label` is defined, but evaluates to `False` (see link for more details)
```
{{ display.link_label|default(_("Go to link"), true) }}
```
I think this might work too...
```
{{ display.link_label or _("Go to link") }}
```

portal/models/intervention_strategies.py
```diff
@@ -463,14 +463,20 @@ def completed_card_html(assessment_status):
                     name=user.display_name,
                     registry=assessment_status.assigning_authority)
             else:
+                greeting = _(u"Hi, {}").format(user.display_name)
```
ivan-c: I've recently remembered `_()` accepts format strings and keyword arguments to `_()`. 
This would give the translators a little more context, and would be recognized as a "placeholder" in Smartling, (as opposed to `{}`), preventing them from editing the placeholder:
```
_(u"Hi, %(full_name)s", full_name=user.display_name)
```

achen2401: @ivan-c thank you for the tip, ivan!

PR #2315, opened 2018-06-22
TN-1163 updated test_clinical.py for py2/py3 compatability

John0Randolph:


tests/test_clinical.py
```diff
@@ -47,27 +50,19 @@ def test_datetime_for_concept(self):
...
-            ['display'])
-        self.assertEqual(
-            '2',
-            clinical_data['entry'][0]['content']['valueQuantity']['value'])
-        self.assertEqual(
-            json.dumps(Reference.patient(TEST_USER_ID).as_fhir()),
-            clinical_data['entry'][0]['content']['performer'][0])
+        response = self.client.get('/api/patient/%s/clinical' % TEST_USER_ID)
+
+        clinical_data = json.loads(response.get_data(as_text=True))
```
ivan-c: Can you use `response.json` instead for this and all the below?

tox.ini
```diff
@@ -17,7 +17,7 @@ whitelist_externals = /bin/bash
 commands =
     {[base]commands}
     py27: py.test -v --cov portal --cov-report xml:"{toxinidir}/coverage.xml" []
-    py34: py.test -v tests/test_dict_tools.py tests/test_date_tools.py tests/test_access_on_verify.py tests/test_access_url.py tests/test_address.py tests/test_app_text.py tests/test_assessment_engine.py tests/test_assessment_status.py tests/test_audit.py tests/test_auth.py []
+    py34: py.test -v tests/test_dict_tools.py tests/test_date_tools.py tests/test_access_on_verify.py tests/test_access_url.py tests/test_address.py tests/test_app_text.py tests/test_assessment_engine.py tests/test_assessment_status.py tests/test_audit.py tests/test_auth.py tests/test_clinical.py[]
```
ivan-c: Could you add the space back? `[]` is a placeholder for extra arguments to `py.test` and without it you would need to pass an extra space every time.

tox.ini
```diff
@@ -8,7 +8,11 @@ deps =
     -rrequirements.txt
     pytest-cov
 passenv = TRAVIS* PG* SQLALCHEMY_DATABASE_TEST_URI PERSISTENCE_DIR FLASK_APP SECRET_KEY
-commands = py.test --cov portal --cov-report xml:"{toxinidir}/coverage.xml" []
+whitelist_externals = /bin/bash
```
ivan-c: Can you undo the changes to this file (so it's the same as the version in `develop`)?

While you were gone I merged [this PR](https://github.com/uwcirg/true_nth_usa_portal/pull/2319) which deletes the list of files in `tox.ini` and moves them to each individual test file (so I can run the same command for `py27` and `py34` tox environments)

John0Randolph: Gotcha, will do

PR #2321, opened 2018-06-25
TN-1286 Hotfix/recurring in process qbs

pbugni:
Sloppy check for presence of iteration failed for iteration ZERO.


portal/views/assessment_engine.py
```diff
@@ -42,6 +42,7 @@
 from ..models.role import ROLE
 from ..models.user import User, current_user, get_user_or_abort
 from .portal import check_int
+from ..trace import dump_trace, establish_trace
```
ivan-c: I think this import needs to be in the line above for the block to remain alphabetically sorted (`".."` < `".t"`)

PR #2373, opened 2018-06-29
TN-1279 Consent API improvements

pbugni:
Extended UserConsent to clearly report `acceptance_date` and a single audit row to capture when the action took place.


portal/migrations/versions/c49b7f7944a7_.py
```diff
@@ -0,0 +1,64 @@
...
+
+
+def downgrade():
+    # Don't see much value in supporting downgrade, so raising for now,
+    # as we'll lose data.
+    #
+    # Those consents w/ differing values in their audit rows from
+    # `acceptance_date` would need recreation of "recorded" audits
+    # in the downgrade.
+    raise RuntimeError(
```
ivan-c: This will exception will make it difficult to bisect changes automatically.
I think there's always a risk of losing data when reversing a migration, maybe a logging warning or error instead?

portal/migrations/versions/c49b7f7944a7_.py
```diff
@@ -0,0 +1,64 @@
...
+        recorded = session.query(Audit).filter(
+            Audit.comment == comment)
+        recorded = recorded.first()
+        if recorded:
+            assert(uc.audit.user_id == recorded.user_id)
+            assert(uc.audit.context == 'consent')
+            uc.audit.timestamp = recorded.timestamp
+            delete_em.append(recorded)
+
+    # Purge the now obsolete recorded audit rows
```
ivan-c: Would it make sense to add an audit record for these changes?

portal/migrations/versions/c49b7f7944a7_.py
```diff
@@ -0,0 +1,64 @@
...
+# revision identifiers, used by Alembic.
+revision = 'c49b7f7944a7'
+down_revision = '0f1576a4e220'
+Session = sessionmaker()
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('user_consents', sa.Column(
+        'acceptance_date', sa.DateTime(), nullable=True))
```
ivan-c: Should this be `acceptance_datetime`?

portal/migrations/versions/c49b7f7944a7_.py
pbugni: if we just log an error, the data is lost, as it'll be allowed to finish.  we want it to fall over.
keep in mind, this is on a `downgrade` only, not part of any rollout process.

portal/migrations/versions/c49b7f7944a7_.py
pbugni: i don't think so - all the data we need is still there, simply a better format for storage.

portal/migrations/versions/c49b7f7944a7_.py
pbugni: it's mirroring the string used in all the consent APIs - pretty significant change if you want that...

ivan-c: Fair enough :)

portal/migrations/versions/c49b7f7944a7_.py
ivan-c: >keep in mind, this is on a `downgrade` only, not part of any rollout process

That's why I'm not worried about protecting against a `downgrade`. 

The use-case I'm worried about this affecting is using [`git bisect`](https://git-scm.com/docs/git-bisect) to try and determine when a change was introduced. 
If the starting range of commits is large enough, it will span several migrations. In this situation, having a migration to automatically downgrade the DB to match the commit in question makes the bisect process much more "hands off" (and easier to automate).

PR #2380, opened 2018-07-02
TN-1279 Use acceptance date to deliver predictable sort in consent API

pbugni:


tests/test_consent.py
```diff
@@ -72,6 +73,40 @@ def test_user_consent(self):
...
+            organization_id=0, user_id=TEST_USER_ID,
+            agreement_url=self.url, audit=audit, acceptance_date=old)
+        with SessionScope(db):
+            db.session.add(uc1)
+            db.session.add(uc2)
+            db.session.add(uc3)
+            db.session.commit()
+        self.test_user = db.session.merge(self.test_user)
+        self.login()
+        rv = self.client.get('/api/user/{}/consent'.format(TEST_USER_ID))
```
ivan-c: Please use a variable name that doesn't conflict with `pdb` commands if you can :)

PR #2383, opened 2018-07-02
TN-1163 cleanup of files test_access_on_verify.py through test_audit.py

John0Randolph:


tests/test_app_text.py
```diff
@@ -110,17 +110,17 @@ def test_asset_variable_replacement(self):
...
         rf_id = int(resource.asset.split()[-1])
-        self.assertEqual(rf_id, TEST_USER_ID)
+        assert rf_id == TEST_USER_ID
 
         invalid_asset = "Not a real {variable}!"
         resource = UnversionedResource(test_url,
                                        asset=invalid_asset,
                                        variables=test_vars)
         error_key = resource.asset.split()[-1]
         if sys.version_info[0] < 3:
```
ivan-c: Is this version-dependent check still required even with the `unicode_literals` import?

tests/test_assessment_status.py
```diff
@@ -442,31 +443,30 @@ def test_metastatic_on_time(self):
...
+        assert not a_s.instruments_in_progress('all')
 
     def test_metastatic_due(self):
         # hasn't taken, but still in "Due" period
         self.bless_with_basics()  # pick up a consent, etc.
         self.mark_metastatic()
         self.test_user = db.session.merge(self.test_user)
         a_s = AssessmentStatus(user=self.test_user, as_of_date=now)
-        self.assertEqual(a_s.overall_status, "Due")
+        assert a_s.overall_status, "Due"
```
ivan-c: I think you meant `==` not `,` :)

tests/test_audit.py
```diff
@@ -1,4 +1,6 @@
 """Test module for audit functionality"""
+from __future__ import unicode_literals
```
ivan-c: Please add the `  # isort:skip` directive

tests/test_app_text.py
John0Randolph: It looks like it is

pbugni: Pulled this branch, and see why this check is needed:
```
>>> p asset
u'Missing asset variable u\\'variable\\''
```
That hardcoded `u\\` shouldn't be in there, me thinks...

John0Randolph: Agreed, just found where it's being added. Removing it...

PR #2386, opened 2018-07-03
TN-1163 updated test_demographics.py for py2/py3 compatability

John0Randolph:


tests/test_demographics.py
```diff
@@ -114,47 +112,47 @@ def test_demographicsPUT(self):
...
+        assert fhir['gender'] == gender.lower()
+        assert fhir['name']['family'] == family
+        assert fhir['name']['given'] == given
         # ignore added timezone and empty extensions
-        self.assertEqual(2, len(
-            [ext for ext in fhir['extension']
-             if 'valueCodeableConcept' in ext]))
-        self.assertEqual(3, len(fhir['careProvider']))
-        self.assertTrue(Reference.practitioner(pract_id).as_fhir() in fhir['careProvider'])
+        assert 2 == len([ext for ext in fhir['extension']
```
ivan-c: I might switch the assertion around next time to the more customary form (so the value you're checking against is on the right) if you're already changing the line

No need to change this now though
```
assert len([ext for ext in fhir['extension'] == 2
```

ivan-c:
Don't forget to tag py3! :)

tests/test_demographics.py
```diff
@@ -1,9 +1,9 @@
 """Unit test module for Demographics API"""
+from __future__ import unicode_literals
```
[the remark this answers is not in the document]
John0Randolph: Right, thanks

tests/test_demographics.py
John0Randolph: Gotcha

PR #2390, opened 2018-07-03
TN-1163 updated, cleaned up test_encounter, test_exclusion_persistence for py2/py3 compatability

John0Randolph:
test_encounter.py worked in py3 but was not marked as such. Test_exclusion_persistence contained a pair of duplicate methods.


portal/config/model_persistence.py
```diff
@@ -74,8 +75,9 @@ def persistence_filename(self):
 
     @staticmethod
     def _log(msg):
-        current_app.logger.info(msg)
-        trace(msg)
+        msg = msg.encode('utf-8')
```
ivan-c: I don't think you should need this if you're importing `unicode_literals`, what specific log entry/string is it failing on?

tests/test_exclusion_persistence.py
```diff
@@ -193,72 +191,10 @@ def test_connected_user(self):
             ep.import_(keep_unmentioned=True)
 
         result = User.query.get(owner_id)
-        self.assertEqual(result.email, 'sm-owner@gmail.com')
-        self.assertEqual(AuthProvider.query.count(), 1)
-        self.assertEqual(Token.query.count(), 1)
-        self.assertEqual(UserRelationship.query.count(), 1)
-
-    def test_preflight_valid(self):
```
ivan-c: What happened to these test functions (can't tell if they were deleted or not...)? Were they duplicates?

John0Randolph: Yes, they were duplicates (see below them)

portal/config/model_persistence.py
John0Randolph: The error I'm getting is: `UnicodeEncodeError: 'charmap' codec can't encode character '\u2713' in position 619: character maps to <undefined>`

And the string it's failing on is: `Message: "User 1 collision on import.  added {'id': 1}\nreplace {'status': 'registered'} with {'status': 'unknown'}\nreplace {'identifier': [{'value': '1', 'system': 'http://us.truenth.org/identity-codes/TrueNTH-identity', 'use': 'official'}, {'value': 'testy@ex
ample.com', 'system': 'http://us.truenth.org/identity-codes/TrueNTH-username', 'use': 'secondary'}]} with {'identifier': [{'value': 'None', 'system': 'http://us.truenth.org/identity-codes/TrueNTH-identity', 'use': 'official'}]}\nreplace {'photo': [{'url': 'http:
//examle.com/photo.jpg'}]} with {'photo': []}\nreplace {'name': {'given': '\u2713', 'family': 'Last'}} with {'name': {'given': None, 'family': None}}\n"`

tests/test_encounter.py
```diff
@@ -39,32 +37,29 @@ def test_encounter_as_fhir(self):
...
+        assert enc.auth_method == data['auth_method']
 
     def test_encounter_on_login(self):
         self.login()
-        self.assertEqual(len(self.test_user.encounters), 1)
-        self.assertEqual(
-            self.test_user.current_encounter.auth_method,
-            'password_authenticated')
+        assert len(self.test_user.encounters) == 1
+        assert self.test_user.current_encounter.auth_method == \
```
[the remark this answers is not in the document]
John0Randolph: Gotcha, on it

PR #2405, opened 2018-07-08
TN-1269 Feature/pca localized 4 assessment status

pbugni:
Now using the user_observation status to determine if a user has PCa_localized.

### DIFF
--- a/portal/migrations/versions/b4dcf331317e_.py
+++ b/portal/migrations/versions/b4dcf331317e_.py
@@ -1,0 +1,83 @@
+from alembic import op
+from flask import current_app
+import sqlalchemy as sa
+from sqlalchemy.orm import sessionmaker
+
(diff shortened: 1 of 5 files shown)


portal/migrations/versions/b4dcf331317e_.py
```diff
@@ -0,0 +1,83 @@
...
+        User.id == UserOrganization.user_id,
+        User.id == UserRoles.user_id,
+        UserRoles.role_id == patient_id,
+        UserOrganization.organization_id.in_(org_list))).with_entities(User)
+
+    for user in query:
+        if user.concept_value(CC.PCaLocalized) == 'true':
+            # user already has positive PCaLocalized observation
+            continue
+        else:
```
ivan-c: Don't think you need this else :)

PR #2406, opened 2018-07-09
TN-1163 cleanup & py3 compatability for more test files

John0Randolph:


requirements.txt
```diff
@@ -68,7 +68,7 @@ pycparser==2.18
 pycrypto==2.6.1
 pyflakes==2.0.0
 Pygments==2.2.0
-pytest==3.6.2
```
ivan-c: Can you revert this change? [Paul found that the newer version breaks our tests](https://github.com/uwcirg/true_nth_usa_portal/pull/2397)

PR #2409, opened 2018-07-09
TN-1312 Adds IRONMAN org Juravinski

mcjustin:
https://jira.movember.com/browse/TN-1312

### DIFF
--- a/portal/config/eproms/Organization.json
+++ b/portal/config/eproms/Organization.json
@@ -1321,6 +1321,34 @@
     {
       "extension": [
         {
+          "timezone": "America/New_York",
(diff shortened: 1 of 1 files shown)


portal/config/eproms/Organization.json
```diff
@@ -1318,6 +1318,34 @@
       },
       "resourceType": "Organization"
     },
+    {
+      "extension": [
+        {
+          "timezone": "America/New_York",
```
ivan-c: I think this should technically be [`America/Toronto`](https://time.is/Hamilton), but the offset should be the same

PR #2410, opened 2018-07-09
TN-1163 cleanup and py3 compatability for more test files

John0Randolph:


tests/test_user.py
```diff
@@ -809,27 +812,24 @@ def test_deep_tree_check_role(self):
         for perm in ('view', 'edit'):
             for patient in (
                 patient_w_id, patient_x_id, patient_y_id, patient_z_id):
```
ivan-c: This is an example of a good test to parameterize, you could pass each patient as a parameter FYI

tests/test_user_document.py
```diff
@@ -62,35 +58,36 @@ def test_post_patient_report(self):
...
 
-        test_contents = "This is a test."
+        test_contents = 'This is a test.'
         with NamedTemporaryFile(
             prefix='udoc_test_',
             suffix='.pdf',
             delete=True,
         ) as temp_pdf:
-            temp_pdf.write(test_contents)
+            temp_pdf.write(test_contents.encode('utf-8'))
```
ivan-c: I don't think you should need to do this since you're importing `unicode_literals`...

tests/test_user_document.py
```diff
@@ -62,35 +58,36 @@ def test_post_patient_report(self):
...
-                          INTERVENTION.SEXUAL_RECOVERY.description)
+        assert udoc.user_id == TEST_USER_ID
+        assert (udoc.intervention.description
+                == INTERVENTION.SEXUAL_RECOVERY.description)
 
 
     def test_download_user_document(self):
         self.login()
-        test_contents = "This is a test."
+        test_contents = b'This is a test.'
```
ivan-c: Please change back to native string :) (you'll likely need to change `BytesIO` as well)

portal/views/user.py
```diff
@@ -1642,7 +1642,7 @@ def posted_filename(req):
             abort(
                 400, "no file found - please POST a single file using "
                      "standard multipart/form-data parameters")
-        key = req.files.keys()[0]  # either 'file' or actual filename
+        key = list(req.files.keys())[0]  # either 'file' or actual filename
```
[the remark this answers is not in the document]
John0Randolph: What I found is that `.keys()` returns an iterator in py3 (instead of a list in py2), so wrapping `req.files.keys()` in `list()` forces it into a list so that the `[0]` works in py3

tests/test_user.py
John0Randolph: Gotcha

portal/views/user.py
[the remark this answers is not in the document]
ivan-c: Thanks for the detailed explanation, that makes perfect sense :)
In that case, try using [`next()`](https://stackoverflow.com/questions/4741243/how-to-pick-just-one-item-from-a-generator-in-python/4741269#4741269). that way the whole generator doesn't need to be evaluated:
```
key = next(req.files.keys())
```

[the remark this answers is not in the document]
John0Randolph: Gotcha, that works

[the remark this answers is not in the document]
ivan-c: Looks like it's erroring in python2 now, maybe try this?:
```
key = next(iter(req.files.keys()))

```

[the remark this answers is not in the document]
John0Randolph: Good catch, just saw that too. I'll try that.

tests/test_user_document.py
ivan-c: I couldn't find an easy way to fix this without specifying the encoding somewhere, but I realized that creating the `NamedTemporaryFile` is entirely unnecessary.

We're writing a string to `temp_pdf`, then just reading it into a `BytesIO`; ie `self.client.post` doesn't even reference `temp_pdf` (except for a filename). Can you remove the `NamedTemporaryFile` block and just create the file with `StringIO`? eg:

```
test_contents = 'This is a test.'
response = self.client.post(
    '/api/user/{}/patient_report'.format(TEST_USER_ID),
    content_type='multipart/form-data', 
    data=dict({'file': (StringIO(test_contents), 'udoc_test.pdf')}),
)

```

tests/test_user_document.py
ivan-c: Probably going to run into the same issue as above, please repeat above solution here

tests/test_user_document.py
John0Randolph: Good catch on removing the `NamedTemporaryFile`, but I'm still getting the same error - this time it's on the `StringIO(test_contents)` line.

`nose.proxy.TypeError: 'str' does not support the buffer interface`

Aren't we writing a string with `StringIO` this way, too (and thus have to encode it)?

pbugni: Looks like you (should NOT need it in python3)[https://stackoverflow.com/a/9368909] but will in py2.

pbugni: Please try [Martijn Pieters solution](https://stackoverflow.com/a/34872005) via imports.

John0Randolph: Good call, I'll take test_user_document into its own branch and try that solution.

John0Randolph: So that seems to be working for the `StringIO` bit, but now I'm getting an error on the line with `'udoc_test.pdf'` in this block:

        `response = self.client.post(
            '/api/user/{}/patient_report'.format(TEST_USER_ID),
            content_type='multipart/form-data',
            data=dict({'file':
                           (StringIO(test_contents),
                            'udoc_test.pdf')})
        )`

error: `nose.proxy.TypeError: 'str' does not support the buffer interface`
But if I encode `'udoc_test.pdf'` as a bytestring, then I get the error: `nose.proxy.TypeError: can't use a string pattern on a bytes-like object`

Any idea what is going on?

PR #2412, opened 2018-07-10
followup fixes to PR #2401 set response header for no caching

achen2401:
Followup fixes to PR:  https://github.com/uwcirg/true_nth_usa_portal/pull/2403
set the HTTP response headers to disallow caching
address 400 errors from retrieving optional core data for user

### DIFF
--- a/portal/static/js/profile.js
+++ b/portal/static/js/profile.js
@@ -616,7 +616,7 @@
                 if (targetSection.length > 0) {
                     var loadingElement = targetSection.find(".profile-item-loader");
                     loadingElement.show();
-                    this.modules.tnthAjax.getOptionalCoreData(self.subjectId, {useWorker: true, cache: true}, function(data) { //cache this request as change is rare if ever for optional data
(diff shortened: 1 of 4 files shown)


portal/static/js/profile.js
```diff
@@ -614,7 +614,7 @@
                 if (targetSection.length > 0) {
                     var loadingElement = targetSection.find(".profile-item-loader");
                     loadingElement.show();
-                    this.modules.tnthAjax.getOptionalCoreData(self.subjectId, {useWorker: true}, function(data) {
+                    this.modules.tnthAjax.getOptionalCoreData(self.subjectId, {useWorker: true, cache: true}, function(data) { //cache this request as change is rare if ever for optional data
```
ivan-c: I think we might be able to cache more requests without any effort if the default were to always cache, unless we know for sure we need fresh data

achen2401: @ivan-c ivan, i think more often than not we want fresh data especially in profile and initial queries page - will set default to always cache if I know for sure it is safe to do so after more rounds of testing.

PR #2413, opened 2018-07-10
TN-1269 followup fix and remove unused code

achen2401:
Followup fixes to this story:
https://jira.movember.com/browse/TN-1284

- remove putClinical API function code as it is not used by frontend (use postClinical uniformly now)
- add clinical code enum for global access


portal/static/js/main.js
```diff
@@ -18,6 +18,20 @@ var SYSTEM_IDENTIFIER_ENUM = {
     CANCER_TREATMENT_CODE: "118877007",
     NONE_TREATMENT_CODE: "999"
 };
+var CLINICAL_CODE_ENUM = {
```
ivan-c: @pbugni Just curious, is there an API that could be queried for this info?

PR #2414, opened 2018-07-10
TN-1125 followup fix FF scrolling fix

achen2401:
https://jira.movember.com/browse/TN-1125
FF scrolling fix - removing webkit scrollbar css changes


achen2401:
@mcjustin for aesthetics, added this sprint to not show scrolling bar - but should have tested out in all browsers before implementing it.

PR #2415, opened 2018-07-11
TN-1324 add `tree_view` to `api/organization`

pbugni:
Added an additional option to see a `tree` view of the registered organizations.

See issue including screen shot.


ivan-c:
@achen2401 Can you resolve the conflicts when you get a chance? Thanks!

portal/templates/org_tree.html
```diff
@@ -0,0 +1,78 @@
...
+    color: #FFF;
+}
+#fillOrgs .parent-singleton label {
+	top: 2px;
+	padding: 4px;
+}
+</style>
+<div id="fillOrgs" class="org-tree"></div>
+{% endblock %}
+{% block document_ready %}
```
ivan-c: Any way to do this without an inline script?

portal/templates/org_tree.html
```diff
@@ -0,0 +1,78 @@
+{% extends "layout.html" %}
+{% block main %}
+<style>
```
ivan-c: Is there a reason this can't be a separate CSS file?

PR #2417, opened 2018-07-11
TN-1329 Correct pcctc identifier for "Robert H..."

pbugni:
…nter Northwestern University"

This is preventing Medidata from associating patients with either org that had the accidental duplicate ID.

### DIFF
--- a/portal/config/eproms/Organization.json
+++ b/portal/config/eproms/Organization.json
@@ -695,7 +695,7 @@
         {
           "system": "http://pcctc.org/",
           "use": "secondary",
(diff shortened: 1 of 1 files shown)


ivan-c:
@pbugni Should this trigger an exception during `add_static_organization()`?

mcjustin:
Looks good, can't seem to officially review from my phone...

pbugni:
@ivan-c - there are legit uses for duplicate identifiers, such as to name which decision support tool applies to an org.  I don't think such business logic should be in `add_static_organization()`

PR #2422, opened 2018-07-12
TN-1163 cleanup & py3 compatability for test_portal.py

John0Randolph:


tests/test_portal.py
```diff
@@ -168,21 +171,25 @@ def test_message_sent(self):
...
         body = message.style_message(message.body)
-        self.assertTrue(u'DOCTYPE' in body)
-        self.assertTrue(u'style' in body)
-        self.assertTrue(isinstance(body, unicode))
+        assert u'DOCTYPE' in body
+        assert u'style' in body
+        if sys.version_info[0] < 3:
+            assert isinstance(body, unicode)
+        else:
+            assert isinstance(body, str)
```
John0Randolph: @pbugni @ivan-c this if statement seemed to be the best way to handle this test, as unicode in py2 = str in py3, and any other solution that I could think of seemed to defeat the point of the test. Thoughts?

tests/test_portal.py
```diff
@@ -168,21 +171,25 @@ def test_message_sent(self):
 
         # confirm styling unicode functions
         body = message.style_message(message.body)
-        self.assertTrue(u'DOCTYPE' in body)
-        self.assertTrue(u'style' in body)
-        self.assertTrue(isinstance(body, unicode))
+        assert u'DOCTYPE' in body
+        assert u'style' in body
+        if sys.version_info[0] < 3:
```
ivan-c: Try using `basestring` for your `isinstance()` check, then you won't need version-dependent code

tests/test_portal.py
```diff
@@ -168,21 +171,25 @@ def test_message_sent(self):
 
         # confirm styling unicode functions
         body = message.style_message(message.body)
-        self.assertTrue(u'DOCTYPE' in body)
-        self.assertTrue(u'style' in body)
-        self.assertTrue(isinstance(body, unicode))
+        assert u'DOCTYPE' in body
```
ivan-c: Please remove the unicode literal marks, they shouldn't be necessary since you imported `unicode_literals`

tests/test_portal.py
```diff
@@ -152,37 +155,40 @@ def test_invite(self):
...
 
     def test_message_sent(self):
         """Email invites - test view for sent messages"""
         sent_at = datetime.strptime("2000/01/01 12:31:00",
                 "%Y/%m/%d %H:%M:%S")
         message = EmailMessage(subject='a subject', user_id=TEST_USER_ID,
-                sender="testuser@email.com",
-                body=u'Welcome to testing \u2713', sent_at=sent_at,
-                recipients="one@ex1.com two@two.org")
+                               sender="testuser@email.com",
```
pbugni: Generic white space request: if you add a newline immediately after the open paren for these multi line statements, the indent is consistent and gives more room.  See https://www.python.org/dev/peps/pep-0008/#indentation

In other words, avoid `Aligned with opening delimiter` pattern - limits space and requires whole code blocks move on edits to the opening line.

John0Randolph: Gotcha, makes sense.

PR #2427, opened 2018-07-13
TN-1347 Don't allow future dates in consent API.

pbugni:
Makes sense to do as a hotfix, just to prevent Medidata from repeating the error.


pbugni:
@ivan-c it can probably just go in develop.  Added after Medidata made a call this change would prevent, that user was inaccessible.

PR #2428, opened 2018-07-13
Feature/verification email button fix

achen2401:
address https://jira.movember.com/browse/TN-1348
Gmail classic email app appears to ignore class selector - using **inline style** for button instead
Tested in IPAD, IPhone, IOS Chrome, IOS Safari, IOS Firefox, IOS email app, Windows IE, Windows Chrome.


portal/models/communication.py
```diff
@@ -85,7 +83,7 @@ def make_button(text):
             return text.replace('<a href', '<a class="btn" href')
 
     def _lookup_assessment_button():
-        return make_button(_lookup_assessment_link())
+        return make_button(_lookup_assessment_link(), True)
```
ivan-c: Can you pass `inline` as a keyword argument (easier to read)? ie:
```
return make_button(_lookup_assessment_link(), inline=True)
```

PR #2437, opened 2018-07-16
TN-1163 cleanup & py3 compatibility for test_celery.py

John0Randolph:


tests/test_celery.py
```diff
@@ -1,20 +1,17 @@
...
         """Try simply add task handed off to celery"""
         x = 151
         y = 99
-        rv = self.client.get('/celery-test?x={x}&y={y}&redirect-to-result=True'. \
-                             format(x=x, y=y), follow_redirects=True)
-        self.assert200(rv)
-        self.assertEqual(rv.data, str(x + y))
+        response = (self.client.get(
+            '/celery-test?x={x}&y={y}&redirect-to-result=True'.
+            format(x=x, y=y), follow_redirects=True))
```
ivan-c: Going forward , can you refactor URLs to use `requests.compat.urlencode` (if you're changing a line  for another issue already):
```
response = self.client.get(
    ''.join(
        '/celery-test?',
        requests.compat.urlencode(
            'redirect-to-result': 'True',
            'x': x,
            'y': y,
        ),
    )
)

```

John0Randolph: Ok, got it

PR #2444, opened 2018-07-17
TN-1033 & TN-1285 withdrawn patients

pbugni:
This PR includes fixes for the following sub-tasks:

TN-1039 consult consent status before sending reminders
TN-1353 Extend assessment_status.overall_status to hand out `Withdrawn`
TN-1354 Extend intervention_strategies.update_card_html_on_completion for Withdrawn
TN-1355 handle `withdrawn` in /api/present-needed

### DIFF
--- a/portal/models/assessment_status.py
+++ b/portal/models/assessment_status.py
@@ -14,6 +14,7 @@ from .fhir import (
 from .organization import OrgTree
 from .questionnaire_bank import QuestionnaireBank
 from .user import User
+from .user_consent import UserConsent
 
 
 def recent_qnr_status(user, questionnaire_name, qbd):
@@ -158,6 +159,9 @@ class QuestionnaireBankDetails(object):
 
     def overall_status(self):
         """Returns the `overall_status` for the given QB"""
+        if self.withdrawn():
+            return 'Withdrawn'
+
         if not (
             self.qbd.questionnaire_bank and
             self.qbd.questionnaire_bank.trigger_date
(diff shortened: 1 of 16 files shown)


portal/models/assessment_status.py
```diff
@@ -186,6 +190,37 @@ def overall_status(self):
...
+        intervention_qbs = QuestionnaireBank.query.filter(
+            QuestionnaireBank.intervention_id.isnot(None))
+        if self.qbd.questionnaire_bank in intervention_qbs:
+            return False
+
+        # With multiple root organizations, the consent lookup would
+        # be indeterminate - don't allow
+        root_orgs = OrgTree().find_top_level_org(self.user.organizations)
+        if len(root_orgs) > 1:
+            current_app.logger.error(
```
ivan-c: Should you also exit or throw an exception at this point (to prevent the lookup)?

portal/models/assessment_status.py
```diff
@@ -186,6 +190,37 @@ def overall_status(self):
...
+        # that if the QBD isn't intervention tied, the user is associated
+        # with the intervention by org.
+        intervention_qbs = QuestionnaireBank.query.filter(
+            QuestionnaireBank.intervention_id.isnot(None))
+        if self.qbd.questionnaire_bank in intervention_qbs:
+            return False
+
+        # With multiple root organizations, the consent lookup would
+        # be indeterminate - don't allow
+        root_orgs = OrgTree().find_top_level_org(self.user.organizations)
```
ivan-c: Out of scope for PR, but these frequently-used queries might benefit from being baked (caching the python -> SQL step). 

[OpenStack - Baked Queries](https://wiki.openstack.org/wiki/OpenStack_and_SQLAlchemy#Baked_Queries)
[SQLAlchemy - Baked Queries](http://docs.sqlalchemy.org/en/latest/orm/extensions/baked.html)

portal/models/assessment_status.py
pbugni: *shrug* - perhaps, but then we risk bringing everything to a halt when processing the hourly reminder tasks.

tests/test_procedure.py
```diff
@@ -175,9 +175,8 @@ def test_treatment_started(self):
         for code, display, system in started_codes:
             self.add_procedure(code, display, system)
             self.test_user = db.session.merge(self.test_user)
-            assert (known_treatment_started(self.test_user),
-                    "treatment {} didn't show as started".format(
-                        (system, code)))
+            assert known_treatment_started(self.test_user),\
+                "treatment {} didn't show as started".format((system, code))
```
ivan-c: What's the function of this line? I don't think it should be part of the `assert`...

tests/test_procedure.py
```diff
@@ -204,7 +203,7 @@ def test_treatment_not_started(self):
         for code, display, system in not_started_codes:
             self.add_procedure(code, display, system)
             self.test_user = db.session.merge(self.test_user)
-            assert (known_treatment_not_started(self.test_user),
-                    "treatment '{}' didn't show as not started".format(
-                        (system, code)))
+            assert known_treatment_not_started(self.test_user),\
+                "treatment '{}' didn't show as not started".format(
```
ivan-c: What's the function of this line? I don't think it should be part of the `assert`...

tests/test_procedure.py
pbugni: assert can take a second parameter, displayed when it fails to help us figure out the failure.

When enclosing it in parens, it will never fail with the second parameter.

PR #2448, opened 2018-07-17
TN-1163 removed u from test_communication.py

John0Randolph:


tests/test_communication.py
```diff
@@ -82,10 +82,10 @@ def test_dd_no_key(self):
 
     def test_unicode(self):
         dd = DynamicDictLookup()
-        dd['u'] = u'\u2713'
-        target = u'works {u}'
+        dd['u'] = '\u2713'
```
ivan-c: Can you try using the actual unicode character (`✓`) in the source?

John0Randolph: Which one is the source? Do you mean like:

``dd['u'] = '✓'``

ivan-c: Sorry, should've been clearer, I wanted both replaced, would be nice to have an actual unicode character in our files

PR #2452, opened 2018-07-17
TN-1163 removed u from __init__.py

John0Randolph:


tests/__init__.py
```diff
@@ -41,7 +43,7 @@
 
 TEST_USER_ID = 1
 TEST_USERNAME = 'testy@example.com'
-FIRST_NAME = u'\u2713'
+FIRST_NAME = '\u2713'
```
ivan-c: Can replace this with the actual character (`✓`)?

PR #2489, opened 2018-07-19
TN-351 formatted next 4 files from portal/models

John0Randolph:


portal/models/communication_request.py
```diff
@@ -162,10 +162,10 @@ def add_if_not_found(self, commit_immediately=False):
         existing = CommunicationRequest.query.filter(
             CommunicationRequest.questionnaire_bank_id ==
             self.questionnaire_bank_id).filter(
-                CommunicationRequest.notify_post_qb_start ==
-                self.notify_post_qb_start).filter(
-                    CommunicationRequest.qb_iteration ==
-                    self.qb_iteration).first()
+            CommunicationRequest.notify_post_qb_start ==
```
ivan-c: You might try something like this to keep the nesting obvious:
```
existing = CommunicationRequest.query.filter(
    CommunicationRequest.questionnaire_bank_id ==
    self.questionnaire_bank_id
).filter(
    CommunicationRequest.notify_post_qb_start ==
    self.notify_post_qb_start
).filter(
    CommunicationRequest.qb_iteration ==
    self.qb_iteration
).first()
```

pbugni: Yes, that is more legible.

John0Randolph: good call

PR #2496, opened 2018-07-20
TN-351 formatting for recur.py, reference.py, relationship.py, reporting.py

John0Randolph:


portal/models/reference.py
```diff
@@ -113,30 +113,27 @@ def parse(cls, reference_dict):
...
         def get_object_by_identifier(obj, system, value):
             if obj == Organization:
-                query = obj.query.join(OrganizationIdentifier).join(
-                    Identifier).filter(and_(
-                        Organization.id ==
-                        OrganizationIdentifier.organization_id,
-                        Identifier.id ==
-                        OrganizationIdentifier.identifier_id,
+                query = obj.query.join(
+                    OrganizationIdentifier).join(Identifier).filter(and_(
```
ivan-c: move the `join()`s to the same line, break on the `filter()`

John0Randolph: @ivan-c I can't break in the middle of the `filter()`, line is too long. I could break before the `.filter()`, but that requires wrapping the expression in parents and causes a bit more ugly indentation on the next couple lines - should I do that?

PR #2502, opened 2018-07-20
TN-351 formatting for patch_flask_user.py, patient.py, patients.py, portal.py

John0Randolph:


portal/views/portal.py
```diff
@@ -310,7 +310,9 @@ def verify_token(valid_seconds):
         target_url = getattr(NextStep, next_step)(user)
         if not target_url:
             # Due to access strategies, the next step may not (yet) apply
-            abort(400, "Patient doesn't qualify for '{}', can't continue".format(next_step))
+            abort(
+                400, "Patient doesn't qualify for '{}', can't continue".format(
```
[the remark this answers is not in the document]
John0Randolph: @ivan-c it looks like I still do, unless I'm missing something

portal/views/portal.py
```diff
@@ -825,14 +834,17 @@ def config_settings(config_key):
     )
     if config_key:
         key = config_key.upper()
-        if not any(key.startswith(prefix) for prefix in config_prefix_whitelist):
+        if not any(
+            key.startswith(prefix) for prefix in config_prefix_whitelist
+        ):
             abort(400, "Configuration key '{}' not available".format(key))
         return jsonify({key: current_app.config.get(key)})
     else:
```
ivan-c: Can you remove the `else` and un-indent this section? the `if` block is guaranteed to return/exit so `else` isn't necessary

portal/views/portal.py
```diff
@@ -823,18 +832,12 @@ def config_settings(config_key):
         'MEDIDATA_RAVE_ORG',
         'LOCALIZED_AFFILIATE_ORG'
     )
-    if config_key:
-        key = config_key.upper()
-        if not any(key.startswith(prefix) for prefix in config_prefix_whitelist):
-            abort(400, "Configuration key '{}' not available".format(key))
-        return jsonify({key: current_app.config.get(key)})
-    else:
-        config_settings = {}
```
ivan-c: Sorry, should've been more clear- we still need the logic in the `else` section (it returns ALL config key/values), but the `else` itself is unnecessary since we exit before the end of the `if` block

portal/views/portal.py
```diff
@@ -779,14 +786,16 @@ def settings():
...
+            for qnr in QuestionnaireResponse.query.filter_by(
+                    subject_id=patient.id):
                 qnr.authored = dt
                 document = qnr.document
                 document['authored'] = FHIR_datetime.as_fhir(dt)
                 # Due to the infancy of JSON support in POSTGRES and SQLAlchemy
                 # one must force the update to get a JSON field change to stick
                 db.session.query(QuestionnaireResponse).filter(
-                    QuestionnaireResponse.id == qnr.id).update({"document": document})
+                    QuestionnaireResponse.id == qnr.id).update(
```
ivan-c: Can you add a newline and put the `update()` part on its own line? (trying to keep the methods grouped with their options)

PR #2504, opened 2018-07-20
TN-351 formatting for scheduled_job.py, staff.py, truenth.py user.py

John0Randolph:


portal/views/user.py
```diff
@@ -226,7 +226,7 @@ def account():
         try:
             org_list = [Organization.query.filter_by(
                 id=org['organization_id']).one()
-                    for org in request.json['organizations']]
+                        for org in request.json['organizations']]
```
ivan-c: This should probably be at the least indentation level possible (it's not part of the variable that will get appended to the list)

portal/views/truenth.py
```diff
@@ -233,6 +234,7 @@ def portal_footer_html():
     )
     return make_response(html)
 
+
 ### Depricated rewrites follow
```
ivan-c: Typo ;)

PR #2513, opened 2018-07-20
TN-351 formatting test_procedure.py through test_scheduled_job.py

John0Randolph:


tests/test_procedure.py
```diff
@@ -179,8 +180,9 @@ def test_treatment_started(self):
         for code, display, system in started_codes:
             self.add_procedure(code, display, system)
             self.test_user = db.session.merge(self.test_user)
-            assert known_treatment_started(self.test_user),\
-                "treatment {} didn't show as started".format((system, code))
+            assert (
```
ivan-c: I think you need to use line continuations here, getting the following warning otherwise:

```
>>> assert (
...     False,
...     'blah')
<stdin>:1: SyntaxWarning: assertion is always true, perhaps remove parentheses?
```

ivan-c: @John0Randolph your fix seems to have been reverted in a merge, can you re-apply?

PR #2514, opened 2018-07-20
TN-351 formatting test_site_persistence.py through test_tou.py

John0Randolph:


tests/test_tou.py
```diff
@@ -101,7 +101,7 @@ def test_get_by_type(self):
 
         self.login()
         response = self.client.get('/api/user/{}/tou/privacy-policy'.format(
```
ivan-c: please break after `get(` instead

PR #2518, opened 2018-07-23
OAuth Using Flask Dance - Rough Draft

drryanjames:
These changes remove Authomatic and use [Flask-Dance](https://github.com/singingwolfboy/flask-dance) in its place to authenticate users as they access the portal. I'm still testing, but wanted to get these changes in front of others for an early review. **Please don't merge**

Changes include:
- Removal of Authomatic
- Addition of Flask-Dance routes for Google and Facebook
- A small class architecture for getting user info and storing it in a common format (this happenned automatically in authomatic)
- Common functions for logging the user in and handling errors

Known issues:
- When logging in with facebook at /login/facebook the url changes to /login/facebook/ which causes two auth requests to be sent to the facebook auth route which may cause one of the requests to fail. This should not be an issue in production since users will navigate to the proper url, the one ending in a /, when they click the login button.
- Still need to update tests. Many tests are failing and I'm not yet sure why
- I have not yet tried signing up using truenth creds. I will try this soon, but it should work.

If you find something else please let me know. I'm not sure what all to test but I did my best with the time I had. Looking forward to review comments!


README.rst
```diff
@@ -311,7 +311,7 @@ Testing
 To run the tests, repeat the
 ``postgres createuser && postgres createdb`` commands as above with the
 values for the {user, password, database} as defined in the
-``TestConfig`` class within ``portal.config.py``
+``BaseConfig`` class within ``portal\config\config.py``
```
pbugni: For `Testing`, I do believe it's in the ``TestConfig``:

```
    SQLALCHEMY_DATABASE_URI = testing_sql_url()
```

portal/extensions.py
```diff
@@ -108,45 +104,6 @@ def decorated(*args, **kwargs):  # pragma: no cover
...
-            secret=app.config['SECRET_KEY'],
-            debug=True,
-            logger=app.logger
-        )
-
-
-authomatic = _delay_init()
-
-
 mail = Mail()
```
pbugni: YES!  Thanks for eliminating that hack!

portal/factories/app.py
```diff
@@ -35,7 +34,7 @@
 from ..models.role import ROLE
 from ..views.assessment_engine import assessment_engine_api
 from ..views.audit import audit_api
-from ..views.auth import auth, capture_next_view_function
+from ..views.auth import auth, capture_next_view_function, facebook_blueprint, google_blueprint
```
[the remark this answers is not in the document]
pbugni: We are in the process of becoming pep8 compliant - please do fix all these warnings and limit line length to 79 chars.

portal/models/auth.py
```diff
@@ -27,6 +28,8 @@ class AuthProvider(db.Model):
                         nullable=False)
     user = db.relationship('User')
 
+    db.UniqueConstraint('provider', 'provider_id', name='uix_1')
```
pbugni: can we give that unique constraint a more legible name, say `auth_providers_by_provider`?  also, I think your migration script may need to include adding this constraint.  compare the test db with the one migrated.

README.rst
drryanjames: Updated

portal/extensions.py
drryanjames: Lol. My pleasure.

portal/models/auth.py
drryanjames: 1) Gave it the more legible name: `auth_providers_by_provider`
2) I was surprised this didn't happen automatically. I looked at another table and realized I should have set __table_args__. After I did this the migration set this automatically. Thanks for catching this.

drryanjames:
@pbugni thanks for the review! I will write tests tomorrow and update the PR (or submit a new one if you'd prefer that)

pbugni:
@drryanjames lets submit a new one, just as this is so noisy it's hard to find the content among all the pep8 feedback.

portal/views/auth.py
```diff
@@ -50,6 +52,329 @@
...
+    login_url='/login/google/',
+)
+
+facebook_blueprint = make_facebook_blueprint(
+    scope=['email', 'user_birthday', 'user_gender'],
+    login_url='/login/facebook/',
+)
+
+
+class FlaskDanceProvider:
```
drryanjames: @pbugni where is the best place to put these classes? Should I leave them in this file or put them in their own file(s)/folder(s). I left them here because it's consistent with other files, but this file is getting pretty long.

portal/views/auth.py
```diff
@@ -50,6 +52,329 @@
 
 auth = Blueprint('auth', __name__)
 
+google_blueprint = make_google_blueprint(
```
drryanjames: Should I leave the blueprints in this file or put them somewhere else?

drryanjames:
Creating a new one

PR #2520, opened 2018-07-23
TN-1366 multiple orgs in same branch of tree raise errors

pbugni:
Refactored OrgTree().find_top_level_orgs() to return a set (resolves duplicate problem) and param for obtaining first item.

### DIFF
--- a/portal/models/assessment_status.py
+++ b/portal/models/assessment_status.py
@@ -201,9 +201,7 @@ class QuestionnaireBankDetails(object):
         if self.qbd.questionnaire_bank in intervention_qbs:
             return False
 
-        # With multiple root organizations, the consent lookup would
-        # be indeterminate - don't allow
-        root_orgs = OrgTree().find_top_level_org(self.user.organizations)
+        root_orgs = OrgTree().find_top_level_orgs(self.user.organizations)
         if len(root_orgs) > 1:
             current_app.logger.error(
                 "Indeterminate org lookup - only expecting one root org "
@@ -278,7 +276,7 @@ class AssessmentStatus(object):
         for org in self.user.organizations:
             org_rp = org.research_protocol(self.as_of_date)
             if org_rp and org_rp.id == rp_id:
-                return OrgTree().find_top_level_org([org])[0]
+                return OrgTree().find_top_level_orgs([org], first=True)
         return None
 
     @property
--- a/portal/models/organization.py
+++ b/portal/models/organization.py
@@ -855,20 +855,26 @@ class OrgTree(object):
             node = node.parent
         return ids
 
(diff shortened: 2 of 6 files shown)


portal/models/organization.py
```diff
@@ -855,20 +855,26 @@ def at_and_above_ids(self, organization_id):
...
-        orgs_list = []
+        results = set()
         for org in (o for o in organizations if o.id):
             top_org_id = self.find(org.id).top_level()
-            orgs_list.append(Organization.query.get(top_org_id))
+            results.add(Organization.query.get(top_org_id))
 
-        return orgs_list
+        if first:
+            return next(iter(results)) if results else None
```
ivan-c: I don't think this will necessarily return the first org by id, sets don't retain order. Not sure if that's an issue however.

ivan-c: Why not call `next(iter(results))` on the return value of `find_top_level_orgs()`, after the function call?

portal/models/assessment_status.py
```diff
@@ -278,7 +276,7 @@ def __organization(self):
         for org in self.user.organizations:
             org_rp = org.research_protocol(self.as_of_date)
             if org_rp and org_rp.id == rp_id:
-                return OrgTree().find_top_level_org([org])[0]
+                return OrgTree().find_top_level_orgs([org], first=True)
```
ivan-c: Is this just masking the `Indeterminate org lookup` issue by picking the first one arbitrarily?

portal/models/organization.py
pbugni: Correct - but at this time, we are forced to not care.  I have the one log message to help where it matters.

I noticed *most* usage was just indexing the first element - also not certain to return a "first" - included this code in the common function to make for clean/clear client usage.

portal/models/assessment_status.py
pbugni: no differently than it was before - that issue remains.

portal/models/organization.py
ivan-c: Makes sense, maybe use an `OrderedDict` to mimic the old behavior (ordering), unless you're certain the ordering doesn't matter.

pbugni: I'm certain that it was indeterminate before, and no worse now.
You are assuming we know *which* org we might want, and that therefore ordering would help.  A false assumption.  The real fix is two fold:
1. hard business rule enforcement (we have a number of checks, leading us to primarily holding this true via annoying emails we fix).
2. a refactor to the QB engine so we can obtain the correct QB -> Org association at a low level.

I'd like to push this out as is.  It certainly doesn't make it any worse, and will eliminate at least one known false positive.

PR #2522, opened 2018-07-24
Provide config to avoid expensive cache flushing on sync.

pbugni:


manage.py
```diff
@@ -65,8 +65,9 @@ def upgrade_db():
 
 def flush_cache():
     """Flush redis of all values. Cached values may not longer correspond with new DB entries"""
-    r = redis.from_url(app.config['REDIS_URL'])
-    r.flushdb()
+    if app.config.get('FLUSH_CACHE_ON_SYNC', True):
```
ivan-c: Can you move the default to `config.py`?

PR #2523, opened 2018-07-24
Tune logging down - excessive noise and emails.

pbugni:


portal/models/assessment_status.py
```diff
@@ -471,8 +471,9 @@ def overall_assessment_status(user_id):
 
     """
     user = User.query.get(user_id)
-    current_app.logger.debug("CACHE MISS: {} {}".format(
-        __name__, user_id))
+    if current_app.config.get("LOG_CACHE_MISS", False):
```
ivan-c: I think a config specific to a single log entry is overkill (`grep -v 'CACHE MISS'` accomplishes the same).

Why not just turn down the log level if necessary?

pbugni: lots of useful stuff at debug level - this is just noise at this point.  I'll add to config.py knowing you'd like it there :)

PR #2552, opened 2018-08-09
TN-662 refactor inactivity timeout

pbugni:
See details in document https://docs.google.com/document/d/1LAXMyS0M4ttlArIhPQc6fVuB3oXN_XAI0HqCsSa2VdE/edit#

@achen2401 I think (given the complexity of http cookies) we might want to set `data-cookie-timeout` to the template variable `{{ expires_in }}` in `portal_wrapper.html` ??

### DIFF
--- a/portal/config/config.py
+++ b/portal/config/config.py
@@ -93,8 +93,10 @@ class BaseConfig(object):
     LOG_LEVEL = 'DEBUG'
 
     OAUTH2_PROVIDER_TOKEN_EXPIRES_IN = 4 * 60 * 60  # units: seconds
-    SS_TIMEOUT = 60 * 60  # seconds for session cookie, reset on ping
-    PERMANENT_SESSION_LIFETIME = SS_TIMEOUT
+    DEFAULT_INACTIVITY_TIMEOUT = 30 * 60  # default inactivity timeout
+    PERMANENT_SESSION_LIFETIME = 60 * 60  # defines life of redis session
+    SEXUAL_RECOVERY_TIMEOUT = 60 * 60  # SR users get 1 hour
+
     PIWIK_DOMAINS = ""
     PIWIK_SITEID = 0
     PORTAL_STYLESHEET = 'css/portal.css'
--- a/portal/static/js/sessionMonitor.js
+++ b/portal/static/js/sessionMonitor.js
@@ -6,7 +6,7 @@ var SessionMonitorObj = function() {
     // Set default sessionLifetime from Flask config
     // Subtract 10 seconds to ensure the backend doesn't expire the session first
     var DEFAULT_SESSION_LIFETIME;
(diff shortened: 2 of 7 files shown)


achen2401:
@pbugni paul, I like the idea of using the {{expires_in}} template variable after looking over the code.  It Is a cleaner approach.  I will see if i can refactor the frontend code tomorrow morning when my thinking won’t be as fuzzy.

portal/config/config.py
```diff
@@ -93,8 +93,10 @@ class BaseConfig(object):
     LOG_LEVEL = 'DEBUG'
 
     OAUTH2_PROVIDER_TOKEN_EXPIRES_IN = 4 * 60 * 60  # units: seconds
-    SS_TIMEOUT = 60 * 60  # seconds for session cookie, reset on ping
-    PERMANENT_SESSION_LIFETIME = SS_TIMEOUT
+    DEFAULT_INACTIVITY_TIMEOUT = 30 * 60  # default inactivity timeout
+    PERMANENT_SESSION_LIFETIME = 60 * 60  # defines life of redis session
```
ivan-c: This also sets the lifetime of the cookie FYI
http://flask.pocoo.org/docs/1.0/config/#PERMANENT_SESSION_LIFETIME

pbugni: Yes, but that's fine.  You'll note the cookie used to retain the /settings value is given a different lifetime.

portal/static/js/sessionMonitor.js
```diff
@@ -37,107 +20,113 @@ var SessionMonitorObj = function() {
...
                     onwarning: function() {
-                      var n = Math.round(l.timeBeforeWarning / 60 / 1e3),
-                          o = $('<div id="jqsm-warning">Your session will expire in ' + n + ' minutes. <button id="jqsm-stay-logged-in">Stay Logged In</button><button id="jqsm-log-out">Log Out</button></div>');
-                      $("body").children("div#jqsm-warning").length || $("body").prepend(o), $("div#jqsm-warning").show(), $("button#stay-logged-in").on("click", function(n) {
-                          n && n.stopPropagation(), l.extendsess(n);
-                      }).on("click", function() {
-                          o.hide();
-                      }), $("button#jqsm-log-out").on("click", l.logout);
+                        var n = Math.round(l.timeBeforeWarning / 60 / 1e3),
+                            o = $("<div id='jqsm-warning'>Your session will expire in " + n + " minutes. <button id='jqsm-stay-logged-in'>Stay Logged In</button><button id='jqsm-log-out'>Log Out</button></div>");
```
ivan-c: Should this be translated?

achen2401: @ivan-c ivan, this is default html if custom one is not provided - which we do provide and the custom html is tagged for translation.

PR #2556, opened 2018-08-10
OAuth: Auth should succeed when birthdate and gender are not returned by providers

drryanjames:
ACTUAL:
Facebook auth fails when a user hides their birthdate or gender

EXPECTED:
Auth succeeds when birthdate or gender are hidden.

Change Summary:
- Moved parsing logic to common function
- Added tests for failing scenarios
- Updated documentation

Before these changes, Facebook and Google json were parsed in separate functions. I checked for birthdate and gender properly in Google's parser, but didn't do so in Facebook's (oops). In addition, neither parser was tested because they were hard to test. These changes move parsing logic to a common function which minimizes provider specific code and allows me to test parse logic using a mock. This should also make new providers easier to add in the future.

### DIFF
--- a/portal/models/flaskdanceprovider.py
+++ b/portal/models/flaskdanceprovider.py
@@ -11,21 +11,121 @@ class FlaskDanceProvider:
     """
     __metaclass__ = ABCMeta
 
-    def __init__(self, blueprint, token):
+    def __init__(self, blueprint, token, standard_key_to_provider_key_map):
+        """constructor
+
+        :param blueprint: The provider blueprint
+        :param token: The user's access token
+        :param standard_key_to_provider_key_map: Each provider uses
+            different keys to encode user values. This maps
+            our standard keys, such as 'first_name' to provider keys,
+            such as 'given_name' in Google's case
+        """
         self.blueprint = blueprint
         self.token = token
+        self.standard_key_to_provider_key_map = \
+            standard_key_to_provider_key_map
 
     @property
     def name(self):
         return self.blueprint.name
 
-    @abstractmethod
     def get_user_info(self):
         """gets user info from the provider
 
+        This function parses json returned from the provider
+        and returns an instance of FlaskProviderUserInfo that is
+        filled with the user's information
+
+        :return an instance of FlaskProviderUserInfo with the
+            user's info
+        """
+
+        # Get the user's json
+        resp = self.send_get_user_json_request()
+
+        if not resp.ok:
+            current_app.logger.debug(
+                'Failed to fetch user info from {}'.format(self.blueprint.name)
+            )
+            return None
+
+        # Parse the json into a standard format
+        user_json = resp.json()
+        return self.parse_json(user_json)
+
+    def parse_json(self, user_json):
+        """parses the user's json and returns it in a standard format
+
+        Providers encode user information in json. This function parses
+        the json and stored values in an instance of FlaskProviderUserInfo
+
+        :param user_json: info about the user encoded in json
+
+        :return instance of FlaskProviderUserInfo with the user's info
+        """
+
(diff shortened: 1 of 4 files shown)


portal/models/flaskdanceprovider.py
```diff
@@ -11,21 +11,134 @@ class FlaskDanceProvider:
...
+                # Handle these cases gracefully
+                if not required and key not in value:
+                    return None
+
+                value = value[key]
+
+            return value
+
+        # Attempt to parse the user's json
+        try:
```
ivan-c: Would it make sense to move this try/except into `get_value_from_json()`, to get more granular exception info?

portal/models/flaskdanceprovider.py
```diff
@@ -11,21 +11,134 @@ class FlaskDanceProvider:
...
+                required=False
+            )
+            user_info.birthdate = get_value_from_json(
+                'birthdate',
+                required=False
+            )
+
+            return user_info
+        except KeyError as err:
+            current_app.logger.debug(
```
ivan-c: If we're missing a required value, I would probably consider that an error, ie `current_app.logger.error`

portal/models/flaskdanceprovider.py
```diff
@@ -11,21 +11,121 @@ class FlaskDanceProvider:
...
+        return self.parse_json(user_json)
+
+    def parse_json(self, user_json):
+        """parses the user's json and returns it in a standard format
+
+        Providers encode user information in json. This function parses
+        the json and stores values in an instance of FlaskProviderUserInfo
+
+        :param user_json: info about the user encoded in json
+
```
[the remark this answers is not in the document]
ivan-c: The mystery of the sphinx ;)

ivan-c:
FYI added @pbugni as a reviewer as well

portal/models/flaskdanceprovider.py
```diff
@@ -11,21 +11,134 @@ class FlaskDanceProvider:
...
+        filled with the user's information
+
+        :return FlaskProviderUserInfo with the user's info
+        """
+
+        # Ask the provider for details about the user
+        resp = self.send_get_user_json_request()
+
+        if not resp.ok:
+            current_app.logger.debug(
```
pbugni: I'm assuming this is an error/unexpected scenario?  Lets use `current_app.logger.error()` if so, so it doesn't go unnoticed.

portal/models/flaskdanceprovider.py
```diff
@@ -11,21 +11,134 @@ class FlaskDanceProvider:
...
+            self.standard_key_to_provider_key_map links standard keys to
+            provider specific keys, including those that are nested, which
+            allows our parsing code to stay as generic as possible.
+
+            :param standard:key: the standard key
+            :param required: is this property required?
+            :return value from the user's json or None
+            """
+
+            # Ge the key used by the provider
```
pbugni: typo `Ge`

portal/models/flaskdanceprovider.py
pbugni: Agreed

portal/models/flaskdanceprovider.py
drryanjames: I like that idea. Added a trycatch in ```get_value_from_json``` to record a more granular exception, then rethrow.

portal/models/flaskdanceprovider.py
drryanjames: Not sure why I keep making this mistake :open_mouth:

portal/models/flaskdanceprovider.py
drryanjames: I'm going back to elementary school.

PR #2558, opened 2018-08-10
fix handling of callback upon finish of ajax calls in profile/account creation pages

achen2401:
an attempt to address the issue encountered by user that was not able to set email address(es) of patients.  Fixes include:
- in profile page: prevent users from editing sections in profile until ajax calls to load section data are completed (e.g. the number of ajax requests in progress may have caused the cancellation of ajax call to update patient's demographics data)
- in account creation page:  wait for ajax calls to update to complete before re-directing to profile page
- minor fix to use lower-case email address to check for duplicate
- streamline code for params for send request


portal/static/js/main.js
```diff
@@ -2271,7 +2263,7 @@ var Global = {
                         addUserId = "&user_id=" + $el.attr("data-user-id");
                     }
                     if (emailReg.test(emailVal)) {  // If this is a valid address, then use unique_email to check whether it's already in use
-                        tnthAjax.sendRequest("/api/unique_email?email=" + encodeURIComponent(emailVal) + addUserId, "GET", "", null, function(data) {
+                        tnthAjax.sendRequest("/api/unique_email?email=" + encodeURIComponent(emailVal.toLowerCase()) + addUserId, "GET", "", null, function(data) {
```
mcjustin: Looks good per https://jira.movember.com/browse/TN-1418, thanks @achen2401

mcjustin:
I just created a story for this, please reference https://jira.movember.com/browse/TN-1420

portal/static/js/main.js
ivan-c: I'm not confident that [email addresses are entirely case-insensitive](https://stackoverflow.com/a/9808332), and even so I think this should probably be implemented in the `unique_email` API.

Could you open a new PR for this either way?

portal/templates/profile/profile_macros.html
```diff
@@ -10,6 +10,7 @@
     </div>
 {%- endmacro %}
 {% macro profileEditButton(class="") -%}
+    <span class="section-loader tnth-hide"><i class="fa fa-spinner fa-spin"></i> {{_("loading section data ...")}}</span>
```
ivan-c: Can you remove the extra space after "data "?

PR #2559, opened 2018-08-10
TN-1418 fix unique email check to ignore case

achen2401:
addressing TN 1418: https://jira.movember.com/browse/TN-1418

fix unique email API check to ignore case

*think* this is a right fix, but feedback on this is appreciated

### DIFF
--- a/portal/views/user.py
+++ b/portal/views/user.py
@@ -1398,7 +1398,8 @@ def unique_email():
     """
     email = request.args.get('email')
     validate_email(email)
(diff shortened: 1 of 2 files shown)


achen2401:
@mcjustin justin, hmm, i think so - assuming by accommodating, you mean that whatever a user entered for his/her email is preserved in the db.  note, this is only checking whether the email is **unique** regardless of case, no modification of existing email has occurred here.

PR #2566, opened 2018-08-15
TN-1407 API to reactivate deleted account.

pbugni:


portal/models/user.py
```diff
@@ -1575,8 +1575,8 @@ def delete_user(self, acting_user):
 
         if self == acting_user:
             raise ValueError("can't delete self")
-        if self is None or acting_user is None:
-            raise ValueError("both user and acting_user must be well defined")
+        if acting_user is None:
```
ivan-c: Just curious, why not `not acting_user` vs `is None`?

portal/models/user.py
```diff
@@ -1605,6 +1605,40 @@ def delete_user(self, acting_user):
...
+        for tracking purposes, and a new one will be created for posterity
+
+        :param self: user to reactivate
+        :param acting_user: individual executing the command, for audit trail
+
+        """
+        if not self.deleted:
+            raise ValueError("can't reactivate active user")
+        if self == acting_user:
+            raise ValueError("can't delete self")
```
ivan-c: should that be "reactivate"?

portal/models/user.py
```diff
@@ -1605,6 +1605,40 @@ def delete_user(self, acting_user):
...
+        if acting_user is None:
+            raise ValueError("reactivate requires well defined acting_user")
+
+        # The email was masked during delete.  Need to confirm a user didn't
+        # sneak in with the same address while deleted.  The accessor returns
+        # the unmasked value.
+        unmasked = self.email
+        if User.query.filter(User.email == unmasked).count() > 0:
```
ivan-c: Do we need to consider (lower)casing here?

portal/models/user.py
pbugni: no good reason - is None is clearly too specific for this case.  will fix

portal/models/user.py
pbugni: aye!

PR #2569, opened 2018-08-17
Feature/fe patients list deactivation reactivation ui

achen2401:
address https://jira.movember.com/browse/TN-1252
add UI changes to allow staff admin to deactivate/reactivate patient accounts
feature is enabled by config value
- deactivated accounts will be hidden by default
- checkbox is added to allow viewing of deactivated accounts
- UI button added to allow reactivation of a deactivated patient account


portal/static/js/admin.js
```diff
@@ -197,83 +234,155 @@
                 };
                 $("#adminTable").bootstrapTable(this.getTableConfigOptions(options));
             },
-            getTableConfigOptions: function(options) {
-                if(!options) {
+            getTableConfigOptions: function (options) {
+                if (!options) {
                     return this.tableConfig;
                 } else {
```
ivan-c: Don't think you need an `else` if you `return` in the `if`...

achen2401:
@ivan-c ivan, using ajax call is a good idea.  I will try if I get a chance.  The problem I was running into was trying to make sure the state of each data row (e.g. deactivated vs reactivated) stayed when the view of the bootstrap table changed - as when pagination or filtering is introduced.  Had to research the documentation a while (e.g. finding out which event to listen for and which method to use to update field data, etc).

PR #2576, opened 2018-08-20
TN-1252 followup fixes - css fix and event binding fix

achen2401:
followup fixes to story, TN-1252:
- noting that event is bound to the popover element multiple times - added fix to prevent that
- add css styling to popover in eproms, only did it for truenth USA before
- prevent reports from viewable for deactivated users (i.e. server error occurs if it happens)


achen2401:
closing this, need to make one more fix

PR #2579, opened 2018-08-21
TN-1252 followup fix - remove rowlink event from deleted account row

achen2401:
followup fix to https://jira.movember.com/browse/TN-1252

- prevent clicking on deactivate account row to navigate away to profile page
- prevent row from clickable from account is being deactivated


achen2401:
@ivan-c ivan, thank you, conflicts resolved

PR #2581, opened 2018-08-22
Add ApiKey authentication to Swagger Docs

John0Randolph:
Note that only 1/2 to 2/3 of the methods work with this (the ones with the lock icon), and none of the ones that have `roles required` - for those I'm in the process of adding `oauth`. But I think it merits its own PR as it should be able to be standalone, and likely will require a decent amount of QA and bug fixes.


portal/views/portal.py
```diff
@@ -919,6 +919,13 @@ def spec():
             },
         },
         "schemes": ("http", "https"),
+        "securityDefinitions": {
+            "Authorization": {
```
ivan-c: Can you rename this to `ServiceToken`? `Authorization` can be an overloaded term

portal/views/assessment_engine.py
```diff
@@ -51,10 +52,11 @@
...
     '/patient/<int:patient_id>/assessment',
     defaults={'instrument_id': None},
-)
+    methods=('OPTIONS', 'GET'))
 @assessment_engine_api.route(
-    '/patient/<int:patient_id>/assessment/<string:instrument_id>'
-)
+    '/patient/<int:patient_id>/assessment/<string:instrument_id>',
+    methods=('OPTIONS', 'GET'))
+@crossdomain(origin='*')
```
ivan-c: We'll need to protect these a little better (not allowing any domain), but still let Swagger UI to make CORS requests.

@pbugni thoughts?

Some ideas:
* Only allow in non-prod environments (`SYSTEM_TYPE != 'production'`)
* Only allow from truenth.org domains (our Swagger UI is hosted at stg.us.truenth.org/dist)

portal/views/assessment_engine.py
```diff
@@ -774,8 +778,9 @@ def get_assessments():
 
 @assessment_engine_api.route(
     '/patient/<int:patient_id>/assessment',
-    methods=('PUT',),
+    methods=('OPTIONS', 'PUT')
 )
+@crossdomain(origin='*', headers=('Content-Type', 'Authorization'))
```
ivan-c: Annoyed to find that the JSON `Content-Type` isn't listed here :/ (and thus why we need to add it to `Access-Control-Allow-Headers`)
https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS
```
The only allowed values for the Content-Type header are:
application/x-www-form-urlencoded
multipart/form-data
text/plain
```

portal/views/assessment_engine.py
```diff
@@ -774,8 +778,9 @@ def get_assessments():
 
 @assessment_engine_api.route(
     '/patient/<int:patient_id>/assessment',
-    methods=('PUT',),
+    methods=('OPTIONS', 'PUT')
```
ivan-c: @pbugni 
Can we add these additional methods (required for CORS) with the `@crossdomain` decorator?

portal/views/crossdomain.py
```diff
@@ -12,7 +12,11 @@
 def crossdomain(
         origin=None,
         methods=None,
-        headers=('Authorization', 'X-Requested-With', 'X-CSRFToken'),
+        headers=(
+            'Authorization',
+            'X-Requested-With',
+            'X-CSRFToken',
+            'Content-Type'),
```
ivan-c: I think you already added `Content-Type` to every place you're calling the `@crossdomain` decorator.

If I understand correctly and it's duplicated, I would remove it from the default `headers` here, or remove it from each place you added the `@crossdomain`

portal/views/assessment_engine.py
pbugni: I think we need a whitelist.  We have one in `models.client.validate_origin` - can this be extended with the known swagger hosts?

portal/views/assessment_engine.py
pbugni: I don't believe so - I think we will need to call out the additional `OPTIONS` method in the `methods` parameter to `route()`

ivan-c:
Closing in favor #2708, which is based on the latest work by John

PR #2583, opened 2018-08-23
TN-977 correct off by one error

pbugni:
Off by two, actually.  Zero index and strictly less than comparison.  Note, the failure is bumped AFTER the check if the current attempt is bad.

### DIFF
--- a/portal/views/auth.py
+++ b/portal/views/auth.py
@@ -374,7 +374,7 @@ def flask_user_password_failed_event(app, user, **extra):
     If this happens too often, for security reasons,
     the user will be locked out of the system.
     """
-    count = user.add_password_verification_failure()
+    count = user.password_verification_failures
     auditable_event(
         'local user failed password verification. Count "{}"'.format(
             count
--- a/portal/views/extend_flask_user.py
+++ b/portal/views/extend_flask_user.py
@@ -51,14 +51,18 @@ class LockoutLoginForm(LoginForm):
     def validate(self):
         """prevent locked out users from logging in
 
-        Before verifying the user's credentials
-        check to see if the user is locked out.
-        If the user is locked out display an error
+        If user has exceeded failed attempts, display an error
         message below the password field.
+
         """
+        success = super(LockoutLoginForm, self).validate()
+
         # Find user by email address (email field)
         user_manager = current_app.user_manager
         user = user_manager.find_user_by_email(self.email.data)[0]
+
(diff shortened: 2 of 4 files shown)


portal/views/extend_flask_user.py
```diff
@@ -51,15 +51,19 @@ class LockoutLoginForm(LoginForm):
...
         """prevent locked out users from logging in
 
-        Before verifying the user's credentials
-        check to see if the user is locked out.
-        If the user is locked out display an error
+        If user has exceeded failed attempts, display an error
         message below the password field.
+
         """
+        success = super(LockoutLoginForm, self).validate()
```
drryanjames: This can fail for reasons other than a password verification failure. See https://github.com/uwcirg/Flask-User/blob/v0.6.21.1/flask_user/forms.py#L192

portal/views/extend_flask_user.py
```diff
@@ -51,15 +51,19 @@ class LockoutLoginForm(LoginForm):
...
+
         # Find user by email address (email field)
         user_manager = current_app.user_manager
         user = user_manager.find_user_by_email(self.email.data)[0]
 
+        if not success:
+            user.add_password_verification_failure()
```
drryanjames: This should only be added when the user_password_failed signal is fired: https://github.com/uwcirg/Flask-User/blob/v0.6.21.1/flask_user/forms.py#L225

portal/views/auth.py
```diff
@@ -374,7 +374,7 @@ def flask_user_password_failed_event(app, user, **extra):
     If this happens too often, for security reasons,
     the user will be locked out of the system.
     """
-    count = user.add_password_verification_failure()
+    count = user.password_verification_failures
```
drryanjames: This is the only place `add_password_verification_failure` should be called.

portal/views/extend_flask_user.py
drryanjames: Took another look through this code and think we're safe. Thanks for making this change.

portal/views/extend_flask_user.py
drryanjames: Never mind, this is better. I was relying too heavily on the signal and should have though more about the desired logic

portal/views/auth.py
drryanjames: Nope. You're right. I like it in the other location better too.

PR #2592, opened 2018-08-25
Add lockout variables to config

drryanjames:
These changes add lockout variables (i.e. lockout period and # of failed attempts) to config to make them easier to change/update.

I also updated the names of a few variables to make the logic easier to follow (I confused myself).

### DIFF
--- a/instance/application.cfg.default
+++ b/instance/application.cfg.default
@@ -3,8 +3,6 @@ DEBUG = True
 SECRET_KEY = 'generate some random secret string'
 SQLALCHEMY_DATABASE_URI = 'postgresql://user:password@localhost/dbname'
 SERVER_NAME = '<the name and port number of the server (e.g.: 'myapp.dev:5000')>'
-LOCKOUT_PERIOD_MINUTES = 30
-FAILED_LOGIN_ATTEMPTS_BEFORE_LOCKOUT = 5
 
 # CONTACT_SENDTO_EMAIL = '<email address(es) that should receive contact form submissions>'
 # LOG_FOLDER = '/home/username/example/path'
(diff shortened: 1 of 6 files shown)


portal/views/auth.py
```diff
@@ -376,7 +376,7 @@ def flask_user_password_failed_event(app, user, **extra):
     If this happens too often, for security reasons,
     the user will be locked out of the system.
     """
-    count = user.password_verification_failures
+    count = user.add_password_verification_failure()
```
drryanjames: This is the appropriate place to increment verification failures. Since @pbugni added the call to ```LoginForm.validate()``` in his last fix incrementing here now works as expected.

portal/views/extend_flask_user.py
```diff
@@ -61,9 +61,6 @@ def validate(self):
         user_manager = current_app.user_manager
         user = user_manager.find_user_by_email(self.email.data)[0]
 
-        if user and not success:
-            user.add_password_verification_failure()
```
drryanjames: See previous comment in auth.py

instance/application.cfg.default
```diff
@@ -3,6 +3,8 @@ DEBUG = True
 SECRET_KEY = 'generate some random secret string'
 SQLALCHEMY_DATABASE_URI = 'postgresql://user:password@localhost/dbname'
 SERVER_NAME = '<the name and port number of the server (e.g.: 'myapp.dev:5000')>'
+LOCKOUT_PERIOD_MINUTES = 30
```
ivan-c: You'll want to move these defaults to the [`BaseConfig` class in config.py](https://github.com/uwcirg/true_nth_usa_portal/blob/develop/portal/config/config.py#L42)

`application.cfg.default` (and `application.cfg`) ideally are for config values that are required (do not and should not have a default)

portal/models/user.py
```diff
@@ -692,6 +688,21 @@ def locale_name_from_code(locale_code):
 
         return locale_options
 
+    @property
+    def lockout_period_minutes(self):
```
ivan-c: We usually just reference the `config` dict directly. I think I like this approach better though, makes things a little more decoupled

drryanjames: Same here. I was referencing it in multiple places at first and it felt like I was relying too much on the underlying implementation.

instance/application.cfg.default
drryanjames: I'll remember this next time.

PR #2595, opened 2018-08-27
Feature/initial queries error display fixes

achen2401:
address TN-1447: https://jira.movember.com/browse/TN-1447

allow error message to display upon return of API call to save data
changed included:
- adding indication that data is being saved
- wait for data saving to occur before proceeding to next section
- styling fixes
- remove unused/unnecessary code, refactor/streamline code


portal/static/js/initialQueries.js
```diff
@@ -481,6 +452,9 @@
                     case "select":
                         isComplete = field.val() !== "";
                         break;
+                    case "hidden":
```
ivan-c: Might be able to save a few lines if "select" and "hidden" [have the same case](https://stackoverflow.com/questions/7559531/javascript-switch-statement-is-there-a-way-for-two-cases-to-run-the-same-code/7559559#7559559)...

PR #2597, opened 2018-08-28
Fix selenium test and update documentation

drryanjames:


README.rst
```diff
@@ -358,9 +358,24 @@ TravisCI allows Selenium tests to be run with any number of browser/OS
...
 tests <https://saucelabs.com/open_sauce/user/ivan-c>`__.
 
-UI tests can also be run locally (after installing ``xvfb``) by passing
-Tox the virtual environment that corresponds to the UI tests (``ui``):
+UI tests can also be run locally (after installing ``xvfb`` and geckodriver) by passing
+Tox the virtual environment that corresponds to the UI tests (``ui``).
 
+Setup
+^^^^^
+- ``sudo apt-get install xvfb python-xvfbwrapper``
```
ivan-c: Could you remove `python-xvfbwrapper` here and add the below config to tox.ini (in the `[ui]` section)? Ideally, python modules should not be installed system-wide

```
deps =
    xvfbwrapper
    {[testenv]deps}
```

drryanjames: xvfbwrapper is already part of requirements.text, so I left it there. Added ```deps = {[testenv]deps}``` though

.travis.yml
```diff
@@ -81,6 +81,18 @@ before_install:
 
   - pip install --upgrade pip setuptools
 
+  # Install  plugins for ui tests
+  - if [ "$TOXENV" = "ui" ]; then
```
drryanjames: @ivan-c, is this where I should put this stuff? I was thinking either here or Dockerfile.build, but the Dockerfile seems to general for this specific case.

ivan-c: Sorry, I should've been more clear and given you more background. 

You shouldn't need to make any changes to this file, Sauce Labs uses a ["remote" web driver connection](https://github.com/SeleniumHQ/selenium/wiki/RemoteWebDriver), through [a tunnel established by the Sauce Labs plugin for TravisCI](https://docs.travis-ci.com/user/sauce-connect/).

The "geckodriver" is Firefox-specific, and only needed when running tests locally (the remote web driver is included in selenium, but the gecko driver isn't :/)

FYI If you ever see TravisCI complaining about the missing geckodriver, it's likely that the Sauce Labs tunnel/connection failed and the `ui` job needs to be restarted

tox.ini
```diff
@@ -17,6 +17,7 @@ commands = sphinx-build -W -n -b html -d {envtmpdir}/doctrees source {envtmpdir}
 
 [testenv:ui]
 description = Run selenium tests
+deps = {[testenv]deps}
```
ivan-c: Didn't realize `xvfbwrapper` was in `requirements.txt`, could you remove it and add it here as it appears below?
```
deps =
    xvfbwrapper
    {[testenv]deps}
```

Ideally testing dependencies are only installed during testing, thanks for fixing my oversight!

FYI `[testenv:ui]` inherits from `[testenv]`, so `deps = {[testenv]deps}` isn't required as it's already inherited :)

.travis.yml
mcjustin: TN-1468 <- @drryanjames please add "TN-1468 " to the front of your next commit comment

drryanjames: @mcjustin I created a new PR with the Jira ticket in the title.

https://github.com/uwcirg/true_nth_usa_portal/pull/2602

drryanjames:
I'm going to close this PR and go with the new one from a local branch that includes the Jira ticket number in the title.

PR #2600, opened 2018-08-28
Feature/initial queries followup cleanup fixes

achen2401:
followup code clean-up/fixes related to previous PR for this story:  
https://jira.movember.com/browse/TN-1447
- removal of all frontend code that compile incomplete data fields - relying solely on still needed data API as the source of truth for still needed data
- dynamically generate section objects via template attributes
- remove unused/unnecessary code
- next button styling fix


portal/static/js/initialQueries.js
```diff
@@ -908,72 +742,42 @@
...
+        $("input[name='user_type']").on("click", function() {
+            var roles = [], theVal = $(this).val();
+            roles.push({name: theVal});
+            tnthAjax.putRoles(self.userId,{"roles": roles}, $("#rolesGroup"));
+            self.handlePostEvent(self.getSectionContainerId($(this)));
         });
     };
 
-    FieldsChecker.prototype.patientQEvent = function(fields) {
+    FieldsChecker.prototype.patientQEvent = function() {
```
ivan-c: Could these validation functions be generalized for specific element types, or reused in some way?

achen2401: @ivan-c think so, ivan..although i think some of these behaviors are specific enough to warrant custom validation functions.  let me think this through a bit, will do a PR when I have a better idea how to change these.

PR #2601, opened 2018-08-29
TN-1447 followup fix - fix timing of saving loader apperance

achen2401:
found issues while testing this story:  https://jira.movember.com/browse/TN-1447
- remove synchronous ajax call that locked up UI, use async call instead
- hide saving loading indicator on callback of ajax call, otherwise there was a brief blank screen when UI pieces are transitioning.


drryanjames:
The same integration test started failing for me today. @ivan-c do you know what's going on

https://travis-ci.org/uwcirg/true_nth_usa_portal/jobs/421894444

PR #2609, opened 2018-08-30
Feature/login registration mobile style fixes

achen2401:
address https://jira.movember.com/browse/TN-1442
Fix mobile styling issues for login and registration pages:
- center content
- fix logo content overflowing issue in footer
- general styling fixes e.g. spacing


portal/templates/flask_user/login.html
```diff
@@ -1,5 +1,5 @@
 {% extends 'flask_user/public_base.html' %}
-
+{% block mainclass -%}class="login-form"{%- endblock %}
```
ivan-c: Was just looking at the [whitespace stripping](http://jinja.pocoo.org/docs/dev/templates/#whitespace-control) with `-%}`. Will that have any effect since there's no whitespace to remove between the block tags?

achen2401: @ivan-c ivan, you are right, there won't be any effect.  thank you catching this. I was doing this out of habit.

PR #2612, opened 2018-08-30
Refactor long running celery tasks

pbugni:
The big celery tasks that loop over all necessary patients have been refactored to a classic producer / consumer model, to massively shorten the time to wait when shutting down celery, and create a more efficient asynchronous system where multiple workers can in parallel work on a number of tasks at a time.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -220,11 +220,12 @@ Alternatively, install an init script and configure. See
 `Daemonizing Celery <http://docs.celeryproject.org/en/latest/tutorials/daemonizing.html>`__
 
 Should the need ever arise to purge the queue of jobs, run the following
-**destructive** command
+**destructive** command.  All tasks should be idempotent by design, so doing
+this is suggested, especially if the server is struggling.
 
 .. code:: bash
 
-    $ celery purge --app portal.celery_worker.celery
+    $ celery purge --force --app portal.celery_worker.celery
 
 Without running ``purge``, celery will resume any unfinished tasks when it restarts
 
--- a/README.rst
+++ b/README.rst
@@ -224,7 +224,9 @@ Should the need ever arise to purge the queue of jobs, run the following
 
 .. code:: bash
 
-    $ celery --app portal.celery_worker.celery purge
+    $ celery purge --app portal.celery_worker.celery
+
+Without running ``purge``, celery will resume any unfinished tasks when it restarts
 
 DATABASE
 --------
--- a/portal/tasks.py
+++ b/portal/tasks.py
@@ -10,8 +10,11 @@ NB: a celery worker must be started for these to ever return.  See
 from datetime import datetime
 from functools import wraps
 import json
+import random
+import time
 from traceback import format_exc
 
+from celery.exceptions import SoftTimeLimitExceeded
 from celery.utils.log import get_task_logger
 from flask import current_app
 from requests import Request, Session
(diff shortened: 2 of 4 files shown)


portal/tasks.py
```diff
@@ -138,6 +141,31 @@ def test_args(*args, **kwargs):
...
+    except SoftTimeLimitExceeded:
+        logger.info("timed out")
+
+
+@celery.task(name="tasks.produce_list")
+def produce_list():
+    j = 0
+    step = 5
+    numlists = 50
+    for i in range(step, (step*numlists)+1, step):
```
ivan-c: I think it might be more straightforward to create one range (1-250), then slice from it in the loop

portal/tasks.py
```diff
@@ -182,41 +212,81 @@ def update_patient_loop(update_cache=True, queue_messages=True):
     valid_patients = User.query.join(UserRoles).filter(and_(
         User.id == UserRoles.user_id,
         User.deleted_id.is_(None),
-        UserRoles.role_id == patient_role_id))
+        UserRoles.role_id == patient_role_id)).with_entities(User.id)
+
+    patients = valid_patients.all()
+    j = 0
+    batchsize = 16
+    while True:
```
ivan-c: this is [my favorite way to chunk lists](https://stackoverflow.com/a/435712) I've found, 

ie:
```
for chunk in zip(*[iter(patients)]*batchsize):
...
```
[this might be better though ](https://stackoverflow.com/a/8312908)(doesn't skip the "remainder" if `patients` is not divisible by chunk size)

portal/tasks.py
```diff
@@ -182,41 +212,81 @@ def update_patient_loop(update_cache=True, queue_messages=True):
     valid_patients = User.query.join(UserRoles).filter(and_(
         User.id == UserRoles.user_id,
         User.deleted_id.is_(None),
-        UserRoles.role_id == patient_role_id))
+        UserRoles.role_id == patient_role_id)).with_entities(User.id)
+
+    patients = valid_patients.all()
+    j = 0
+    batchsize = 16
```
ivan-c: Can you make this configurable? much easier to test on stg* than editing code. would be willing to remove configurability later once we've settled on a good default

portal/tasks.py
```diff
@@ -138,6 +141,31 @@ def test_args(*args, **kwargs):
...
+@celery.task(name="tasks.consume_list", ignore_result=True, soft_time_limit=4)
+def consume_list(id_list, priority):
+    try:
+        time.sleep(random.randint(1, 5))
+        logger.info("priority: {} consuming {}".format(priority, id_list))
+    except SoftTimeLimitExceeded:
+        logger.info("timed out")
+
+
+@celery.task(name="tasks.produce_list")
```
ivan-c: How is this triggered? I saw it on stg-eproms but I didn't call `/test-producer-consumer`

README.rst
```diff
@@ -224,7 +224,9 @@ Should the need ever arise to purge the queue of jobs, run the following
 
 .. code:: bash
 
-    $ celery --app portal.celery_worker.celery purge
+    $ celery purge --app portal.celery_worker.celery
```
ivan-c: Can you add `--force` here? otherwise yes/no will be prompted

portal/tasks.py
```diff
@@ -162,17 +190,19 @@ def cache_assessment_status(**kwargs):
...
 def prepare_communications(**kwargs):
     """Move any ready communications into prepared state """
-    update_patient_loop(update_cache=False, queue_messages=True)
+    update_patient_loop(
+        update_cache=False, queue_messages=True, as_task=True)
 
 
-def update_patient_loop(update_cache=True, queue_messages=True):
+def update_patient_loop(
+        update_cache=True, queue_messages=True, as_task=False):
```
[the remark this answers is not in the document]
pbugni: perhaps - then the tests all need to change.  it needs special handling to run as a task, which is why i chose that default.

portal/tasks.py
```diff
@@ -138,6 +141,31 @@ def test_args(*args, **kwargs):
     return "{}|{}".format(",".join(args), json.dumps(kwargs))
 
 
+@celery.task(name="tasks.consume_list", ignore_result=True, soft_time_limit=4)
+def consume_list(id_list, priority):
```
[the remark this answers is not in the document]
pbugni: Yes, I'll add docstrings - yes only for testing/PoC

portal/tasks.py
```diff
@@ -138,6 +141,31 @@ def test_args(*args, **kwargs):
...
+    try:
+        time.sleep(random.randint(1, 5))
```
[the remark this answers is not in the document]
pbugni: sleep is to make it actually take some time, so we can observe the PoC, otherwise, it's immediate and you can't see how a long running task with varying priorities behaves.

portal/tasks.py
pbugni: it's just PoC, but if you want to submit a patch, that'd be fine :)  as is, you can toggle each independently

portal/tasks.py
pbugni: sure.

portal/tasks.py
pbugni: that is the trigger -  GET `/test-producer-consumer`

README.rst
pbugni: good point.

portal/tasks.py
pbugni: not sure that makes the code any easier to read, we'd like to process the whole list, and then i need to make the consumer None safe...

I'll change it if you want it, tho?

ivan-c: I'm just paranoid about infinite loops, no need to change

PR #2617, opened 2018-09-05
TN-1408 Prevent promoting a temporary account with a power user account

drryanjames:
https://jira.movember.com/browse/TN-1408

Problem:
When a patient is registered it's possible for their account to be promoted to a power user account (e.g. staff, admin, ect...)

Repo Steps:
Login as a staff member
Navigate to /patients (or use the menu and select the Patients menu item)
Create a new patient
Send that patient and invitation
Open the new patient's email and click the "verify account" button
Verify the patient's identity by entering in their name and DOB and clicking the verify button
Critical Step instead of registering the new patient, login as a staff member
 

ACTUAL BEHAVIOR:
Temp account created by the staff member is promoted to power user and the staff member's account is deleted

 

EXPECTED BEHAVIOR:
Failure - 400. Temp account should not be promoted to power user account.

 

Diagnosis:
When a staff creates a temporary record for a new patient the temporary record is later merged with, or promoted to, the user account that the patient creates when they register. We do this by tracking the id of the temporary account through our registration workflow and merge it with the first user that logs in. We expect this user to be a newly registered created by the patient. However, during this ticket a power user was logged in instead of creating a new user account. This caused the temporary patient record to be merged with, or promoted to, the power user.

Fix:
Fail with a 400 when a temporary account tries to promote to a power user. When a user is being promoted, we will check to see if the signed in account is a power user. If so, we'll fail the request.

### DIFF
--- a/portal/config/config.py
+++ b/portal/config/config.py
@@ -1,7 +1,8 @@
 """Configuration"""
 import os
-
 import redis
+
+from portal.models.role import ROLE
 
 SITE_CFG = 'site.cfg'
 
@@ -173,6 +174,16 @@ class BaseConfig(object):
 
     LOCKOUT_PERIOD_MINUTES = 30
     FAILED_LOGIN_ATTEMPTS_BEFORE_LOCKOUT = 5
+
+    RESTRICTED_FROM_PROMOTION_ROLES = [
+        ROLE.ADMIN.value,
+        ROLE.APPLICATION_DEVELOPER.value,
+        ROLE.CONTENT_MANAGER.value,
+        ROLE.INTERVENTION_STAFF.value,
+        ROLE.STAFF.value,
+        ROLE.STAFF_ADMIN.value,
+        ROLE.SERVICE.value,
+    ]
 
 
 class DefaultConfig(BaseConfig):
--- a/portal/models/user.py
+++ b/portal/models/user.py
@@ -1497,8 +1497,7 @@ class User(db.Model, UserMixin):
                 error_message = 'Attempted to promote temporary user {} \
                         to registered user {} with role {}' \
                         .format(self.id, registered_user.id, restricted_role)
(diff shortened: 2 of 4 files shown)


portal/config/config.py
```diff
@@ -174,6 +175,16 @@ class BaseConfig(object):
     LOCKOUT_PERIOD_MINUTES = 30
     FAILED_LOGIN_ATTEMPTS_BEFORE_LOCKOUT = 5
 
+    RESTRICTED_FROM_PROMOTION_ROLES = [
```
ivan-c: Will a list of strings in `application.cfg` work to configure this? (ie is a list of objects required?)

portal/models/user.py
```diff
@@ -1516,6 +1529,13 @@ def promote_to_registered(self, registered_user):
             user_id=self.id, subject_id=self.id,
             context='account'))
 
+        current_app.logger.info(
```
ivan-c: I think this should probably be an audit or debug entry...

portal/models/user.py
```diff
@@ -1487,6 +1487,19 @@ def merge_with(self, other_id):
...
+        # https://jira.movember.com/browse/TN-1408
+        restricted_roles = \
+            current_app.config['RESTRICTED_FROM_PROMOTION_ROLES']
+        for restricted_role in restricted_roles:
+            if registered_user.has_role(restricted_role):
+                error_message = 'Attempted to promote temporary user {} \
+                        to registered user {} with role {}' \
+                        .format(self.id, registered_user.id, restricted_role)
+                current_app.logger.error(error_message)
+                return False
```
pbugni: I think it'd be better to raise an exception here - this is an exceptional case, and we don't want to rely on the client to check the return value (plus, they don't know *why* it failed).

PR #2627, opened 2018-09-05
TN-1447 fix music initial queries flow

achen2401:
followup fix to this story:  https://jira.movember.com/browse/TN-1447
fix broken initial queries work flow for MUSIC after UI refactor


portal/static/js/initialQueries.js
```diff
@@ -473,6 +473,12 @@
                 $("#aboutForm").fadeIn();
             }
         }
+        if ($("#next").hasClass("accept-on-next")) { //related to MUSIC - Accept on Next - special case
```
pbugni: I'd rather avoid business rules in the code - is this specific to MUSIC, or more general if a config turns on the `accept-on-next` feature?

pbugni: therefore, i'd be tempted to drop the "related to MUSIC" and make the comment more like what you just told me :)

PR #2634, opened 2018-09-06
Remove noisy warning message polluting the logs, no value.

pbugni:


portal/models/assessment_status.py
```diff
@@ -223,8 +223,10 @@ def withdrawn(self):
                 if valid_consent and valid_consent[1] == 'suspended':
                     return True
 
-        if not valid_consent_found:
-            current_app.logger.warn("No consent found for {}".format(self.user))
+        # no longer considered a concern, ignore this situation, well handled.
```
ivan-c: In an unrelated matter, I've seen these entries in the celery workers. 

Would it save us any computation to exclude unconsented users from caching/emails?

PR #2635, opened 2018-09-06
Sanely handle non-ascii identifier values.

pbugni:
Fix problem noticed when a user w/ non-ascii chars in email address caused server error.


tests/test_identifier.py
```diff
@@ -39,3 +39,9 @@ def testPOST(self):
         assert len(response.json['identifier']) == expected
         user = User.query.get(TEST_USER_ID)
         assert len(user.identifiers) == expected
+
+    def test_unicode_value(self):
```
ivan-c: Does this necessarily need to depend on the base unit test class? (Don't bother if it's not worth the effort :)

PR #2636, opened 2018-09-07
TN-1479 IRONMAN org + Ottawa, fix Emory parent (wasn't live)

mcjustin:


mcjustin:
Noting that https://github.com/uwcirg/true_nth_usa_portal/pull/2626 had created Emory (not yet live) as a child of Canada - now fixed to be a child of USA.

PR #2638, opened 2018-09-07
TN-1253 allow deactivated user view on staff list, fix staff admin role

achen2401:
followup fixes for this story:  https://jira.movember.com/browse/TN-1253
- allow granting of staff admin role for both eproms and truenthUSA, remove the need for config var
- allow deactivated users filter view on staffs list - allow admin staff to reactivate account if necessary
- styling fixes


portal/config/gil/site_persistence_file.json
```diff
@@ -20,7 +20,6 @@
         "TRUENTH_LINK_URL='https://us.movember.com/'\n",
         "MOVEMER_LINK_URL='https://us.movember.com/'\n",
         "STAFF_BULK_DATA_ACCESS=False\n",
-        "ALLOW_STAFFADMIN_DELEGATION = True\n",
```
ivan-c: Does this config also need to be deleted from this template?

https://github.com/uwcirg/true_nth_usa_portal/blob/develop/portal/templates/profile/staff_profile_create.html#L19

achen2401: @ivan-c ivan, thank u very much for catching that.

PR #2639, opened 2018-09-07
TN-845 Add PSA Tracker to TN USA

pbugni:
Copied over and slightly modified details for PSA Tracker intervention to fit TN USA - most significant, the access_strategy doesn't require the TNG org (as it doesn't exist in TN USA).

Note, we don't have the dynamic layout the issue requested - took a compromised guess at which `display_rank` to use.


mcjustin:
display rank looks good to me

PR #2641, opened 2018-09-11
TN-1503 fix localized q disappearing issue on UI

achen2401:
https://jira.movember.com/browse/TN-1504
address clinical localized question disappearing issue during initial queries
-  remove the culprit statement that caused this - think it was a debugging statement that was left behind from testing
- make sure save clinical information loading indicator does not display initially (was showing before answer is saved)


achen2401:
wrong story referenced - will re-open with the correct one

PR #2664, opened 2018-09-25
TN-1463 add site ids to report

pbugni:


pbugni:
@mcjustin that's a config value - just add additional identifier systems (once we know what they are) to the list, and they'll be included.

portal/models/fhir.py
```diff
@@ -422,10 +422,16 @@ def aggregate_responses(instrument_ids, current_user, patch_dstu2=False):
...
 
         if subject.organizations:
-            questionnaire_response.document["subject"]["careProvider"] = [
-                Reference.organization(org.id).as_fhir()
-                for org in subject.organizations
-            ]
+            providers = []
+            for org in subject.organizations:
+                org_ref = Reference.organization(org.id).as_fhir()
+                identifiers = [i.as_fhir() for i in org.identifiers if
```
ivan-c: This might work better as a dict, outside the for loop, so the identifiers for an org don't need to be looked up again if an org has already been seen

PR #2668, opened 2018-09-25
TN-1482: /healthcheck

CodeRhymesLife:
Create an API that checks the health of our services dependencies. Namely:
- Celery
- Celery beat
- Postgresql
- Redis

To test:
1) Start the server: ```flask run```
2) Start celery: ```celery worker -A portal.celery_worker.celery --loglevel=info```
3) Start celery beat ```celery beat -A portal.celery_worker.celery --loglevel=info```
4) GET /healthcheck ```curl "http://localhost:5000/healthcheck"```


portal/views/healthcheck.py
```diff
@@ -0,0 +1,138 @@
...
+# Healthcheck functions below
+##############################
+
+def celery_available():
+    """Determines whether celery is available"""
+    # Ping celery. If we get a response
+    # then celery is available. Otherwise
+    # it is not available
+    FNULL = open(os.devnull, 'w')
+    code = subprocess.call([
```
drryanjames: I have a feeling this won't work in production since Celery isn't required to run on the same machine. I'm not sure how else to check whether celery is running other than creating an API on the celery process that we ping from our APIs. But that method seems like overkill, and even unneccessary since we have the celery_beat_available() check that would fail is celery isn't running.

@pbugni @ivan-c do you guys have any thoughts on this? I'm planning on removing this check unless you think this will work or if there is a good alternate method available.

tests/test_healthcheck.py
```diff
@@ -0,0 +1,91 @@
...
+from portal.views.healthcheck import (
+    celery_available,
+    celery_beat_available,
+    get_celery_beat_threshold,
+    postgresql_available,
+    redis_available,
+)
+
+
+class TestHealthcheck(TestCase):
```
drryanjames: I generally don't like writing unit tests that test implementation, but I decided to go this route so I wouldn't have to start and stop processes like redis, celery, and postgresql. Please share thoughts if you have them. Do you think there's a better way to test this? My main goal was to make sure the functions I wrote behave as expected given expected values from dependent services.

portal/config/config.py
```diff
@@ -84,6 +84,8 @@ class BaseConfig(object):
     FLUSH_CACHE_ON_SYNC = True
 
     CELERY_IMPORTS = ('portal.tasks',)
+    CELERY_BEAT_PING_INTERVAL_SECONDS = 60 * 5
```
drryanjames: Any thoughts on what these values should be?
CELERY_BEAT_PING_INTERVAL_SECONDS - how often celery beat pings our APIs
CELERY_BEAT_MISSED_PINGS_BEFORE_FAIL - the number of times we can miss a ping without failing

portal/views/healthcheck.py
pbugni: I'd suggest you simply make a request to the celery test endpoint to confirm a simply task (addition) can be executed by celery:  `/celery-test&redirect-to-result=True`

drryanjames: awesome! Will do.

portal/config/config.py
drryanjames: @pbugni I should have given more background. I expect the ```/healthcheck``` request to originate from our monitoring service too. These variables don't control how often ```/healthcheck`` is called. They control how often we monitor celery beat.

There might be a better way to do this. If so, please let me know. Here's how I did it: I created a celery beat task that periodically (i.e. ```CELERY_BEAT_PING_INTERVAL```) calls ```/celery_beat_ping```. If that API isn't called within in a certain amount of time (i.e. ```get_celery_beat_threshold()```) we will assume celery beat is not running.

ivan-c: Can you remove the `_SECONDS` suffix, and add that information as a comment?

portal/tasks.py
```diff
@@ -375,3 +376,11 @@ def token_watchdog(**kwargs):
     if error_emails:
         return '\nUnable to reach recipient(s): {}'.format(
             ', '.join(error_emails))
+
+
+@celery.task
+def celery_beat_health_check():
+    """Pings the celery beat helath check API for monitoring"""
```
ivan-c: >helath

typo ;)

portal/tasks.py
```diff
@@ -375,3 +376,11 @@ def token_watchdog(**kwargs):
...
+    return requests.get(
```
ivan-c: I might persist this through redis, worried how global variables behave in a multi-threaded environment. eg:
```
redis.from_url(app.config['REDIS_URL']).setex(
    'celerybeat_ping',
    'ping', 
    current_app.config['CELERY_BEAT_PING_INTERVAL'],
)
```

portal/celery_worker.py
```diff
@@ -49,3 +49,11 @@ def setup_periodic_tasks(sender, **kwargs):
                                          task.s(*args_in, **kwargs_in))
             except Exception as exc:
                 logger.error(exc)
+
+    # add the healthcheck task
+    logger.info("Scheduling celery beat health check")
+    sender.add_periodic_task(
```
ivan-c: @pbugni Should this be persisted in ScheduledJobs.json like the others?

portal/config/config.py
pbugni: @drryanjames I'm with you - no other obvious mech to confirm the *beat* service is functional.  This seems like a good solution.

portal/celery_worker.py
pbugni: Yes, for consistency, I think it should.
@drryanjames ask if that doesn't make sense - you'll note there's a `ScheduledJobs.json` file for each application branch.  Bring into the system via `flask sync`.

drryanjames: @pbugni @ivan-c will do. During my first pass I actually used ScheduledJobs.json but thought cron tabs were only executed once a day. I should have read the docs more carefully. Will update.

portal/tasks.py
drryanjames: https://stackoverflow.com/questions/32815451/are-global-variables-thread-safe-in-flask-how-do-i-share-data-between-requests

portal/views/healthcheck.py
```diff
@@ -0,0 +1,104 @@
...
+def celery_beat_ping():
+    """Periodically called by a celery beat task
+
+    Updates the last time we recieved a call to this API.
+    This allows us to monitor whether celery beat tasks are running
+    """
+    rs = redis.StrictRedis.from_url(current_app.config['REDIS_URL'])
+    rs.setex(
+        'last_celery_beat_ping',
+        current_app.config['LAST_CELERY_BEAT_PING_EXPIRATION_TIME'],
```
ivan-c: Can you use keyword arguments for these, so it's obvious what parameter is which?

pbugni:
> @pbugni, I don't think we're missing a file. HealthCheck is imported in requirements.txt

Sorry - clearly I skipped a step...

portal/views/healthcheck.py
ivan-c: for reference:
```
setex(self, name, value, time) method of redis.client.Redis instance
    Set the value of key ``name`` to ``value`` that expires in ``time``
    seconds. ``time`` can be represented by an integer or a Python
    timedelta object.
```

pbugni:
Nicely done @CodeRhymesLife. It's worth noting there's up to a 5 min delay between first deploy and the celery-beat completing its first scheduled check, but I think that's a very reasonable compromise.

Once this is merged in, we should work w/ @jhowe-uw on integration w/ monitoring apps.  I'll note the 500 status returned if any one of our dependencies is offline, should make this easy.

CodeRhymesLife:
@pbugni thanks! And I think the 5 min delay is ok too since it won't cause /healthcheck to fail. The beat check will only fail if beat doesn't ping within 15 mins (5 * 3, where both numbers are configurable).

PR #2669, opened 2018-09-26
Feature/exclusion import

pbugni:
@ivan-c need these in place on titanium to test the stg->test migration.

### DIFF
--- a/docker/docker-compose.yaml
+++ b/docker/docker-compose.yaml
@@ -29,6 +29,10 @@ services:
       - redis
       - celeryworker
       - celerybeat
+    volumes:
+      - source: tmp
+        target: /tmp
+        type: volume
 
   celeryworker:
     <<: *service_base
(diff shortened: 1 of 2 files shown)


docker/docker-compose.yaml
```diff
@@ -29,6 +29,10 @@ services:
       - redis
       - celeryworker
       - celerybeat
+    volumes:
+      - source: tmp
+        target: /tmp
```
ivan-c: Can you pick a more specific place, maybe in /var? /tmp might get filled with system files unrelated to the portal

docker/docker-compose.yaml
```diff
@@ -73,3 +77,4 @@ services:
 
 volumes:
     postgres-data: {}
+    tmp: {}
```
ivan-c: Can you pick something more appropriate and easier to grep for? Maybe something with  "persistence" or "export" or the like?

PR #2678, opened 2018-09-27
TN-1251 writable tmp dir for exclusions via docker

pbugni:
Not certain this is perfect, or where to best set variables such as PERSISTENCE_EXCLUSIONS_DIR.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -51,7 +51,12 @@ ENV \
     PORT="${PORT:-8008}" \
     FLASK_APP="${PROJECT_DIR}/bin/manage.py" \
     PATH="${PROJECT_DIR}/bin:${PATH}" \
-    PERSISTENCE_DIR="${PERSISTENCE_DIR:-gil}"
+    PERSISTENCE_DIR="${PERSISTENCE_DIR:-gil}" \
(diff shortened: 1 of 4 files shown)


docker/Dockerfile
```diff
@@ -51,7 +51,12 @@ ENV \
     PORT="${PORT:-8008}" \
     FLASK_APP="${PROJECT_DIR}/bin/manage.py" \
     PATH="${PROJECT_DIR}/bin:${PATH}" \
-    PERSISTENCE_DIR="${PERSISTENCE_DIR:-gil}"
+    PERSISTENCE_DIR="${PERSISTENCE_DIR:-gil}" \
+    PERSISTENCE_EXCLUSIONS_DIR="${PERSISTENCE_EXCLUSIONS_DIR:-/var/tmp/exclusions}"
```
ivan-c: I might move this to an earlier `ENV` block, so the `mkdir` and `chown` commands can be combined into the ones for `"${PROJECT_DIR}/var/portal-instance/"`. 

This is to reduce the number of layers in a docker image, [but apparently that may only apply to older versions of docker...](https://docs.docker.com/develop/develop-images/dockerfile_best-practices#minimize-the-number-of-layers)

PR #2685, opened 2018-10-01
TN-1453 Return valid FHIR Bundle results from several APIs

pbugni:
Fix for TN-1453, which evolved into a site wide clean-up of our FHIR Bundle management.
Now using a common `models.fhir.bundle_results()` function, and either compliant with FHIR DSTU2 Bundle spec by default, or taking the `patch_dstu2=True` query string.

Fair amount of refactoring and moving code around as `models.fhir` was a bit of a catch-all, making circular imports a problem.


portal/models/clinical_constants.py
```diff
@@ -0,0 +1,133 @@
+""" TrueNTH Clinical Codes """
+import requests
+
+from ..database import db
```
ivan-c: Import order out of whack (`..`/`.`), I'll re-run `isort` in a subsequent PR

portal/models/observation.py
```diff
@@ -0,0 +1,112 @@
...
+            "{status}".format(self, at=at, status=status))
+
+    def as_fhir(self):
+        """Return self in JSON FHIR formatted string"""
+        fhir = {"resourceType": "Observation"}
+        if self.issued:
+            fhir['issued'] = as_fhir(self.issued)
+        if self.status:
+            fhir['status'] = self.status
+        fhir['id'] = self.id
```
ivan-c: Slightly easier to read if these were folded into the initial `fhir` dict initialization, before any of the conditional code

portal/static/js/profile.js
```diff
@@ -2146,7 +2146,7 @@
                 var self = this;
                 self.modules.tnthAjax.getTreatment(self.subjectId, {useWorker:true}, function(data) {
                     self.updateTreatment(data);
-                    self.modules.tnthAjax.getClinical(self.subjectId, {useWorker:true}, function(data) {
+                    self.modules.tnthAjax.getClinical(self.subjectId,{data:{patch_dstu2: true}}, function(data) {
```
ivan-c: `useWorker` option is missing, not sure if intentional...

portal/views/patient.py
```diff
@@ -24,20 +26,27 @@
...
 
-    Example search:
+    Example searches:
         /api/patient?email=username@server.com
+        /api/patient?identifier=http://us.truenth.org/identity-codes/external-study-id|123-45-678
 
-    Returns a FHIR patient resource (http://www.hl7.org/fhir/patient.html)
-    formatted in JSON if a match is found, 404 otherwise.
+    Identifier search pattern:
+        ?identifier=<system>|<value>
```
ivan-c: Why not pass `system` and `value` as individual querystring parameters (eg `id_system`, `id_value`)? Seems error-prone (better than as JSON-in-URL though ;)

pbugni: FHIR compliance.  See https://www.hl7.org/fhir/DSTU2/search.html#token

PR #2699, opened 2018-10-03
get default terms of use url - allow default back to this if for some

achen2401:
In initial queries
allow default terms of use url for fallback use if for some reason, terms of use url that comes with with each LR tou asset is missing.

### DIFF
--- a/portal/static/js/initialQueries.js
+++ b/portal/static/js/initialQueries.js
@@ -81,6 +81,13 @@
         var self = this, sections = self.getSections(), tnthAjax = this.__getDependency("tnthAjax");
         var initDataObj = {
             "topTerms": function() {
+                tnthAjax.getTermsUrl(false, function(data) {
+                    if (!data || !data.url) {
+                        return false;
+                    }
+                    $("#termsURL").attr("data-url", data.url); //default terms of use url
+                    $("#termsCheckbox_default .terms-url").attr("href", data.url);
+                });
                 tnthAjax.getTerms(self.userId, "", "", function() {
                     self.setSectionDataLoadedFlag("topTerms", true);
                 });
--- a/portal/static/js/main.js
+++ b/portal/static/js/main.js
@@ -1202,8 +1202,6 @@ var tnthAjax = {
                 if (!data.error) {
                     $(".get-tou-error").html("");
                     if (data.url) {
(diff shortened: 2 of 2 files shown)


ivan-c:
If I understand correctly, this will use the default/stock ToU if the frontend is unable to retrieve the org-specific ToU from Liferay

@mcjustin does this have any legal ramifications?

mcjustin:
Good question @ivan-c  - I think Mo legal wouldn't want that, but @achen2401 can you clarify which of these we're talking about: https://promowiki.movember.com/display/TU/Website+Consent%2C+General+Terms%2C+Terms+of+Use%2C+Privacy+Policy   ?
Do we even have default content for this? In what situations is it used? The TN USA and ePROMs use cases differ in this regard (ePROMs is org-specific, TN USA isn't).
Isn't the default just another LR content item? Not clear to me in what situations a default might be available from LR when an org-specific version isn't.
> 
> 
> If I understand correctly, this will use the default/stock ToU if the frontend is unable to retrieve the org-specific ToU from Liferay
> 
> @mcjustin does this have any legal ramifications?

achen2401:
@mcjustin Justin, I think this refers to the "All" - "General Terms" - consent terms presented at initial queries - in the document link you provided.  See **https://stg-lr7.us.truenth.org/c/portal/truenth/asset?version=2.2&uuid=d57adc33-738c-67bf-187d-a6a9c5af81c5&editorUrl=true**  via call to **api/tou** which get the initial consent terms url as defined in the AppText.json in each respective blueprint(Truenth USA vs Eproms).

mcjustin:
Thanks for clarifying @achen2401 . We do need to adhere to the rules described there. So, no default for ePROMs, and TN USA should default to the patient version. If these aren't available, we need errors, not other defaults.

> 
> 
> @mcjustin Justin, I think this refers to the "All" - "General Terms" - consent terms presented at initial queries - in the document link you provided. See **https://stg-lr7.us.truenth.org/c/portal/truenth/asset?version=2.2&uuid=d57adc33-738c-67bf-187d-a6a9c5af81c5&editorUrl=true** via call to **api/tou** which get the initial consent terms url as defined in the AppText.json in each respective blueprint(Truenth USA vs Eproms).

PR #2708, opened 2018-10-08
TN-1565 Update Swagger Docs and Spec API

ivan-c:
Updated from #2707

Todo:
* Update `crossdomain` decorator [validation](https://github.com/uwcirg/true_nth_usa_portal/blob/develop/portal/models/client.py#L263-L288) to add configurable Swagger UI host (default to https://uwcirg.github.io/swagger-ui/dist)
  * Update `crossdomain` decorator invocations (remove `origin='*'` references)


portal/views/user.py
```diff
@@ -2085,7 +2178,8 @@ def invite(user_id):
     return jsonify(message=message)
 
 
-@user_api.route('/user/<int:user_id>/messages')
+@user_api.route('/user/<int:user_id>/messages', methods=('OPTIONS','POST'))
```
pbugni: Did you intend to alter the method GET -> POST?  I understand the addition of OPTIONS, but this is why there's a failing test...

CodeRhymesLife: I think that was already there, or I did that inadvertently. Nice catch (exclamation). I will fix.

portal/config/config.py
```diff
@@ -192,6 +192,10 @@ class BaseConfig(object):
         ROLE.SERVICE.value,
     ]
 
+    CORS_DOMAIN_WHITELIST = [
+        '',
+    ]
+
```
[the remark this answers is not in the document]
CodeRhymesLife: I must have forgotten to sync my change that added `uwcirg.github.io` there. I think that's what @ivan-c wanted.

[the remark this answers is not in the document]
CodeRhymesLife: Done.

portal/views/assessment_engine.py
```diff
@@ -52,10 +53,11 @@
...
     '/patient/<int:patient_id>/assessment',
     defaults={'instrument_id': None},
-)
+    methods=('OPTIONS', 'GET'))
 @assessment_engine_api.route(
-    '/patient/<int:patient_id>/assessment/<string:instrument_id>'
-)
+    '/patient/<int:patient_id>/assessment/<string:instrument_id>',
+    methods=('OPTIONS', 'GET'))
+@crossdomain()
```
ivan-c: Did you try passing [`automatic_options=True`](https://github.com/uwcirg/true_nth_usa_portal/blob/develop/portal/views/crossdomain.py#L32)? It might obviate the need to add `OPTIONS` to all the `methods` arguments to `route()`.

[Good reference implementation here](https://github.com/corydolphin/flask-cors/blob/master/flask_cors/decorator.py#L106-L119)

CodeRhymesLife: I didn't, but I will. This will make the change a heck of a lot simpler.

portal/views/tou.py
```diff
@@ -11,11 +11,14 @@
...
 from ..models.tou import ToU, tou_types
 from ..models.user import current_user, get_user_or_abort
+from .crossdomain import crossdomain
 
 tou_api = Blueprint('tou_api', __name__, url_prefix='/api')
 
 
 @tou_api.route('/tou')
+@crossdomain()
+@oauth.require_oauth()
```
CodeRhymesLife: Should this require oauth?

ivan-c: Nice catch, probably accidentally added; I'll remove

ivan-c:
>@ivan-c i'm surprised with the number of origin='*' we had - do you think these were added for the front end? it seems any interventions would already be white listed - just slightly concerned we're gonna break something, or was this all added for the swagger auth?

@pbugni 
John added those while he was testing, but Ryan replaced them with vanilla `cross_domain()` decorators that now allow the Swagger UI domain (in addition to the usually whitelisted interventions etc)

PR #2714, opened 2018-10-09
Fail tests on bad assertions

ivan-c:
* Fail tests when [bad assertion warning raised](https://docs.pytest.org/en/latest/warnings.html)
* Fix bad assertion


setup.cfg
```diff
@@ -80,4 +80,7 @@ dev =
 addopts = --color yes --verbose
 console_output_style = classic
 filterwarnings =
+    default
     ignore:^"localhost" is not a valid cookie domain, it must contain a.*:Warning
+    # error on bad assertions
+    error:.*assertion is always true.*
```
CodeRhymesLife: Would you mind explaining what this does?

CodeRhymesLife: And how it's different from existing behavior. Just curious.

pbugni: @ivan-c sorry to see this late, but I *disagree* - we do NOT want to filter out that warning.  the *assertion is always true* issues should be resolved, as they will fail to accurately test what we want.

History: unittest had an assertion that would take a second parameter to *explain* the situation on failure.  Unfortunately some of these snuck through the port to py.test.  A statement such as `assert (False, 'this should have failed')` will *never* fail, as what was intended to be an explanation string will result in a true value for the assertion itself.

ivan-c: Happy to!
Paul fixed the underlying issue in a recent PR, but [this test will always evaluate to `True`](https://github.com/uwcirg/true_nth_usa_portal/commit/1f9dd1bc088e1a9c23996e225f7dbb025e49c6c3#diff-39b4ec60781bc5394513b105eb07a790L211) since it's actually a tuple ([see TravisCI log](https://travis-ci.org/uwcirg/true_nth_usa_portal/jobs/439404039#L1035-L1040)).

py.test allows configuring a warnings filter ([apparently a builtin feature from python](https://docs.python.org/3/library/warnings.html#warning-filter)), and I've used it to elevate that specific warning to an error, so they don't go unnoticed (people rarely open a passing build log ;).

[A little more background](https://dbader.org/blog/catching-bogus-python-asserts), but I think my solution is more elegant ;D

ivan-c: @pbugni you beat me to it! See above explanation, this PR will make a build fail if it contains a bad assertion

PR #2717, opened 2018-10-11
Hotfix/default acceptance date

pbugni:
Fix for TN-1599


pbugni:
@ivan-c merging this into develop so we can test on stg.  the same code should be treated as a hotfix merged into master

PR #2718, opened 2018-10-12
Revert "Hotfix/default acceptance date"

pbugni:
It would appear the altering of user_consents within a query loop generates (in some cases) additional unwanted near duplicate rows of user_consents.

Back out this migration till it's patched.

@ivan-c , @mcjustin FYI
Reverts #2717


ivan-c:
@pbugni I believe this will need special handling on the stg* systems, downgrading to alembic revision [`67c2bea62313`](https://github.com/uwcirg/true_nth_usa_portal/pull/2717/files#diff-be8db2b4c5d60f63e5356d342eef64fcR15) before deploying...

pbugni:
Other PRs since have patched the source of the problem.  Closing.

PR #2720, opened 2018-10-15
Feature/frontend cookie check

achen2401:
address this story:
https://jira.movember.com/browse/TN-1427

still need possible verbiage change from Kelly/Vivek per Diana's comment, but want to get these changes out so they can be tested ahead.  have tested myself in IE, Firefox, Chrome and Safari


portal/templates/flask_user/_macros.html
```diff
@@ -116,3 +116,28 @@
     <a href="{%-if config.MOVEMER_LINK_URL-%}{{config.MOVEMER_LINK_URL}}{%-endif-%}" class="logo-link" title="{{ _('Movember') }}" target="_blank">
     {%-if white_theme -%}<img id="footerLockup" src="{{PORTAL}}{{url_for('static',filename='img/Movember-Footer-Logo.png')}}" alt="{{_('Movember Logo')}}" />{%-else-%}<img src="{{PORTAL}}{{url_for('static',filename='img/movember_logo.jpg')}}" alt="Movember Logo"/>{%-endif-%}</a>
 {%- endmacro %}
+{% macro checkCookieMonster(target="") -%}
```
ivan-c: I think this might be more appropriate in a different directory as it's not specific to flask-user...

portal/views/portal.py
```diff
@@ -280,6 +302,10 @@ def access_via_token(token, next_step=None):
         vocabulary - see `NextStep`
 
     """
+    # Before going any further, confirm cookies are enabled
+    if not request.args.get('cookies_tested'):
+        return redirect(url_for('portal.require_cookies', target=request.url))
```
ivan-c: Is there a way to implement this entirely in the frontend? I'd prefer that we don't add another redirection to the login process.

I understand that `/challenge` [will fail](https://github.com/uwcirg/true_nth_usa_portal/blob/develop/portal/views/portal.py#L437) before any frontend code can be loaded, because [the session cookie is never set, and unable to find the user](https://github.com/uwcirg/true_nth_usa_portal/blob/develop/portal/views/portal.py#L424). 
I think in that case, we should either redirect to an error page about cookies (similar to the `/no-script` page) and/or return a cookie-spefic error message than [what's returned](https://github.com/uwcirg/true_nth_usa_portal/blob/develop/portal/models/user.py#L1830).

@pbugni thoughts?

portal/static/js/CookieMonster.js
```diff
@@ -0,0 +1,167 @@
+(function() {
+    var CookieMonster = window.CookieMonster = function() {
+        this.modalElementId = "modalCookieEnableWarning";
+        this.testCookieName = "testCookieMonster";
+        this.resizeTimer = 0;
+    };
+    CookieMonster.prototype.getSelectorsToDisable = function() {
+        return this.selectorsToDisable || this.defaultSelectorsToDisable;
+    };
+    CookieMonster.prototype.deleteTestCookie = function() {
```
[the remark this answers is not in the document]
achen2401: yea, agree, ivan.  I, however, am hesitant to add another external library that the portal frontend code will depend on, on top of others that were already there.  At least I think using plain JS for this check is sufficient.

portal/views/portal.py
achen2401: hmm, this one is tough.  My understand from my discussion with Paul is that the originating call comes not from /challenge but /access - which makes a pure frontend implementation difficult.  Also, I think the error generated here - stemming from invalid user id - could come from reason other than browser cookie being disabled, e.g. session timed out, token expired, etc...  I however will defer to you and Paul on what the best approach here and will make the change accordingly.

portal/static/js/CookieMonster.js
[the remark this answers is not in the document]
pbugni: @achen2401 's logic works for me - @ivan-c ?

portal/views/portal.py
pbugni: i only added this as i was under the impression that we could *not* implement this entirely in the frontend - @achen2401 didn't you experiment with this and prove (to yourself) that it required server side checks?

ivan-c: >that it required server side checks?

I confirmed this too, roughly:

1. User clicks on email link (`/access`)
2. redirected to `/challenge`
3. [`/challenge` fails](https://github.com/uwcirg/true_nth_usa_portal/blob/develop/portal/views/portal.py#L437) because cookie never set during `/access`

I think a better solution would be to improve the error message, by checking explicitly checking for the session cookie

achen2401: think the request is aborted here:  https://github.com/uwcirg/true_nth_usa_portal/blob/c38fda20478fb75682552973941bcc35dab245b0/portal/views/portal.py#L435
so the response won't reach the frontend thereafter - I guess server side change is needed  there regardless for a pure frontend implementation.  I think that request could be aborted **for a number of reasons**, etc. no session cookie set, or bad token (token expired), which is why I thought it a good idea for redirecting for a frontend cookie check prior, so at least that will eliminate browser disabling of cookies as one of the reasons.  that is my thinking but am open to other implementations.

ivan-c: >I think that request could be aborted for a number of reasons

Fair enough, what about redirecting only if there's a potential issue (ie `user_id is None` or `'session' not in request.cookies`), not every time?

>think the request is aborted here:

I think it's [actually here](https://github.com/uwcirg/true_nth_usa_portal/blob/c38fda20478fb75682552973941bcc35dab245b0/portal/views/portal.py#L437), the `POST` is for handling the validation of the challenge data AFAIK

>token expired

I think the error is `404 - Access token has expired` in that case

portal/views/portal.py
```diff
@@ -442,15 +439,28 @@ def challenge_identity(
...
+        necessary in no cookie situations, to redirect user back
 
     """
+    # At this point, we can expect a session, or the user likely
+    # doesn't have cookies enabled.  (ignore misleading `_fresh`
+    # and `_permanent` keys)
+    session_keys = [k for k in session if k not in ('_fresh', '_permanent')]
+    if not session_keys:
+        request_path = request.args.get('request_path', request_path)
+        return redirect(url_for(
```
ivan-c: @pbugni Should we have some kind of logging when this happens (info/debug)?

tests/test_access_url.py
```diff
@@ -39,14 +39,19 @@ def test_use_access_url(self):
         onetime = db.session.merge(onetime)
 
         token = user_manager.token_manager.generate_token(onetime.id)
-        access_url = url_for('portal.access_via_token', token=token)
+        access_url = url_for(
+            'portal.access_via_token', token=token, cookies_tested=True,
```
ivan-c: @pbugni Did you need to add `cookies_tested=True` to work around these tests because they do not use cookies?...

I don't that QS parameter will prevent the redirection, it's added after redirecting through `/require_cookies` but isn't checked otherwise.

portal/views/portal.py
pbugni: sure.  good idea

tests/test_access_url.py
pbugni: no - just testing that the redirection to challenge happens.  but thanks for noticing - that param is no longer needed, removing from the tests.

PR #2739, opened 2018-10-22
TN-1619 - send back to `next_after_login` if lacking required state

pbugni:


mcjustin:
not following how code could have been simplified this much, but happy to trust your change

PR #2746, opened 2018-10-24
Always return 200 from /healthcheck

CodeRhymesLife:
These changes return a 200 status code when a successful JSON response is returned from the healthcheck API, despite if one of our dependencies is down.

Healthcheck by default returns a 500 when one of the dependent services is down. The 500 status code should be reserved for cases when "the server encountered an unexpected condition that prevented it from fulfilling the request" ([citation](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/500)). While one of our dependencies going down is not good, it does not mean the /healthcheck API was unable to fulfill the request. As long as we successfully process the request, despite the health of our dependencies, we should return a 200.

I considered submitting a pull request to Healthcheck since I think the default behavior is incorrect, but I decided not to as a change in this behavior would probably break a lot of users.

### DIFF
--- a/portal/factories/app.py
+++ b/portal/factories/app.py
@@ -45,7 +45,11 @@ from ..views.extend_flask_user import (
 from ..views.fhir import fhir_api
 from ..views.filters import filters_blueprint
 from ..views.group import group_api
-from ..views.healthcheck import HEALTH_CHECKS, healthcheck_blueprint
+from ..views.healthcheck import (
+    HEALTH_CHECKS,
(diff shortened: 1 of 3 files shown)


CodeRhymesLife:
@ivan-c we have @jhowe-uw to thank for this one!

ivan-c:
@pbugni was leaning to `develop` but it doesn't look like we'll bring any changes from `develop` into the RC if I change the base branch in the PR (Ryan's topic branch was started before the RC)

PR #2748, opened 2018-10-24
Add ids to celery_beat_health_check scheduled jobs

CodeRhymesLife:
This should make sure last_runtime is update.

Explanation:
To resolve this I think we need to give the celery_beat_health_check an id in scheduledjob.json. It looks like last_runtime [only gets updated when the job has a job_id](https://github.com/uwcirg/true_nth_usa_portal/blob/99b10b5891af0b480426a1fec9db6b311dbf7bd7/portal/tasks.py#L80). job_id is set from the id property in scheduledjob.json [here](https://github.com/uwcirg/true_nth_usa_portal/blob/develop/portal/models/scheduled_job.py#L85). I didn't assign an id for the celery_beat_health_check  in either the [eproms scheduledjob.json](https://github.com/uwcirg/true_nth_usa_portal/blob/develop/portal/config/eproms/ScheduledJob.json#L76) or in the [gil version](https://github.com/uwcirg/true_nth_usa_portal/blob/develop/portal/config/gil/ScheduledJob.json#L39). Giving them both an id should fix this issue.

### DIFF
--- a/portal/config/eproms/ScheduledJob.json
+++ b/portal/config/eproms/ScheduledJob.json
@@ -69,6 +69,7 @@
     },
     {
       "active": true,
+      "id": 8,
       "kwargs": null,
       "name": "Celery Beat Health Check",
       "resourceType": "ScheduledJob",
--- a/portal/config/eproms/ScheduledJob.json
+++ b/portal/config/eproms/ScheduledJob.json
@@ -69,7 +69,6 @@
     },
     {
       "active": true,
-      "id": 8,
       "kwargs": null,
       "name": "Celery Beat Health Check",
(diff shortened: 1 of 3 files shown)


ivan-c:
Thanks for taking a closer look! I don't see why we can't pass the name of the job, but I'll defer to @pbugni . 
I'm going to change the branch you branched from (and force-push) so we can merge this into our current RC. (You may need to run `git reset --hard origin/fix/celery_beat_health_check_id` after checking out this branch)

CodeRhymesLife:
@pbugni I haven't tested this PR yet to see if it works. Need to head out but will loop back this afternoon

PR #2762, opened 2018-10-29
TN-1569 WIP: DON'T MERGE YET

pbugni:
Work in progress for questionnaire bank timeline.

Getting close!  Now includes a view to see any patient's timeline at:

`/api/patient/<patient_id>/timeline`


portal/models/qb_timeline.py
```diff
@@ -0,0 +1,290 @@
...
+    @property
+    def status(self):
+        return self._status
+
+    @status.setter
+    def status(self, value):
+        self._status = getattr(OverallStatus, value).name
+
+
+system_trigger = datetime(2015, 1, 1, 12, 0, 0)
```
ivan-c: Maybe something further back and standardized, like the Unix Epoch?

portal/models/questionnaire_bank.py
```diff
@@ -638,6 +685,23 @@ def add_if_not_found(self, commit_immediately=False):
         return self
 
 
+@dogpile_cache.region('qb_query_cache')
```
ivan-c: Are you caching DB objects? I know we've had issues with stale DB objects with celery

portal/models/qb_timeline.py
```diff
@@ -0,0 +1,290 @@
...
+
+from ..database import db
+from ..date_tools import RelativeDelta
+from .assessment_status import OverallStatus
+from .questionnaire_bank import QuestionnaireBank, qbs_by_rp
+from .questionnaire_response import QNR_results
+from .user_consent import consent_withdrawal_dates
+
+
+class QBT(db.Model):
```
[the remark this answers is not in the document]
pbugni: dogpile doesn't have the same granular controls we get from SQL.  plus, we treat redis as a volatile cache - this really will alleviate all the work to build out users data other than when submissions come in, and then just the one user at a time.

portal/models/qb_timeline.py
pbugni: we ran into trouble with dates < 1900 w/ one of the python date libraries.  can go to that if you'd rather?  (jan 1, 1900)

portal/models/questionnaire_bank.py
pbugni: yes - and i believe i found the workaround - not the updated comment and all clients refresh the cached objects in the session.

pbugni: I am finding the *win* by caching these is minimal (like 15% over the entire run) - tempted to remove this caching as it does add complications.

portal/models/qb_timeline.py
pbugni: killed this approach after finding it dysfunctional w/ the "month math" we require.

portal/models/questionnaire_bank.py
pbugni: I *did* remove this caching - premature optimization, more trouble to get right than the minor savings, which we only see on the *first* run.

PR #2764, opened 2018-10-30
Ensure scheduled jobs have the appropriate scheduled_task decorator

CodeRhymesLife:
Untested. Having trouble starting celery beat. It just hangs. I've had this problem before; I don't think it's due to my changes.

### DIFF
--- a/portal/celery_worker.py
+++ b/portal/celery_worker.py
@@ -41,7 +41,8 @@ def setup_periodic_tasks(sender, **kwargs):
         if task:
             if not task.func_dict.get('scheduled_task', False):
                 logger.error("Scheduled job {} does not have "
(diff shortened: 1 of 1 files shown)


portal/celery_worker.py
```diff
@@ -39,6 +39,10 @@ def setup_periodic_tasks(sender, **kwargs):
     for job in ScheduledJob.query.all():
         task = getattr(tasks, job.task, None)
         if task:
+            if not task.func_dict.get('scheduled_task', False):
+                logger.error("Scheduled job {} does not have "
```
ivan-c: I think `logging` has built-in functionality for [sprintf-style string formatting](https://docs.python.org/2/library/logging.html#logging.Logger.debug); slightly more readable

PR #2766, opened 2018-10-30
TN-1642 Delay celery beat healthcheck warnings/errors

CodeRhymesLife:
/healthcheck returns false until celery beat starts, which doesn't happen until after the web service starts. These changes ping the celery beat health check endpoint while celery beat is starting up to "trick" /healthcheck into thinking celery beat is running. In effect, this delays warnings a few minuets (based on LAST_CELERY_BEAT_PING_EXPIRATION_TIME).

### DIFF
--- a/docker/docker-compose.yaml
+++ b/docker/docker-compose.yaml
@@ -55,7 +55,7 @@ services:
     command: bash -c '
       env &&
       wait-for-it --timeout=120 --host=web --port="$$PORT" --strict &&
-      flask initialize_celery_beat_healthcheck &&
+      flask set-celery-beat-healthy &&
       celery beat
         --app portal.celery_worker.celery
         --loglevel debug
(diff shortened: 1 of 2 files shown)


manage.py
```diff
@@ -306,3 +307,8 @@ def config(config_key):
         {k: v for k, v in app.config.items() if isinstance(v, basestring)},
         indent=2,
     ))
+
+
+@app.cli.command()
+def initialize_celery_beat_healthcheck():
```
ivan-c: Maybe something more obvious, like `celery_beat_healthcheck` or `set_celerybeat_healthy`? Could be used outside of initialization/deployment

CodeRhymesLife: I liked ```set_celerybeat_healthy```

docker/docker-compose.yaml
```diff
@@ -55,6 +55,7 @@ services:
     command: bash -c '
       env &&
       wait-for-it --timeout=120 --host=web --port="$$PORT" --strict &&
+      flask initialize_celery_beat_healthcheck &&
```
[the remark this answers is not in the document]
CodeRhymesLife: Nice suggestion. I like this feature.

PR #2770, opened 2018-10-31
Expose celery config as env var

ivan-c:


portal/config/config.py
```diff
@@ -86,6 +86,8 @@ class BaseConfig(object):
     FLUSH_CACHE_ON_SYNC = True
 
     CELERY_IMPORTS = ('portal.tasks',)
+    CELERYD_MAX_TASKS_PER_CHILD = os.environ.get('CELERYD_MAX_TASKS_PER_CHILD')
```
pbugni: Can we add a "sane" default if the env var isn't defined?  Dunno what that should be, something in the 1-4 range?

ivan-c: I want to experiment with some values before deciding on what's sane, will add a default in a subsequent PR

PR #2771, opened 2018-10-31
Update celery health check and standardize responses

CodeRhymesLife:
PROBLEMS:
1) The celery health check returns true when celery failed to process jobs
2) Responses were not standardized which makes monitor code cumbersome

AFTER THESE CHANGES:
1) The celery health checks the result of a job using the celery_test API
2) Responses are standardized to make monitoring easier

### DIFF
--- a/manage.py
+++ b/manage.py
@@ -10,11 +10,13 @@ import alembic.config
 import click
 from flask import url_for
 from flask_migrate import Migrate
+import json
 from past.builtins import basestring
 import redis
 import requests
 from sqlalchemy import func
 from sqlalchemy.orm.exc import NoResultFound
+import sys
 
 from portal.audit import auditable_event
 from portal.config.site_persistence import SitePersistence
@@ -322,9 +324,8 @@ def healthcheck():
     result = requests.get(
         url_for('check')
     )
-    print(result.text)
+    print(json.dumps(result.json(), indent=4))
 
-    if not result.ok:
-        return result.status_code
-
-    return 0
+    # Return success (0) if passing status code
+    # otherwise return fail (1)
+    return sys.exit(int(not result.ok))
--- a/manage.py
+++ b/manage.py
@@ -323,3 +323,8 @@ def healthcheck():
         url_for('check')
     )
     print(result.text)
+
+    if not result.ok:
+        return result.status_code
+
+    return 0
--- a/manage.py
+++ b/manage.py
@@ -8,9 +8,11 @@ import os
 
 import alembic.config
 import click
+from flask import url_for
 from flask_migrate import Migrate
 from past.builtins import basestring
 import redis
+import requests
 from sqlalchemy import func
 from sqlalchemy.orm.exc import NoResultFound
 
@@ -306,3 +308,12 @@ def config(config_key):
         {k: v for k, v in app.config.items() if isinstance(v, basestring)},
         indent=2,
     ))
+
+
+@app.cli.command()
+def healthcheck():
+    """Calls the healthcheck API"""
+    result = requests.get(
+        url_for('check')
+    )
+    print(result.text)
--- a/manage.py
+++ b/manage.py
@@ -327,5 +327,8 @@ def healthcheck():
     print(json.dumps(result.json(), indent=4))
 
     # Return success (0) if passing status code
-    # otherwise return fail (1)
-    return sys.exit(int(not result.ok))
+    if result.ok:
+        return sys.exit()
+
+    # Healthcheck failed. Return a failing status code
+    return sys.exit(result.status_code)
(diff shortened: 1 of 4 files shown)


portal/views/portal.py
```diff
@@ -1051,7 +1051,7 @@ def celery_test(x=16, y=16):
     context = {"id": res.task_id, "x": x, "y": y}
     result = "add((x){}, (y){})".format(context['x'], context['y'])
     task_id = "{}".format(context['id'])
-    result_url = url_for('.celery_result', task_id=task_id)
+    result_url = url_for('portal.celery_result', task_id=task_id)
```
CodeRhymesLife: Needed to explicitly reference the full path since celery_test is called from a different blueprint.

manage.py
```diff
@@ -312,3 +314,12 @@ def config(config_key):
...
     return celery_beat_health_check()
+
+
+@app.cli.command()
+def healthcheck():
+    """Calls the healthcheck API"""
+    result = requests.get(
+        url_for('check')
+    )
+    print(result.text)
```
ivan-c: Nice addition! Can you return a non-`0` exit code if the response isn't OK?

portal/views/healthcheck.py
```diff
@@ -55,14 +66,9 @@ def celery_beat_available():
...
-        return (
-            True,
-            'Celery beat is available. Last check: {}'.format(
-                last_celery_beat_ping
-            )
-        )
+        return True, 'Celery beat is available.'
 
-    return False, 'Celery beat is not running jobs'
+    return False, 'Celery beat is not aviailable.'
```
ivan-c:
```suggestion
    return False, 'Celery beat is not available.'
```

manage.py
CodeRhymesLife: Done! Let me know if this is what you were looking for. If not, I'll update.

portal/views/healthcheck.py
CodeRhymesLife: Thanks!

manage.py
```diff
@@ -312,3 +314,17 @@ def config(config_key):
...
+
+    if not result.ok:
+        return result.status_code
```
ivan-c: Did this work? I thought you had to the below, but not sure if flask adds some magic:
```suggestion
        sys.exit(result.status_code)
```

CodeRhymesLife: I simplified this a little bit. Let me know if what I did works.

manage.py
```diff
@@ -312,3 +316,16 @@ def config(config_key):
...
+
+@app.cli.command()
+def healthcheck():
+    """Calls the healthcheck API"""
+    result = requests.get(
+        url_for('check')
+    )
+    print(json.dumps(result.json(), indent=4))
+
+    # Return success (0) if passing status code
```
ivan-c: I'd prefer the HTTP code as the exit code (or `0` for 200/OK/success) if it's easy enough.

manage.py
```diff
@@ -312,3 +316,16 @@ def config(config_key):
...
+    # otherwise return fail (1)
+    return sys.exit(int(not result.ok))
```
ivan-c: I don't think the exit code is working correctly:
```
$ flask healthcheck
{
    "status": "failure",
    "timestamp": 1541090957.455822,
    "hostname": "zinc",
    "results": [
        {
            "output": "Celery is not available.",
            "checker": "celery_available",
            "expires": 1541090966.440825,
            "passed": false,
            "timestamp": 1541090957.440825
        },
        {
            "output": "Celery beat is not available.",
            "checker": "celery_beat_available",
            "expires": 1541090966.442809,
            "passed": false,
            "timestamp": 1541090957.442809
        },
        {
            "output": "PostgreSQL is available.",
            "checker": "postgresql_available",
            "expires": 1541090984.454893,
            "passed": true,
            "timestamp": 1541090957.454893
        },
        {
            "output": "Redis is available.",
            "checker": "redis_available",
            "expires": 1541090984.4558,
            "passed": true,
            "timestamp": 1541090957.4558
        }
    ]
}
$ echo $?
0
```

ivan-c: @CodeRhymesLife Sorry, on second thought this seems like the same `500`/`200` debate we had. I think it should only return non-`0` if there was an issue performing the check, but still `0` if unhealthy

CodeRhymesLife:
@jhowe-uw please see thread on Slack for more info, but we decided to keep the celery beat check how it is. The celery worker and celery beat currently rely on redis for scheduling, so if redis goes down we're hosed anyways. For that reason we think it's ok to have the beat healthcheck rely on redis.

manage.py
CodeRhymesLife: just added this in.

PR #2779, opened 2018-11-02
TN-1648-USA-GIL-hotfix (minor edit)

mcjustin:


mcjustin:
need another change first

PR #2780, opened 2018-11-02
Tn 1648 usa gil hotfix

mcjustin:


mcjustin:
@ivan-c  will need this on USA prod and demo today, and merged into develop.
https://jira.movember.com/browse/TN-1648

PR #2784, opened 2018-11-06
TN-1234 followup fix - fix call for config

achen2401:
Saw this while testing - was seeing unnecessary ajax calls for configuration being sent off in system outage message rendering function in authenticated pages
- Fix is to add check for the presence of system outage element before sending off ajax call to get configuration variables


achen2401:
@mcjustin Justin, that is inline JS though..I was trying to avoid that.

PR #2787, opened 2018-11-07
fix button not displayed issue for self-registered music patient

achen2401:
stumbled on this edge case when testing:  When self-registering as a patient affiliated with one of the Michigan clinics (MUSIC clinics), the next button does not display if **there is** still needed required core data.
Fix is to allow patient to continue via window reload as next button is not displayed by default in continue to next work flow.
https://jira.movember.com/browse/TN-1654 @mcjustin Justin, FYI

### DIFF
--- a/portal/static/js/initialQueries.js
+++ b/portal/static/js/initialQueries.js
@@ -327,12 +327,12 @@
         }(), 800);
     };
 
+    FieldsChecker.prototype.isAcceptOnNext = function() {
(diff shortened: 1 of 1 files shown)


mcjustin:
@achen2401 I'm not entirely clear what you're describing here... please attach a screenshot next time (a picture is worth a thousand words :) )

achen2401:
Justin, sure, I will create a story for this.

PR #2789, opened 2018-11-08
TN-1655: Integration test foundation

CodeRhymesLife:
Much of the foundation was already laid by Ivan. Working off of his code I made the following changes to make it easy to add new integration tests: 

1) Created a clear division between unit and integration tests with the following folder structure
&nbsp;&nbsp;tests/
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;/integration_tests/
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;/unit_tests/
2) Update a few references since things got moved around
3) Updated tox.ini to reflect the above changes

In order for integration tests to run the following items need to be installed:
- xvfb ```sudo apt-get install xvfb```
- geckodriver ```wget -qO- https://github.com/mozilla/geckodriver/releases/download/v0.23.0/geckodriver-v0.23.0-linux64.tar.gz | sudo tar xvz -C /usr/local/bin```
- Firefox ```sudo apt-get install iceweasel``` <- on debian

If this all seems appropriate, can someone help me automate the above installation on dev deployments and on travis?


tests/integration_tests/test_login.py
```diff
@@ -0,0 +1,22 @@
+from .pages import LoginPage  # noqa isort:skip
```
ivan-c: No longer necessary since the imports are no longer in the middle of the file
```suggestion
from .pages import LoginPage
```

tests/integration_tests/test_login.py
```diff
@@ -0,0 +1,22 @@
...
+from tests.integration_tests import IntegrationTestCase  # noqa isort:skip
```
ivan-c:
```suggestion
from tests.integration_tests import IntegrationTestCase
```

tests/integration_tests/test_login.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+class TestLogin(IntegrationTestCase):
```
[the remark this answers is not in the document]
CodeRhymesLife: Yes, but I am planning on modifying it tomorrow when I check in an example test.

tox.ini
```diff
@@ -8,7 +8,7 @@ deps =
     -rrequirements.txt
     pytest-cov
 passenv = TRAVIS* PG* SQLALCHEMY_DATABASE_TEST_URI PERSISTENCE_DIR FLASK_APP SECRET_KEY
-commands = py.test --cov portal --cov-report xml:"{toxinidir}/coverage.xml" []
+commands = py.test --ignore=tests/integration_tests --cov portal --cov-report xml:"{toxinidir}/coverage.xml" []
```
ivan-c: I think `--ignore=tests/integration_tests` will be required for every invocation of `py.test` , if you don't want the selenium/integration tests to run.

@pbugni 
Should the integration tests run when `py.test` is called?

CodeRhymesLife:
I'm having trouble getting a simple integration test to pass. The test enters a test user's credentials and clicks the login button. The server then crashes due to a Broken Pipe Error (below). This could be due to a bug in geckodriver/selenium according to this [stackoverflow post](https://stackoverflow.com/questions/51239512/broken-pipe-error-selenium-webdriver-when-there-is-a-gap-between-commands). But when I followed the post's instructions and downgraded geckodriver I got the same failure. I will keep debugging. 

```-------------------------------------------------------------------------------------------------------------------------------------------------- Captured stdout call -------------------------------------------------------------------------------------------------------------------------------------------------- * Serving Flask app "portal" (lazy loading)
 * Environment: production
   WARNING: Do not use the development server in a production environment.
   Use a production WSGI server instead.
 * Debug mode: off
----------------------------------------
Exception happened during processing of request from ('127.0.0.1', 37502)
----------------------------------------
-------------------------------------------------------------------------------------------------------------------------------------------------- Captured stderr call --------------------------------------------------------------------------------------------------------------------------------------------------[2018-11-12 07:09:40,972] AUDIT in audit: performed by 1 on 1: login: local user login
[2018-11-12 07:09:40,983] DEBUG in user: resetting lockout variables - failures: 0, last_failure: None
[2018-11-12 07:09:41,050] DEBUG in coredata: intial NOT obtained for at least RoleData
[2018-11-12 07:09:41,050] DEBUG in auth: next_after_login: [need data] -> initial_queries
  File "/usr/lib/python2.7/SocketServer.py", line 596, in process_request_thread
    self.finish_request(request, client_address)
  File "/usr/lib/python2.7/SocketServer.py", line 331, in finish_request
    self.RequestHandlerClass(request, client_address, self)
  File "/usr/lib/python2.7/SocketServer.py", line 654, in __init__
    self.finish()
  File "/usr/lib/python2.7/SocketServer.py", line 713, in finish
    self.wfile.close()
  File "/usr/lib/python2.7/socket.py", line 283, in close
    self.flush()
  File "/usr/lib/python2.7/socket.py", line 307, in flush
    self._sock.sendall(view[write_offset:write_offset+buffer_size])
error: [Errno 32] Broken pipe```

**Comment by ivan-c** [comment]
@CodeRhymesLife If I understand correctly, that's output from your local dev environment. 

If you get a chance, could you push your changes to the PR so we can see if it passes on TravisCI?

(selenium tests can be flaky)

**Comment by CodeRhymesLife** [comment]
The test passed!

File: tests/integration_tests/__init__.py [hunk 5]
```diff
@@ -79,8 +72,38 @@ def setUp(self):
...
-
-        super(TestUI, self).setUp()
+        self.driver.implicitly_wait(30)
+        self.verificationErrors = []
+        self.wait = WebDriverWait(self.driver, 60)
+        self.accept_next_alert = True
+
+        super(IntegrationTestCase, self).setUp()
+
+    def is_element_present(self, how, what):
```
CodeRhymesLife: These functions are auto generated by Selenium IDE and will be used during more complicated tests. Checking them in now do we don't have to later.

tox.ini
pbugni: @ivan-c I'm not sure how to best integrate - this is a CI issue in my mind, which I feel you're more qualified to answer.

Shall we start w/ full integration, and back off if we find it too noisy?

tests/integration_tests/__init__.py
```diff
@@ -79,8 +72,38 @@ def setUp(self):
         self.addCleanup(self.driver.quit)
 
         self.driver.root_uri = self.get_server_url()
-
-        super(TestUI, self).setUp()
+        self.driver.implicitly_wait(30)
+        self.verificationErrors = []
+        self.wait = WebDriverWait(self.driver, 60)
```
ivan-c: I don't think this will do anything unless used by an explicit wait (ie [Expected Condition](https://selenium-python.readthedocs.io/waits.html))

Can you remove it if that's the case?

tests/integration_tests/__init__.py
ivan-c: Can you add some comments to that effect (ie "functions exported from Selenium IDE plugin")?

tests/integration_tests/__init__.py
```diff
@@ -95,20 +118,6 @@ def tearDown(self):
...
-        page.get("/user/sign-in")
-
-        assert "Uh-oh" not in page.w.find_element_by_tag_name("body").text
-
-    def test_login_form_fb_exists(self):
-        """Ensure Facebook button present on login form"""
-
-        page = LoginPage(self.driver)
-        page.get("/user/sign-in")
+        self.assertEqual([], self.verificationErrors)
```
ivan-c: This seems to be exported from Selenium IDE, but I don't see any tests using it. I'm also hesitant about having an assertion outside of a test function.

Can you remove it?

tests/integration_tests/test_login.py
```diff
@@ -0,0 +1,64 @@
...
+
+        driver.find_element_by_name("email").click()
+        driver.find_element_by_name("email").clear()
+        driver.find_element_by_name("email").send_keys(TEST_USERNAME)
+        driver.find_element_by_name("password").click()
+        driver.find_element_by_name("password").clear()
+        driver.find_element_by_name("password").send_keys(DEFAULT_PASSWORD)
+        driver.find_element_by_xpath(
+            "//input[@class='btn btn-tnth-primary btn-lg' and @value='LOG IN']").click()
+        driver.find_element_by_id("tnthUserBtn").click()
```
ivan-c: This is failing when I run the tests locally.
Until I can figure out how to use [`expected_condition`](https://selenium-python.readthedocs.io/waits.html) correctly (checking for the correct element to appear), can you add a `time.sleep(10)` to let the login process complete in the previous step.

tests/integration_tests/test_login.py
```diff
@@ -0,0 +1,64 @@
...
+        driver.find_element_by_link_text("Log Out of TrueNTH").click()
+
+    def test_consent_after_login(self):
```
ivan-c: Can you add this test in a subsequent PR, or set as an [expected failure with pytest](https://docs.pytest.org/en/latest/skipping.html#xfail-mark-test-functions-as-expected-to-fail)?

It seems to have [broken every `ui` job](https://travis-ci.org/uwcirg/true_nth_usa_portal/jobs/456642263) since you added. NB: you'll have to open each TravisCI build to see since the `ui` "job" is allowed to fail in TravisCI

tests/integration_tests/__init__.py
CodeRhymesLife: I don't know how that wait got in there!

tests/integration_tests/test_login.py
CodeRhymesLife: Would that fix it? I added a few implicit waits a few days ago and still couldn't get tests to pass locally. I was running into a socket issue though RE slack.

tests/integration_tests/test_login.py
CodeRhymesLife: I've been looking at them and I've been trying to fix it. I think I know why it's failing. At some point through the consent process a dialogue pops up and I don't think selenium likes it. I think there is a consent workflow that I'm hoping gets this test to pass. I will modify test.

tests/integration_tests/test_login.py
ivan-c: As I understand it, changing the implicit wait time will give selenium more time to check the DOM for a given element before failing. I think the "broken pipe" error is because too many HTTP requests are being made at once, but I'm not 100% sure (related [Stack Overflow](https://stackoverflow.com/questions/37897926/how-to-keep-werkzeug-from-crashing-after-nginx-closes-pipe)).

I was able to get `test_login_page()` to pass locally on my laptop after adding `time.sleep(5)`, when before I kept getting the "broken pipe" error at that step.

tests/integration_tests/test_login.py
ivan-c: Thanks for getting some concrete examples checked in!

You may be having issues with the consent/org selection. As far as I know, only a few test organizations are added to the test database, so you may have to choose the "None of the Above" option or pick a known org.

tests/integration_tests/__init__.py
CodeRhymesLife: @ivan-c I think we should keep this so adding tests is really simple. If we remove this, Warren (and others) will have to worry about setup and teardown which makes things slightly more complicated. Having this will let him simply copy and paste just the test and avoid worrying about any other functions (I think). That was my goal for these changes.

tests/integration_tests/__init__.py
CodeRhymesLife: Actually, this tells selenium to wait at most 30 seconds and try again if it doesn't see something it expects

tests/integration_tests/__init__.py
```diff
@@ -79,8 +72,52 @@ def setUp(self):
         self.addCleanup(self.driver.quit)
 
         self.driver.root_uri = self.get_server_url()
-
-        super(TestUI, self).setUp()
+        self.driver.implicitly_wait(30)
```
ivan-c: :+1:

tests/integration_tests/__init__.py
ivan-c: I think the above line (setting up the implicit wait, `self.driver.implicitly_wait(30)`) does that.

As far as I know, explicit waits must be invoked on individual elements (see below example), and I don't see any references to the `self.wait` property in any of the added test functions.

---
Explicit wait example from docs ([see *Expected Conditions* section](https://selenium-python.readthedocs.io/waits.html#explicit-waits)):
```
from selenium.webdriver.support import expected_conditions as EC

wait = WebDriverWait(driver, 10)
element = wait.until(EC.element_to_be_clickable((By.ID, 'someid')))
```

tests/integration_tests/__init__.py
```diff
@@ -95,20 +132,6 @@ def tearDown(self):
         ):
             self.driver.execute_script("sauce:job-result=passed")
```
ivan-c:
```suggestion

        # Added to support tests exported from Selenium IDE
```

tests/integration_tests/__init__.py
ivan-c: Sounds reasonable. Until some tests start referencing that property, I suggested a comment to make it obvious what it's there for.

PR #2804, opened 2018-11-16
Add STU3 Questionnaire Resources

ivan-c:
* Replace DSTU2 Questionnaire stubs with [STU3 versions](http://hl7.org/implement/standards/fhir/STU3/questionnaire.html)


ivan-c:
@pbugni I think because the indentation level changed (lost a level of nesting). ie `group.questions` became `items`

PR #2807, opened 2018-11-19
TN-1686 Refactor URL token API to new module.  Add click endpoint to check token

pbugni:


manage.py
```diff
@@ -251,6 +256,22 @@ def purge_user(email, actor):
     permanently_delete_user(email, actor=actor)
 
 
+@click.option('--token', '-t', help="token to lookup/describe")
+@app.cli.command()
+def token_details(token):
+    valid_seconds = app.config.get(
+        'TOKEN_LIFE_IN_DAYS', 30) * 24 * 3600
```
ivan-c: Can you move this implicit default to `config.py` for consistency/DRY?

manage.py
```diff
@@ -251,6 +256,22 @@ def purge_user(email, actor):
     permanently_delete_user(email, actor=actor)
 
 
+@click.option('--token', '-t', help="token to lookup/describe")
```
ivan-c: If the token is always required, this "option" might work better as [an argument](https://click.palletsprojects.com/en/master/arguments/#basic-arguments). Then `--token` or `-t` wouldn't be required

portal/views/portal.py
```diff
@@ -307,21 +308,17 @@ def access_via_token(token, next_step=None):
...
-            current_app.logger.info("token access failed: "
-                                    "expired token {}".format(token))
-            abort(404, "Access token has expired")
-        if not is_valid:
-            abort(404, "Access token is invalid")
-        return user_id
-
     # Confirm the token is valid, and not expired.
     valid_seconds = current_app.config.get(
         'TOKEN_LIFE_IN_DAYS', 30) * 24 * 3600
```
ivan-c: Please remove implicit (duplicate) default

manage.py
pbugni: @ivan-c done - please review again when you get a chance.

PR #2809, opened 2018-11-21
refactor out inline js from flask user/challenge template files

achen2401:
found these while investigating a story:
inline js in login/register/challenge templates

Changes are to move them to separate JS files


portal/static/js/flask_user/register.js
```diff
@@ -0,0 +1,119 @@
...
+        this.setPopoverVis = function(condition, field) {
+            if (condition) {
+                field.removeClass("fail-text").addClass("success-text");
+            } else {
+                field.removeClass("success-text").addClass("fail-text");
+            }
+        };
+        this.initPasswordPopover = function() {
+            var self = this;
+            $("#password").popover({"content": $("#popoverContent").html(), "placement": "top", "html": true, "template": "<div class=\"password-popover popover\" role=\"tooltip\"><div class=\"arrow\"></div><h3 class=\"popover-title\"></h3><div class=\"popover-content\"></div></div>', \"container\": \".password-wrapper\", \"trigger\": \"manual\""}).click(function(e) {
```
ivan-c: Is there a way to keep this HTML in the template and just toggle the visibility?

portal/templates/flask_user/flask_user_base.html
```diff
@@ -1,17 +1,14 @@
 {% extends "layout.html" %}
 
 {% block main %}
-
-            {% block content %}{% endblock %}
-
+	{% block content %}{% endblock %}
 {% endblock %}
 <script>
 {% block document_ready %}
```
ivan-c: If you're able to, I would prefer this block removed from every template :)

I don't think there's an explicit list of everything included in that block, so once it's added to a template it can be difficult to know if it can be removed.

achen2401: yea, block removed, thank you

portal/static/js/flask_user/register.js
achen2401: done, thank you

PR #2818, opened 2018-11-27
TN-1503 prevent user from access cached pages - landing pages only

achen2401:
address: https://jira.movember.com/browse/TN-1503

The frontend fix is to prevent users from accessing cached pages in browser via Back button by manipulating the JS History object - clicking on the Back button will result in user staying on the current page
Changes' scope is narrowed to landing pages for eproms and truenth USA, respectively.  These are the pages the users land on upon logging out.
I myself have tested the fix in Chrome, Firefox, Safari and IE.
**History is cleared in the onclick event of logout button - narrowing down the scope further here**

### DIFF
--- a/portal/static/js/utility.js
+++ b/portal/static/js/utility.js
@@ -343,18 +343,28 @@ function getUrlParameter(name) {
     var results = regex.exec(location.search);
     return results === null ? "" : decodeURIComponent(results[1]);
 }
-function resetBrowserBackHistory() {
+function resetBrowserBackHistory(locationUrl, stateObject, title) {
     var historyDefined = typeof history !== "undefined" && history.pushState;
+    locationUrl = locationUrl || location.href;
     if (historyDefined) {
-        history.pushState(null, null, location.href);
+        history.pushState(stateObject, title, locationUrl);
     }
-    window.addEventListener("popstate", function() {
+    window.addEventListener("popstate", function(e) {
         if (historyDefined) {
-            history.pushState(null, null, location.href);
+            history.pushState(stateObject, title, locationUrl);
         } else {
             window.history.forward(1);
         }
     });
+}
+function handlePostLogout() {
+    if (typeof sessionStorage === "undefined") {
+        return false;
+    }
+    if (sessionStorage.getItem("logout")) {
+        resetBrowserBackHistory(); /* global resetBrowserBackHistory */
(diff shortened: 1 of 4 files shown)


ivan-c:
@achen2401
>clicking on the Back button will result in user staying on the current page...narrowed to landing pages for eproms and truenth USA

Is the history cleared anytime someone visits one of these pages? Although not typical, I can imagine someone incidentally navigating to those pages when exploring the unauthenticated content.

Would instead calling `resetBrowserBackHistory()` at logout cover Warren's test case?

achen2401:
@ivan-c ivan, yea, the history will be cleared upon visiting the page, and your suggestion makes sense - to clear history only on logout - although I am trying to figure out the best way to determine whether a user has logged out - **note** resetBrowserBackHistory code only works on the page from which you want to prevent the user from going back - because it invokes the window's popstate event -  so in this case, the page would be the landing page (the page user lands upon logging out). I am currently looking into perhaps a query string, e.g. /?logged_out=1 l(like we do for a timed out event, e.g. /? timed_out=1 that is passed to the request page), which can serve as a good indicator for a logout event. Will think about this a bit.

achen2401:
ivan, don't know if the change will take but the change I made (frontend code only) is to clear the history on the onclick event of the logout button - this should narrow the scope further

portal/static/js/utility.js
```diff
@@ -343,6 +343,29 @@ function getUrlParameter(name) {
...
+        }
+    });
+}
+function handlePostLogout() {
+    if (typeof sessionStorage === "undefined") {
+        return false;
+    }
+    if (sessionStorage.getItem("logout")) {
+        resetBrowserBackHistory(location.orgin, "logout"); /* global resetBrowserBackHistory */
+        sessionStorage.removeItem("logout");
```
ivan-c: If multiple tabs are open, I think this may only clear the history in the first one. It may be more robust to assume they're not logged in if there's no user cached in sessionStorage

achen2401: > to assume they're not logged in if there's no user cached in sessionStorage

hmm, will this touch on other unauthenticated pages as well if we are making that assumption (e.g.navigate away to an unauthenticated page upon logging out)?

ivan-c: ah right, we wouldn't want to clear it for **every** unauthenticated page

PR #2827, opened 2018-12-03
Unify Smartling translations

ivan-c:
* Add new flask CLI command - `flask download-translations`
  * Downloads translations for every project in `SMARTLING_PROJECT_IDS` and combines
* Refactor Smartling download code
  * Add docstrings, comments


portal/models/i18n_utils.py
```diff
@@ -0,0 +1,163 @@
+"""Module for i18n methods and functionality"""
+from future import standard_library  # isort:skip
+
+standard_library.install_aliases()  # noqa: E402
+
+
+from collections import defaultdict
+import io
+import sys
+import os
```
pbugni: order alpha :)

portal/models/i18n_utils.py
```diff
@@ -0,0 +1,163 @@
...
+from polib import pofile
+
+POT_FILES = (
+    'portal/translations/messages.pot',
+    'portal/translations/js/src/frontend.pot',
+)
+
+
+class BearerAuth(requests.auth.AuthBase):
+    """Add bearer token to HTTP headers for authenticated requests to Smartling"""
```
pbugni: perhaps this is the only location needed, but this utility seems of greater scope than i18n?

ivan-c: @pbugni 
Agreed. Any ideas where a better place might be? It's specific to `requests` module unfortunately

pbugni: @ivan-c  how about `requests_utils.py` ?  or just leave it, and we'll move it out when a second need arises

ivan-c: I thought we might be able to use it in our tests, but it looks like the test [HTTP client isn't `requests` under the covers](http://werkzeug.pocoo.org/docs/0.14/test/#testing-api), so I'll leave it as-is until we can find another use.

ivan-c:
>It would appear that the order in which downloads are handled might matter, as the merge process blindly overwrites a matched key. Should there be a warning if the content doesn't match?

The "translation memory" on Smartling should share translations between projects so they're identical for identical strings.

>Should systems prioritize the respective project to run last?

I want this process to run once at build-time (vs run-time ie `flask sync`), independent of ePROMs/TNUSA configuration. Since ePROMs has significantly more translations, I added it as the last project in the list, just to be safe.
[Added a comment to clarify](https://github.com/uwcirg/true_nth_usa_portal/pull/2827/files#diff-22707aa95790feafb0e4422c44a96006R177)

PR #2833, opened 2018-12-06
Removed the soft_time_limit that causes nothing but thrashing.

pbugni:
It's there if we want it, but we'd need to handle the exception rather than just let it lock up the service, as it interrupts any process when the timeout is exceeded.

### DIFF
--- a/portal/tasks.py
+++ b/portal/tasks.py
@@ -249,7 +249,7 @@ def update_patient_loop(
             update_patients(**kwargs)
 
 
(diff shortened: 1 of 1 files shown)


mcjustin:
"Approved" (don't see a way to do that from my phone)

pbugni:
@ivan-c if you want a high value - fine, i just really want to get this moving and the soft time limits have become a thorn w/o function.
What value can you live with?

PR #2835, opened 2018-12-07
Add nice celery override

ivan-c:
* Add docker-compose override for `celeryworker` service to run with `nice`
  * Prevent celery from consuming all system resources, competing with `web` service
* Add to `COMPOSE_FILE` env var in `docker/.env` list of docker-compose files

### DIFF
--- a/docker/docker-compose.nice-celery.yaml
+++ b/docker/docker-compose.nice-celery.yaml
@@ -1,0 +1,13 @@
+# run celery with nice
+---
+version: "3.2"
+services:
+  celeryworker:
+    cap_add:
+      - SYS_NICE
+    command: bash -c '
+      env &&
+      nice celery worker
+        --app portal.celery_worker.celery
+        --loglevel debug
+      '


ivan-c:
Sure thing Paul!

There's still a section defining celeryworker in docker-compose.yaml - will
this somehow supersede that definition?

Kind of, any details (eg command, restart) in the definition of a service
will be overridden
<https://docs.docker.com/compose/extends/#adding-and-overriding-configuration>
if they're specified in a "later" (appear later in the COMPOSE_FILE list of
files, see .env files on titanium for examples) docker-compose file. You
can use this to separate dev and prod (eg docker-compose.prod.yaml
<https://github.com/uwcirg/true_nth_usa_portal/blob/develop/docker/docker-compose.prod.yaml>)
but still keep it DRY. I'm open to refactoring all the docker-compose YAML
files but I'm not sure the best arrangement (eg will we only need flower on
prod? etc) so I've kept them in separate files for now.


Can we remove the other to eliminate future confusion?
>
I haven't had a chance to test these changes, but assuming they work we can
just update docker.compose.yaml to include everything in
docker-compose.nice-celery.yaml

pbugni:
@ivan-c I've simply waited for your return to merge this one, as I'm sure you'd like to monitor it's success, etc.  Please merge when you are ready.

docker/docker-compose.prod.yaml
```diff
@@ -11,6 +11,15 @@ services:
 
   celeryworker:
     restart: unless-stopped
+    # Set lower CPU priority to prevent blocking web service
```
pbugni: Can you add a comment to explain how one could temporarily disable this feature (for example, by commenting out a few lines within)?  Say we need the celery worker to populate a cache on deploy and want to run in `mean` mode?

pbugni: after conversation, i don't see this as needed.  nice work @ivan-c.

PR #2842, opened 2018-12-12
tn-1734-ironman-org-protocols

mcjustin:
https://jira.movember.com/browse/TN-1734

### DIFF
--- a/portal/config/eproms/Organization.json
+++ b/portal/config/eproms/Organization.json
@@ -1106,6 +1106,13 @@
         {
           "timezone": "America/Chicago",
           "url": "http://hl7.org/fhir/StructureDefinition/user-timezone"
+        },
+        {
+          "research_protocols": [
+			{"name": "IRONMAN v2", "retired_as_of": "2018-07-20T05:00:00Z"},
+			{"name": "IRONMAN v3"}
+          ],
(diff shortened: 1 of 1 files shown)


portal/config/eproms/Organization.json
```diff
@@ -1106,6 +1106,13 @@
         {
           "timezone": "America/Chicago",
           "url": "http://hl7.org/fhir/StructureDefinition/user-timezone"
+        },
+        {
+          "research_protocols": [
+			{"name": "IRONMAN v2", "retired_as_of": "2018-07-20T05:00:00Z"},
```
pbugni: Why is the `retired_as_of` date so far in the past?  This will potentially lead to broken logic, as we assume only a single QB may straddle the transition from old to new.

portal/config/eproms/Organization.json
```diff
@@ -1186,6 +1193,13 @@
         {
           "timezone": "America/Los_Angeles",
           "url": "http://hl7.org/fhir/StructureDefinition/user-timezone"
+        },
+        {
+          "research_protocols": [
+			{"name": "IRONMAN v2", "retired_as_of": "2018-10-18T07:00:00Z"},
```
pbugni: Same as above - please use current and accurate `retired_as_of` dates.

PR #2843, opened 2018-12-13
TN-1617 qb status API

pbugni:
API specialized for MVP, now includes `csv` format option.

### DIFF
--- a/portal/views/reporting.py
+++ b/portal/views/reporting.py
@@ -174,17 +174,17 @@ def questionnaire_status():
     operationId: questionnaire_status
     parameters:
       - name: org_id
-        in: path
-        description: Optional TrueNTH organization ID used to limit results
-          to patients belonging to given organization identifier, and
-          given organization's child organizations
+        in: query
+        description: optional TrueNTH organization ID used to limit results
+          to patients belonging to given organization identifier, and given
+          organization's child organizations
         required: false
         type: integer
         format: int64
       - name: as_of_date
         in: query
-        description: optional query string param to request status
-          at a different (UTC) point in time.  Defaults to now
+        description: optional query string param to request status at a
+          different (UTC) point in time.  Defaults to now
         required: false
         type: string
         format: date-time
--- a/portal/views/reporting.py
+++ b/portal/views/reporting.py
@@ -91,7 +91,8 @@ def generate_overdue_table_html(cutoff_days, overdue_stats, user, top_org):
             for days_overdue, user_id in counts:
                 if days_overdue > curr_min and days_overdue <= cd:
                     uids.append(user_id)
-            count = len([i for i, uid in counts if ((i > curr_min) and (i <= cd))])
+            count = len(
+                [i for i, uid in counts if ((i > curr_min) and (i <= cd))])
             org_row.append(count)
             source_row.append(uids)
             totals[cd] += count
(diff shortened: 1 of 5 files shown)


portal/views/reporting.py
```diff
@@ -152,3 +158,124 @@ def overdue(qstats):
...
+        required: false
+        type: string
+        format: date-time
+      - name: format
+        in: query
+        description: expects json or csv, defaults to json if not provided
+        required: false
+        type: string
+    produces:
+      - application/json
```
mcjustin:
```suggestion
      - application/json
      - text/csv
```

portal/views/reporting.py
```diff
@@ -152,3 +158,124 @@ def overdue(qstats):
...
     output.headers['Content-type'] = "text/csv"
     return output
+
+
+@reporting_api.route('/api/report/questionnaire_status')
+@roles_required(
+    [ROLE.ADMIN.value, ROLE.STAFF.value, ROLE.INTERVENTION_STAFF.value])
+@oauth.require_oauth()
+def questionnaire_status():
+    """Return ad hoc JSON listing questionnaire_status
```
mcjustin:
```suggestion
    """Return ad hoc JSON or CSV listing questionnaire_status
```

portal/views/reporting.py
```diff
@@ -152,3 +158,124 @@ def overdue(qstats):
...
+        in: query
+        description: expects json or csv, defaults to json if not provided
+        required: false
+        type: string
+    produces:
+      - application/json
+    responses:
+      200:
+        description:
+          Returns JSON of the available questionnaire bank status for matching
```
mcjustin:
```suggestion
          Returns JSON or CSV of the available questionnaire bank status for matching
```

portal/views/reporting.py
```diff
@@ -152,3 +158,124 @@ def overdue(qstats):
...
+        def gen(items):
+            desired_order = ['study_id', 'status', 'visit', 'site', 'consent']
+            yield ','.join(desired_order) + '\n'  # header row
+            for i in items:
+                yield ','.join([i.get(k, "") for k in desired_order]) + '\n'
+
+        return Response(
+            gen(results),
+            mimetype='text/plain',
+            headers={'Content-Disposition': 'attachment;filename=status.csv'}
```
mcjustin: just icing: would be nice if this filename had a timestamp (or date at least), and the org param if it exists

portal/views/reporting.py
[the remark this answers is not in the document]
mcjustin: ignore, see 'suggestion' below

mcjustin:
@pbugni  something else, rather significant - this only includes a row for the most recent QB / recurrence for each subject, but we need *all* QB windows (I'm lacking a consensus term for that...). So, a subject at the 6 month point would have a row for baseline, 3 mo, and 6 mo.  This, per the chatter at https://jira.movember.com/browse/TN-1674 .

https://eproms-dev.cirg.washington.edu/api/report/questionnaire_status?org_id=22100&format=csv 
output here for annotation: 
https://docs.google.com/spreadsheets/d/1Pq413qrA-6ibDlL8FWUA02gyW52dw3VyWgj_A7GZ65s/edit#gid=1046676023

PR #2853, opened 2018-12-18
TN-1735 Mechanism to reset cache on research protocol change

pbugni:
Call any site persistence class implementing `invalidation_hook()` when a change is noticed on import.

Only use ATM is for an organization - resets any affected user rows in the qb_timeline as a form of invalidation.

Also cleaned up superfluous log noise on site persistence import - *finally* looks to be only reporting actual changes.

NB: we'd do well to merge this BEFORE #2842 (which is already merged into master).

### DIFF
--- a/portal/models/organization.py
+++ b/portal/models/organization.py
@@ -289,6 +289,35 @@ class Organization(db.Model):
             if rp:
                 return rp
 
+    def invalidation_hook(self):
+        """Endpoint called during site persistence import on change
+
+        Any site persistence aware class may implement ``invalidation_hook``
+        to be notified of changes during import.
+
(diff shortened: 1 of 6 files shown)


portal/models/organization.py
```diff
@@ -289,6 +289,35 @@ def rp_from_org(org):
             if rp:
                 return rp
 
+    def invalidation_hook(self):
```
ivan-c: Should this be part of some base model class that Organization inherits from?

pbugni: if there was a default implementation, yes.  but in general, we don't need it.

for example, you can use an abstract base class to enforce the subclass implements, but again, the general case is to do nothing, so that would just be empty boiler plate, imo.

PR #2856, opened 2018-12-19
TN-1753 address status qb_timeline.at collision

pbugni:
See bug for details - as we add the `completed` status after the `due` row to qb_timeline, adding a second `order_by` criteria (qb_timeline.id) makes this case deterministic.

(Thanks @ivan-c for the idea!)


pbugni:
merging now for use on eproms-stage.

PR #2859, opened 2018-12-20
TN-1737-fix-language-syntax

mcjustin:
... for both 14701 "CHU de Quebec - Universite Laval", and 14669 "Tygerberg Hospital"

### DIFF
--- a/portal/config/eproms/Organization.json
+++ b/portal/config/eproms/Organization.json
@@ -1416,6 +1416,7 @@
           "timezone": "America/New_York",
           "url": "http://hl7.org/fhir/StructureDefinition/user-timezone",
           "valueCodeableConcept": {
+            "url": "http://hl7.org/fhir/valueset/languages",
             "coding": [
               {
                 "code": "fr_CA",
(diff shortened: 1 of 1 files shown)


portal/config/eproms/Organization.json
```diff
@@ -1416,6 +1416,7 @@
           "timezone": "America/New_York",
           "url": "http://hl7.org/fhir/StructureDefinition/user-timezone",
           "valueCodeableConcept": {
+            "url": "http://hl7.org/fhir/valueset/languages",
```
achen2401: @mcjustin Justin, shouldn't there be a bracket for the timezone - as it is a separate extension item, and I think system url shouldn't be part of the valueCodeableConcept dictionary item:
``{
         "timezone": "America/New_York",
          "url": "http://hl7.org/fhir/StructureDefinition/user-timezone"
    **}**, 
   {
         **"url": "http://hl7.org/fhir/valueset/languages"**,
         "valueCodeableConcept": {
         ......
}``

but correct me if I am wrong though, was just doing comparisons with others in the json

mcjustin: @achen2401  can you please review again?

portal/config/eproms/Organization.json
```diff
@@ -1414,8 +1414,11 @@
       "extension": [
         {
           "timezone": "America/New_York",
-          "url": "http://hl7.org/fhir/StructureDefinition/user-timezone",
+          "url": "http://hl7.org/fhir/StructureDefinition/user-timezone"
+        },
+		{
           "valueCodeableConcept": {
+            "url": "http://hl7.org/fhir/valueset/languages",
```
achen2401: @mcjustin Justin, I think, the url item, ```http://hl7.org/fhir/valueset/languages```, is misplaced.  I think it should be present before valueCodeableConcept and after the beginning bracket, right?
```{ "url": "http://hl7.org/fhir/valueset/languages", "valueCodeableConcept": {...```

mcjustin: sigh - please review again @achen2401

PR #2864, opened 2018-12-25
TN-1761 Add translation files to git

ivan-c:
* Track translation files in git
  * ran `flask download-translations` and added files
    * Downloaded and combined translations from ePROMs & TrueNTH USA

* Use `flask download-translations` universally
* updated `flask download-translations`
  * Remove line-wrapping from
  * Fix missing PO metadata in merged files


portal/models/i18n_utils.py
```diff
@@ -201,11 +201,15 @@ def pos_from_zip(zipfile):
         yield locale_code, po
 
 
-def msgcat(*po_files):
-    """Concatenate input po_files together, with later files overwriting earlier ones"""
+def msgcat(*po_files, **kwargs):
+    """
```
pbugni: when you have a single line comment - it should just be a single line, not a block, i.e.
`"""Concatenate input po_files together, with later files overwriting earlier ones"""`

PR #2867, opened 2018-12-26
TN-1764 PO-to-MO compiler

ivan-c:
* Add `flask compile-po-files` for compiling backend PO files to MO files
  * flask requires MO files to load translations from
* Ignore generated MO files


portal/models/i18n_utils.py
```diff
@@ -276,3 +276,24 @@ def download_all_translations(state):
                 "Saved combined MO file: %s",
                 os.path.relpath(dest_mo, current_app.root_path),
             )
+
+
+def compile_pos():
+    """
```
pbugni: should be single line comment w/o params or need for verbosity...

PR #2869, opened 2018-12-27
TN-1760 fix TODO - refactor coredata JS file

achen2401:
address https://jira.movember.com/browse/TN-1760
in coredata JS
- remove global functions - use methods of object instead
- add error message display
- provide default footer, retrieve GIL footer based on configuration

### DIFF
--- a/portal/static/js/coredata.js
+++ b/portal/static/js/coredata.js
@@ -1,6 +1,6 @@
 (function() {  /*global $, tnthAjax, SYSTEM_IDENTIFIER_ENUM, i18next */
     var CoreDataObj = function() {
-        this.subjectId = 0;
+        this.subjectId = null;
         this.init = function(subjectId) { /* entry point for initiating CoreDataObj */
             var self = this;
             this.setSubjectId(subjectId, function() {
@@ -20,7 +20,7 @@
                 callback({success:1});
                 return;
             }
-            tnthAjax.getCurrentUser(function(data) {
+            tnthAjax.getCurrentUser(function(data) { //this will get session storage current user first before firing off ajax
                 if (!data || data.error) {
                     self.setError(i18next.t("Error occurred retrieving subject ID"));
                     callback({error: true});
@@ -114,7 +114,7 @@
                 try {
                     window.location.replace(self.getReturnAddress());
                 } catch(e) {
-                    self.setError(i18next.t("Error occurred when redirecting to destination url"));
+                    self.setError(i18next.t("Error occurred - unable to reach destination"));
                     //report error if invalid return address is used here
                     tnthAjax.reportError(self.subjectId, self.getAPIUrl(), e.message, true);
                     $(this).attr("disabled", false);
@@ -148,5 +148,8 @@
     $(document).ready(function(){
         (new CoreDataObj()).init();
     });
-})();  /*eslint wrap-iife: off */
+})();
+
+
+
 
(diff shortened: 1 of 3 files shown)


portal/static/js/coredata.js
```diff
@@ -1,54 +1,152 @@
...
+                }
+                self.getDemoData();
+                self.initFieldEvents();
+                self.setFooter();
+            });
+        };
+        this.setSubjectId = function(subjectId, callback) {
+            callback = callback || function() {};
+            var self = this;
+            this.subjectId = subjectId || $("#coreDataUserId").val(); //check user id in template, if any
```
ivan-c: Could you fallback to a value from session storage instead? Probably more reliable than the DOM

portal/static/js/coredata.js
```diff
@@ -1,54 +1,152 @@
...
-        } catch(e) {
-            //report error if invalid return address is used here
-            tnthAjax.reportError(subjectId, $("procAPIUrl").val(), e.message, true);
-            $(this).attr("disabled", false);
-            $(".loading-indicator").hide();
-            $(".error-continue").text(e.message);
-        }
+(function() {  /*global $, tnthAjax, SYSTEM_IDENTIFIER_ENUM, i18next */
+    var CoreDataObj = function() {
+        this.subjectId = 0;
```
ivan-c: `null` might be a better uninitialized value, then you can more easily tell if it's been initialized or not

portal/static/js/coredata.js
```diff
@@ -1,54 +1,152 @@
...
+            // Class for both "done" and "skip" buttons
+            var self = this;
+            $(".continue-btn").on("click", function(event){
+                event.preventDefault();
+                $(this).attr("disabled", true);
+                $(".loading-indicator").show();
+                try {
+                    window.location.replace(self.getReturnAddress());
+                } catch(e) {
+                    self.setError(i18next.t("Error occurred when redirecting to destination url"));
```
ivan-c: This is a very technical, specific error message.

I don't think it should necessarily be shown to a user, but would still be helpful to a developer for debugging (maybe change to log message?).

portal/static/js/coredata.js
```diff
@@ -1,54 +1,152 @@
...
+                }
+            });
+        };
+    };
+    $(document).ready(function(){
+        (new CoreDataObj()).init();
     });
-    tnthAjax.getPortalFooter(subjectId, false, "core_data_footer");
-});
+})();  /*eslint wrap-iife: off */
```
ivan-c: If I understand correctly, this disables the `wrap-iife` linting rule for this entire function...

PS some notes about [IIFE style](http://airbnb.io/javascript/#functions--iife)

portal/views/portal.py
```diff
@@ -886,6 +886,7 @@ def config_settings(config_key):
         'LOCALIZED_AFFILIATE_ORG',
         'COPYRIGHT',
         'MAINTENANCE_',
+        'GIL',
```
ivan-c: Can you keep it in alphabetized order? :nerd_face:

portal/templates/coredata.html
```diff
@@ -79,12 +78,13 @@ <h3 class="tnth-headline">{{ _("More About You") }}</h3>
...
     <br/><br/><br/><br/>
 {% endif %}
+<div id="coreDataError" class="error-message"></div>
 <input type="hidden" id="coreDataUserId" value="{{user.id if user}}" />
 <input type="hidden" id="procReturnAddress" value="{{ return_address | safe }}" />
 <input type="hidden" id="procAPIUrl" value="{{url_for('coredata_api.acquire')}}" />
 {% endblock %}
 {% block footer %}
-    <footer id="core_data_footer"></footer>
+    <footer id="core_data_footer">{{footer(user=user)}}</footer>
```
ivan-c: Does the footer require a user object now? Would prefer to get that info from an API or sessionStorage, if possible

PR #2875, opened 2018-12-28
Feature/utility js refactor

achen2401:
https://jira.movember.com/browse/TN-1766
refactor utility JS
- create utility object and remove global functions to prevent pollution of global name space (e.g. prevent name collisions)
- linting code

### DIFF
--- a/portal/eproms/templates/eproms/privacy.html
+++ b/portal/eproms/templates/eproms/privacy.html
@@ -12,12 +12,15 @@
 {% block document_ready %}
 	{% if request.args.get("disableLinks", "") %}
 	$("document").ready(function() {
-		setTimeout("disableHeaderFooterLinks();", 1000);
+		setTimeout(function() {
+			Utility.disableHeaderFooterLinks();
+		}, 1000);
 	});
 	{% endif %}
 {% endblock %}
 {% block additional_scripts %}
 	{% if editorUrl %}
-	<script>appendLREditContainer($("#privacyContentContainer"), "{{editorUrl}}", {% if user and user.has_role(ROLE.CONTENT_MANAGER.value) %}true{% else %}false{%endif%});</script>
+	<!-- TODO in a separate JS-->
+	<script>Utility.appendLREditContainer($("#privacyContentContainer"), "{{editorUrl}}", {% if user and user.has_role(ROLE.CONTENT_MANAGER.value) %}true{% else %}false{%endif%});</script>
 	{% endif %}
 {% endblock %}
--- a/portal/eproms/templates/eproms/privacy.html
+++ b/portal/eproms/templates/eproms/privacy.html
@@ -21,6 +21,6 @@
 {% block additional_scripts %}
 	{% if editorUrl %}
 	<!-- TODO in a separate JS-->
-	<script>Utility.appendLREditContainer($("#privacyContentContainer"), "{{editorUrl}}", {% if user and user.has_role(ROLE.CONTENT_MANAGER.value) %}true{% else %}false{%endif%});</script>
+	<script>Utility.appendLREditContainer($("#privacyContentContainer"), "{{editorUrl}}", {{ user and user.has_role(ROLE.CONTENT_MANAGER.value) | lower }});</script>
 	{% endif %}
 {% endblock %}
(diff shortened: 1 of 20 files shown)


portal/eproms/templates/eproms/privacy.html
```diff
@@ -12,12 +12,15 @@ <h1 class="tnth-headline">{{ _("Privacy") }}</h1>
...
+			Utility.disableHeaderFooterLinks();
+		}, 1000);
 	});
 	{% endif %}
 {% endblock %}
 {% block additional_scripts %}
 	{% if editorUrl %}
-	<script>appendLREditContainer($("#privacyContentContainer"), "{{editorUrl}}", {% if user and user.has_role(ROLE.CONTENT_MANAGER.value) %}true{% else %}false{%endif%});</script>
+	<!-- TODO in a separate JS-->
+	<script>Utility.appendLREditContainer($("#privacyContentContainer"), "{{editorUrl}}", {% if user and user.has_role(ROLE.CONTENT_MANAGER.value) %}true{% else %}false{%endif%});</script>
```
ivan-c: Something like this might be more terse...
You may need to call `.lower()`, but I think `True` and `False` should get correctly evaluated into the javascript equivalent of `true` and `false`
```suggestion
 	<script>Utility.appendLREditContainer($("#privacyContentContainer"), "{{editorUrl}}", {{ user and user.has_role(ROLE.CONTENT_MANAGER.value) }});</script>
```

portal/static/js/admin.js
```diff
@@ -891,7 +891,7 @@
                         });
                         documents.forEach(function (item) {
                             var c = item["contributor"];
-                            if (!existingItems[c] && hasValue(c)) { //only draw the most recent, same report won't be displayed
+                            if (!existingItems[c] && c) { //only draw the most recent, same report won't be displayed
```
ivan-c: I think you would want to check `c` first?...
```suggestion
                            if (c && !existingItems[c]) { //only draw the most recent, same report won't be displayed
```

portal/static/js/initialQueries.js
```diff
@@ -699,14 +699,14 @@
         var termsEvent = function() {
             if ($(this).attr("data-agree") === "false") {
                 var types = $(this).attr("data-tou-type");
-                if (hasValue(types)) {
+                if (types) {
```
ivan-c: This function gets pretty deeply indented. You could remove a layer of indentation if you inverted this into a guard clause and exited early if that condition isn't met.

portal/static/js/profile.js
```diff
@@ -388,10 +388,10 @@
                         var m = "", d = "", y = "", displayDeceasedDate = "";
                         if (data.deceasedDateTime) {
                             var deceasedDateObj = new Date(data.deceasedDateTime);
-                            m = pad(deceasedDateObj.getUTCMonth()+1);
+                            m = self.pad(deceasedDateObj.getUTCMonth()+1);
                             d = deceasedDateObj.getUTCDate();
                             y = deceasedDateObj.getUTCFullYear(); /*global pad*/
-                            displayDeceasedDate = self.modules.tnthDates.displayDateString(pad(m), pad(d), y);
+                            displayDeceasedDate = self.modules.tnthDates.displayDateString(self.pad(m), self.pad(d), y);
```
ivan-c: If you haven't already, it would probably be cleaner to standardize on passing around datetime objects (than strings)...

portal/static/js/utility.js
```diff
@@ -45,59 +17,57 @@ function showMain() {
...
+        }, 100);
+        this.hideLoader(true, 350);
     }
-}
-function _isTouchDevice() {
+};
+UtilityObj.prototype.isDelayLoading = function() { /*global DELAY_LOADING*/
+    return (typeof DELAY_LOADING !== "undefined") && DELAY_LOADING;
+};
+UtilityObj.prototype.isTouchDevice = function() {
```
ivan-c: Is there a way to move this kind of differentiation lower in the code? I worry it highlights the differentiation between devices, instead of smoothing it over.

Ideally we would have a cohesive and singular view of handling any browser client, only handling browser idiosyncrasies as necessary.

portal/static/js/utility.js
```diff
@@ -200,46 +180,46 @@ function LRKeyEvent() {
...
-        }
+};
+UtilityObj.prototype.getLoaderHTML = function(message) {
+    return "<div class=\"loading-message-indicator\"><i class=\"fa fa-spinner fa-spin fa-2x\"></i>" + (message ? "&nbsp;" + message : "") + "</div>";
+};
+UtilityObj.prototype.convertToNumericField = function(field) {
+    if (!field) {
+        return;
+    }
+    if (this.isTouchDevice()) {
```
ivan-c: Would this property be harmful to non-touch devices?

achen2401: thank you ivan, yeah redundant check, as, if the browser support is non-existent, it will fallback to the type of text.

PR #2893, opened 2019-01-08
Quicker celerybeat healthy

ivan-c:
* Set celerybeat health directly through redis (no http)
* Set celerybeat healthy as soon as redis available


ivan-c:
>Just to clarify, we're using local time rather than UTC in health checks? (i.e. datetime.now() vs datetime.utcnow()). I'd rather see us use UTC across the board, but it's not worth a refactor for this PR...

The redis key is self-expiring and unrelated to that value, so no harm in updating it to use GMT

PR #2894, opened 2019-01-08
Cleanup project configuration

ivan-c:
* Remove old email configs
  * `MAIL_SENDER_DEFAULT` dynamically generated now
* Remove old assessment engine configs (replaced with Questionnaire resource stubs)

### DIFF
--- a/portal/config/eproms/site_persistence_file.json
+++ b/portal/config/eproms/site_persistence_file.json
@@ -5,8 +5,6 @@
       "results": [
         "## SITE-PERSISTENCE Managed: File managed with site_persistence - don't edit.\n",
         "DEFAULT_LOCALE = 'en_AU'\n",
-        "MAIL_SENDER_DEFAULT = 'noreply@eproms.truenth.org'\n",
-        "CONTACT_SENDTO_EMAIL = 'contact.eproms@truenth.org'\n",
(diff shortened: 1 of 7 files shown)


portal/config/eproms/site_persistence_file.json
```diff
@@ -5,8 +5,6 @@
       "results": [
         "## SITE-PERSISTENCE Managed: File managed with site_persistence - don't edit.\n",
         "DEFAULT_LOCALE = 'en_AU'\n",
-        "MAIL_SENDER_DEFAULT = 'noreply@eproms.truenth.org'\n",
-        "CONTACT_SENDTO_EMAIL = 'contact.eproms@truenth.org'\n",
```
mcjustin: @ivan-c the nice thing about having a value that's clearly a default (or a key with an empty value) is that the administrator knows what values they need to populate. Do we let them know this in some other way?

ivan-c: Good point. I've been keeping configs that are required in files ending with `.default` that I copy and modify when I setup a new deployment (eg [`portal.env.default`](https://github.com/uwcirg/true_nth_usa_portal/blob/develop/docker/portal.env.default) -> `portal.env`).

I'll add stubbed-out/commented references for values we're currently using

PR #2899, opened 2019-01-09
JS Modules- Build Artifacts & Deploy

ivan-c:
* Rewrite gulp wrapper script for generic `npm` commands
* Use webpack to build all frontend files
* Build front-end files during traditional deploy
* Add generated files to .gitignore


ivan-c:
>ivan, thank you very much, looks good to me. I was wondering if we should add the built files (e.g.JS files in js/dist and maps and css files) to .gitignore file so they don't get checked in accidentally?

:+1:

ivan-c:
@achen2401 I added 2 .gitignore files, but put them in the same directory as the generated files. I couldn't find any "non-generated files" and git requires at least one file in a directory, so I used the these .gitignore files:

[portal/static/maps/.gitignore](https://github.com/uwcirg/true_nth_usa_portal/blob/15fb62aa8306ed41a14e96899e3e876194ebe31b/portal/static/maps/.gitignore)
[portal/static/js/dist/.gitignore](https://github.com/uwcirg/true_nth_usa_portal/blob/15fb62aa8306ed41a14e96899e3e876194ebe31b/portal/static/js/dist/.gitignore)

PR #2930, opened 2019-01-17
TN-1805 Expand scheduled job syntax to allow for random minute

pbugni:


portal/models/scheduled_job.py
```diff
@@ -32,10 +33,14 @@ def schedule(self):
...
     def schedule(self, sc):
         # schedule must match cron schedule pattern * * * * *
-        format = r'([\*\d,-\/]+)\s([\*\d,-\/]+)\s([\*\d,-\/]+)' \
+        # with the option of using an 'r' in the minutes column
+        # to request random minute generation
+        format = r'([\*\d,-\/,r]+)\s([\*\d,-\/]+)\s([\*\d,-\/]+)' \
                  r'\s([\*\d,-\/]+)\s([\*\d,-\/]+)$'
         if not sc or not re.match(format, sc):
             raise Exception("schedule must be in valid cron format")
+        if sc.startswith('r'):
```
ivan-c: Is there any reasonable way to fold this into child class of [`celery.schedules.crontab`](https://celery.readthedocs.io/en/latest/reference/celery.schedules.html#celery.schedules.crontab)?

portal/models/scheduled_job.py
```diff
@@ -32,10 +33,14 @@ def schedule(self):
     @schedule.setter
     def schedule(self, sc):
         # schedule must match cron schedule pattern * * * * *
-        format = r'([\*\d,-\/]+)\s([\*\d,-\/]+)\s([\*\d,-\/]+)' \
+        # with the option of using an 'r' in the minutes column
+        # to request random minute generation
+        format = r'([\*\d,-\/,r]+)\s([\*\d,-\/]+)\s([\*\d,-\/]+)' \
```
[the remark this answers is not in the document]
ivan-c: Helpful suggestion :)

ivan-c:
```suggestion
        cron_format = r'([\*\d,-\/,r]+)\s([\*\d,-\/]+)\s([\*\d,-\/]+)' \
```

portal/models/scheduled_job.py
```diff
@@ -32,10 +33,14 @@ def schedule(self):
...
                  r'\s([\*\d,-\/]+)\s([\*\d,-\/]+)$'
         if not sc or not re.match(format, sc):
```
ivan-c:
```suggestion
         if not sc or not re.match(cron_format, sc):
```

portal/models/scheduled_job.py
pbugni: i don't see the value - you see a second client of `celery.schedules.crontab` in our future?

ivan-c: The schedule setup code seems like it would be more appropriate in the `crontab` class than one of our model classes. 

I agree that there's likely little benefit, I don't see us using more than one instance of `crontab`.

PR #2939, opened 2019-01-22
Feature/reporting data ui

achen2401:
Add export report data UI for this story:  https://jira.movember.com/browse/TN-1754

- add UI element to the patients list that allows export of patient report data
- patients list styling fix
- cleanup admin JS code in the process
  - move reusable code (e.g. current user) out as re-usable component (can be used by others)
  - move research related code to separate module


achen2401:
@mcjustin Justin, will do.  The UI changes were based on feedback from Diana yesterday (I had previous attached screenshots as they appeared in my own instance)  Will attach updated ones.

portal/static/js/src/modules/TnthAjax.js
```diff
@@ -757,28 +757,23 @@ export default { /*global $ */
...
-                            /*
-                             * add instrument name to instruments array for the org - will not add if it is already in the array
-                             * NOTE: inArray returns -1 if the item is NOT in the array
-                             */
-                            if ($.inArray(q.questionnaire.display, qList[orgID]) === -1) {
-                                qList[orgID].push(q.questionnaire.display);
-                            }
-                        });
-                    }
+                if (!item.resource || !item.resource.questionnaires) {
```
ivan-c: :medal_sports:

portal/templates/admin/patients_by_org.html
```diff
@@ -35,6 +35,7 @@ <h4 class="tnth-headline">{% trans %}Patient List{% endtrans %}</h4>
              data-toggle="table"
              data-search="true"
              data-pagination="true"
+             {% if 'status' in config.PATIENT_LIST_ADDL_FIELDS %}data-report-export="true"{% endif %}
```
ivan-c: Might be nice to move that entirely to the frontend :)

portal/templates/research.html
```diff
@@ -1,62 +1,15 @@
...
-              <a v-bind:href="'/api/patient/assessment?' + instruments.selected + '&format=' + instruments.dataType" class="btn btn-default" id="patientsDownloadButton" v-bind:disabled="!hasInstrumentsSelection()">{{ _("Export") }}</a>
-              <button type="button" class="btn btn-default" data-dismiss="modal">{{ _("Close") }}</button>
-            </div>
-          </div>
-        </div>
-    </div>
-    <div id="patientListExportDataContainer" class="btn btn-lg"><a href="#" id="patientAssessmentDownload"  data-toggle="modal" data-target="#dataDownloadModal">{{ _("Export questionnaire data") }} &nbsp;<span class="glyphicon glyphicon-save"></span></a></div>
+    <div id="researchContentContainer"></div>
+  {% else %}
+    <span class="muted">{{_("Content to be determined")}}</span>
```
ivan-c: Please remove translation call until content finalized

portal/templates/admin/patients_by_org.html
achen2401: done, thanks ivan

portal/static/js/src/components/ExportInstruments.vue
```diff
@@ -0,0 +1,150 @@
...
+            getInstrumentList: function () {
+                if (this.instrumentsList && this.instrumentsList.length) {
+                    this.setInstrumentsListContent(this.instrumentsList);
+                    this.setInstrumentInputEvent();
+                    return;
+                }
+                var self = this;
+                tnthAjax.getInstrumentsList(false, function (data) {
+                    if (!data || !data.length) {
+                        document.querySelector("#instrumentsExportErrorMessage").innerText = i18next.t("Error occurred retrieving instruments list");
```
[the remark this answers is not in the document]
achen2401: fixed the message to make it more user-friendly so as to be displayed to the user  - any error encountered during an ajax call via tnthAjax object will be logged automatically

portal/static/js/src/components/ExportInstruments.vue
```diff
@@ -89,7 +89,7 @@
                 var self = this;
                 tnthAjax.getInstrumentsList(false, function (data) {
                     if (!data || !data.length) {
-                        document.querySelector("#instrumentsExportErrorMessage").innerText = i18next.t("Error occurred retrieving instruments list");
+                        document.querySelector("#instrumentsExportErrorMessage").innerText = i18next.t("Unable to retrieving instruments list.");
```
ivan-c: Sorry for not being more clear- 
I don't think we need to include a reference to the specific API that failed- otherwise we would need different error messages for each AJAX call or API (maybe `tnthAjax` could return a generic error instead?).

This wouldn't normally be a big deal, but with i18n and 10+ languages I'd like to minimize the translation churn.

Since we're already translating this string, could you either use the original (so we don't have to re-translate it), or remove the translation call (`i18next.t()`)?

achen2401: thanks, ivan, ok, think I got it now.  I have it just display the generic error returned from the call.  hope this will take.

PR #2942, opened 2019-01-23
allow admin staff to create ironman patient - for testing purpose

achen2401:
restore the ability of admin user to create ironman patient when MedidataRave configuration is set (i.e. production)

### DIFF
--- a/portal/static/js/src/accountCreation.js
+++ b/portal/static/js/src/accountCreation.js
@@ -504,7 +504,7 @@ import {CurrentUserObj} from "./mixins/CurrentUser.js";
                 return OT.getOrgName(item);
             });
         };
-        this.handleMedidataRave = function(orgs, noDisable) {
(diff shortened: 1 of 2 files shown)


portal/static/js/src/accountCreation.js
```diff
@@ -492,16 +492,26 @@ import ProcApp from "./modules/Procedures.js";
...
+            });
+        };
         this.getCurrentUserTopLevelOrgs = function() {
             var topLevelOrgs = OT.getUserTopLevelParentOrgs(leafOrgs);
             return topLevelOrgs.map(function(item) {
                 return OT.getOrgName(item);
             });
         };
-        this.handleMedidataRave = function() {
+        this.handleMedidataRave = function(orgs, noDisable) {
```
ivan-c: Could you add a comment (if there isn't one already) describing why we're disabling editing?

achen2401: thanks, ivan, added

PR #2950, opened 2019-01-25
Feature/js validator code cleanup

achen2401:
Followup code-clean to this story:
https://jira.movember.com/browse/TN-1751

- move field validation code to new module so it can be re-used for others
- remove duplicate from profile.js
- linting JS

### DIFF
--- a/portal/static/js/src/modules/Validator.js
+++ b/portal/static/js/src/modules/Validator.js
@@ -80,5 +80,3 @@ export default ValidatorObj;
 export var emailValidation = ValidatorObj.emailValidation;
 export var htmltagsValidation = ValidatorObj.htmltagsValidation;
 export var birthdayValidation = ValidatorObj.birthdayValidation;
-
-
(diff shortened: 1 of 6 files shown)


portal/static/js/src/modules/Validator.js
```diff
@@ -0,0 +1,81 @@
+import Utility from "./Utility.js";
+import tnthDates from "./TnthDate.js";
+
+var ValidatorObj = { /*global  $ i18next */
+    "birthdayValidation": function() {
+        var m = parseInt($("#month").val()), d = parseInt($("#date").val()), y = parseInt($("#year").val());
```
ivan-c: Probably a little out of scope for this PR, but I might try to pass the "inputs" these functions require as arguments and return errors (instead of writing directly to DOM).

The "inputs" could be passed as values (strings/numbers), or the elements if you're using [HTML5 constraint validation](https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/HTML5/Constraint_validation)

Using arguments and returning errors would allow looser coupling between the DOM and frontend code and more code flexibility (eg write errors to console.log instead of DOM).

achen2401: thanks, ivan, will try to fix

PR #2957, opened 2019-01-28
TN-1799 - Invited user IDs not considered in unique email check

pbugni:
We'll now catch "duplicates" in the unique email check, even if they're stuck in the __invited__ state.

### DIFF
--- a/portal/views/user.py
+++ b/portal/views/user.py
@@ -35,6 +35,7 @@ from ..models.relationship import Relationship
 from ..models.role import ROLE, Role
 from ..models.table_preference import TablePreference
 from ..models.user import (
+    INVITE_PREFIX,
     User,
     UserRelationship,
     current_user,
@@ -1547,6 +1548,12 @@ def unique_email():
         result = match.one()
         if user_id != result.id:
             return jsonify(unique=False)
+
+    # Look out for "masked" emails, as they'll create collisions down the road
+    masked = INVITE_PREFIX + email
+    match = User.query.filter(func.lower(User.email) == masked.lower())
+    if match.count():
+        return jsonify(unique=False)
     return jsonify(unique=True)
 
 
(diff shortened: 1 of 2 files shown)


tests/test_user.py
```diff
@@ -110,6 +110,19 @@ def test_unique_email(self):
...
         assert results['unique'] is True
 
+    def test_unique_hidden_email(self):
+        # hidden emails, such as masked invited user create collisions
+        email = 'fake@email.com'
+        fake_user = self.add_user(username=email)
+        fake_user.mask_email()
+
+        self.login()
+        request = '/api/unique_email?email={}'.format(urllib.parse.quote(
```
ivan-c: Nitpick: could you use `query_string` kwarg to `self.client.get` to build the URL instead of string formatting?

It's a little more work to setup, but is much easier to maintain
eg:
```
response = self.client.get(
    '/api/unique_email', 
    query_string={'email': urllib.parse.quote(email)},
)
```

portal/views/user.py
```diff
@@ -1547,6 +1548,12 @@ def unique_email():
         result = match.one()
         if user_id != result.id:
             return jsonify(unique=False)
+
+    # Look out for "masked" emails, as they'll create collisions down the road
+    masked = INVITE_PREFIX + email
+    match = User.query.filter(func.lower(User.email) == masked.lower())
+    if match.count():
```
ivan-c: Might make sense to move to the top of the function, seems easier to check and independent of a lot of other conditions.

pbugni: Intentionally inserted after the above logic, handling complexity of the user changing his/her own password.

tests/test_user.py
pbugni: Yes, just a cut and paste detail.  Perhaps you'd like to add an issue to go through all of our tests and make this change?  I agree with your nitpick - lets sweep through and get 'em all.

PR #2962, opened 2019-01-30
TN-1751 followup cleanup to Consent/orgs js code

achen2401:
extracting out consent, orgs specific code from profile.js to be re-usable by others (e.g. initial queries JS)
- moving consent specific code to individual modules
- move org specific code from profile.js to OrgTool module
note, content of the code remains relatively the same, just moving around pieces

### DIFF
--- a/portal/static/js/src/modules/Consent.js
+++ b/portal/static/js/src/modules/Consent.js
@@ -1,0 +1,334 @@
+import OrgTool from "./OrgTool";
+import tnthAjax from "./TnthAjax";
+import tnthDates from "./TnthDate";
+import CONSENT_ENUM from "./CONSENT_ENUM";
+import Utility from "./Utility";
+
+export default { /*global i18next datepicker $*/
+    orgTool: null,
+    getOrgTool: function() {
+        if (!this.orgTool) {
+            this.orgTool = new OrgTool();
+            this.orgTool.init();
+        }
+        return this.orgTool;
+    },
+    initFieldEvents: function(userId) {
+        var __self = this, orgTool = this.getOrgTool(), modalElements = $("#consentContainer .modal, #defaultConsentContainer .modal");
+        var closeButtons = modalElements.find("button.btn-consent-close, button[data-dismiss]");
+        $("#consentHistoryModal").modal({"show": false});
+        modalElements.each(function() {
+            var agreemntUrl = $(this).find(".agreement-url").val();
+            if (/stock\-org\-consent/.test(agreemntUrl)) {
+                $(this).find(".terms-wrapper").hide();
+            }
+        });
+        modalElements.find("input[name='toConsent']").off("click").on("click", function(e) {
+            e.stopPropagation();
+            closeButtons.attr("disabled", true);
+            var orgId = $(this).attr("data-org");
+            var postUpdate = function(orgId, errorMessage) {
+                if (errorMessage) {
+                    $("#"+orgId+"_consentAgreementMessage").html(errorMessage);
+                } else {
+                    $("#"+orgId+"_consentAgreementMessage").html("");
+                    setTimeout(function() { modalElements.modal("hide"); __self.removeObsoleteConsent(userId); }, 250);
+                    //setTimeout(function() { __self.reloadConsentList(userId);}, 500);
+                }
+                $("#" + orgId + "_loader.loading-message-indicator").hide();
+                closeButtons.attr("disabled", false);
+            };
+            $("#" + orgId + "_loader.loading-message-indicator").show();
+            if ($(this).val() === "yes") {
+                var params = CONSENT_ENUM.consented;
+                params.org = orgId;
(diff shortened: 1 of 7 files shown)


portal/static/js/src/modules/Consent.js
```diff
@@ -0,0 +1,337 @@
...
+        });
+        $("#consentDateModal").on("hidden.bs.modal", function() {
+            $(this).removeClass("active");
+        });
+        $("#consentDateModal .consent-date").datepicker({"format": "d M yyyy","forceParse": false,"endDate": new Date(),"autoclose": true});
+        $("#consentDateModal .consent-hour, #consentDateModal .consent-minute, #consentDateModal .consent-second").each(function() {
+            Utility.convertToNumericField($(this));
+        });
+        $("#consentDateModal .consent-date, #consentDateModal .consent-hour, #consentDateModal .consent-minute, #consentDateModal .consent-second").each(function() {
+            $(this).on("change", function() {
```
ivan-c: Might make sense to move this to a separate validation function

portal/static/js/src/modules/Consent.js
```diff
@@ -0,0 +1,337 @@
...
+        $("#consentDateModal .btn-submit").on("click", function() {
+            var ct = $("#consentDateModal_date"), o = CONSENT_ENUM[ct.attr("data-status")];
+            if (!ct.val()) {
+                $("#consentDateModalError").text(i18next.t("You must enter a date/time"));
+                return false;
+            }
+            var h = $("#consentDateModal_hour").val(),m = $("#consentDateModal_minute").val(),s = $("#consentDateModal_second").val();
+            var dt = new Date(ct.val()); //2017-07-06T22:04:50 format
+            var pad = function(n) { if (!n) { return "00"; } n = parseInt(n); return (n < 10) ? "0" + n : n; };
+            var cDate = dt.getFullYear()+"-"+(dt.getMonth() + 1)+"-"+dt.getDate()+"T"+pad(h)+":"+pad(m)+":"+pad(s);
```
ivan-c: Is this a standard format that could reuse an existing datetime formatter?

portal/static/js/src/modules/OrgTool.js
```diff
@@ -350,6 +353,300 @@ export default (function() { /*global i18next $ */
...
+                    self.updateOrgs(userId, $("#clinics"),true);
+                    var thisElement = $(this);
+                    setTimeout(function() {
+                        Consent.setConsentBySelectedOrg(userId, thisElement, isConsentWithTopLevelOrg);
+                    }, 500);
+                }
+            });
+        });
+    };
+    OrgTool.prototype.setOrgsVis = function(data, callback) {
```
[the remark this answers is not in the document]
ivan-c: There's a lot of nesting here, you might try to move nested if statements to guard clauses and exit early (`continue`/`return`) where possible.

Another possibility (at least one I'd try in python :) would be to use a [generator function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Iterators_and_Generators#Generator_functions) to yield only the orgs you need to act on.

portal/static/js/src/modules/OrgTool.js
```diff
@@ -350,6 +353,299 @@ export default (function() { /*global i18next $ */
...
+        callback = callback || function() {};
+        if (!data || ! data.careProvider) { callback(); return false;}
+        for (var i = 0; i < data.careProvider.length; i++) {
+            var val = data.careProvider[i];
```
ivan-c: `careProvider` might be a better variable name than `val`

portal/static/js/src/modules/OrgTool.js
```diff
@@ -350,6 +353,299 @@ export default (function() { /*global i18next $ */
...
+            $("#userOrgs .selector-show").hide();
+            self.onLoaded(subjectId, true);
+            self.filterOrgs(self.getHereBelowOrgs());
+            self.morphPatientOrgs();
+            $(".noOrg-container, .noOrg-container *").show();
+        }
+        $("#clinics").attr("loaded", true);
+        callback();
+    };
+    OrgTool.prototype.handleOrgsEvent = function(userId, isConsentWithTopLevelOrg) {
```
ivan-c: Would be really cool to see all the org code as components :D

achen2401: yeah, I agree, there are others too that can be refactored into components.  These are a bit more involved changes though. I will do them when I have downtime again.

portal/static/js/src/modules/Consent.js
```diff
@@ -125,6 +121,23 @@ export default { /*global i18next datepicker $*/
             });
         });
     },
+    validateConsentDateFields: function(d, h, m, s) {
+        var errorMessage = "";
```
ivan-c: I might use an array/list so you can decide how it's rendered at the end (eg concatenate for HTML, keep raw for JSON)

PR #2965, opened 2019-01-30
TN-1847 add QBT.id to queries ordered by QBT.at, due to collisions

pbugni:
This is very likely the source of the problem reported in https://jira.movember.com/browse/TN-1847  (can't reproduce to confirm, but I believe all of the queries adjusted *should* have this fix).

### DIFF
--- a/portal/models/qb_timeline.py
+++ b/portal/models/qb_timeline.py
@@ -533,7 +533,7 @@ def qb_status_visit_name(user_id, as_of_date):
     #  `due` == `completed`, but the row insertion defines priority
     qbt = QBT.query.filter(QBT.user_id == user_id).filter(
         QBT.at <= as_of_date).order_by(
-        QBT.at.desc(), QBT.id.desc()).limit(1).first()
+        QBT.at.desc(), QBT.id.desc()).first()
     if qbt:
         return qbt.status, visit_name(qbt.qbd())
     return OverallStatus.expired, None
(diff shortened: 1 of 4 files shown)


portal/models/qb_timeline.py
```diff
@@ -528,8 +528,12 @@ def qb_status_visit_name(user_id, as_of_date):
     # should be cached, unless recently invalidated
     update_users_QBT(user_id)
 
+    # We order by at (to get the latest status for a given QB) and
+    # secondly by id, as on rare occasions, the time (`at`) of
+    #  `due` == `completed`, but the row insertion defines priority
     qbt = QBT.query.filter(QBT.user_id == user_id).filter(
-        QBT.at <= as_of_date).order_by(QBT.at.desc()).limit(1).first()
+        QBT.at <= as_of_date).order_by(
+        QBT.at.desc(), QBT.id.desc()).limit(1).first()
```
ivan-c: why do you need to `limit(1)` when you're only interested in `first()` anyway?

pbugni: That's not new.  I believe without the `first()`, you simply have a query (including the limit clause), but until you ask for `{first(), one(), all()}` or use it as an iterator, you haven't executed the query.

pbugni: Heh - I see I justified the `first()`, not the `limit(1)`, which you asked about.

Good point, we don't need both.  I'll correct.

PR #2967, opened 2019-01-31
Refactor docker logic to debian packaging

ivan-c:
* Move portal package dependencies into debian
* Add `postinst` script
  * Move steps (in Dockerfile) after package installation to post-install script
* Do not upgrade packages included in base image; rely on base image updates


docker/Dockerfile
```diff
@@ -5,70 +5,40 @@ ENV \
...
-        ca-certificates \
-        libpq5 \
-        python2.7 \
-        wait-for-it && \
-    apt-get clean --quiet
+# Install locally from $ARTIFACT_DIR by default
+ARG debian_sources_entry="deb file:${ARTIFACT_DIR} ./"
 
-ARG debian_repo="https://uwcirg.jfrog.io/uwcirg/portal-deb"
+RUN echo $debian_sources_entry > /etc/apt/sources.list.d/truenth.list
```
pbugni: probably for legacy reasons, but why name the sources "truenth.list" when building the `portal` package?

jhowe-uw: /etc/apt/sources.list.d/* entries must end in *.list to be read by apt/dselect.

PR #2968, opened 2019-01-31
TN-1613 Enforce unique user_identifiers on restricted systems.

pbugni:
Don't merge yet - pending @achen2401 sub-task TN-1853

### DIFF
--- a/portal/models/identifier.py
+++ b/portal/models/identifier.py
@@ -1,13 +1,17 @@
 """Identifier Model Module"""
 from builtins import str
+import json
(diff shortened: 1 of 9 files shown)


achen2401:
@pbugni paul, I added the frontend changes for this.  let me know if you see any issue.  Besides Study ID field, I also added unique identifier check for Site ID field (that is the field  in Truenth USA), since that is also one of the identifier fields, but I don't know if that field is in use at all (and for what purose)?

PR #2972, opened 2019-02-01
TN-1832 creating audit trail for generating questionnaire status reports

achen2401:
see https://jira.movember.com/browse/TN-1754
subtask: https://jira.movember.com/browse/TN-1832

Need to leave audit trail for when a questionnaire status reporting is generated (for later analytics -whether a report has been generated by user(s))

### DIFF
--- a/portal/views/reporting.py
+++ b/portal/views/reporting.py
@@ -22,6 +22,7 @@ from flask_babel import gettext as _
 from flask_user import roles_required
 from werkzeug.exceptions import Unauthorized
 
+from ..audit import auditable_event
 from ..date_tools import FHIR_datetime
 from ..extensions import oauth
 from ..models.fhir import bundle_results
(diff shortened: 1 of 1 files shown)


mcjustin:
> Looks fine, but can't they do this with piwik (@mcjustin)?
@ivan-c I don't recall the source of this story... @achen2401  who suggested this?
Stepping back a bit, our audit log was a bit under-spec'd when we designed it 3+ years ago. Medical records guidelines would say that we should be able to easily generate a log file indicating all CRUD of patient records, whereas our audit log is only for C(not R)UD (hobbling together a mix of audit and apache logs sounds a bit tough, but adding in piwik sounds like a major pain). For this particular story, the data are unidentified, but the service has the ability to export all subjects' data. Movember doesn't have access to the audit logs, nor should they. Anyhoo, let's discuss a bit next time around.

achen2401:
@mcjustin Justin, this is a comment from Diana from this story:  https://jira.movember.com/browse/TN-1754
regarding analytics for generating report - exact quote from Diana, 

> are we able to get any analytics on how many times the patient list has been exported and in what format? It would be good to know if this feature is being used or not so we can reassess the features on the patient list.

Don't think we will be giving them audit log access directly but we will be querying the log to get the analytics she needs.  But sounds like querying audit log is not the best approach?  I think I will close this PR for now.   but as Ivan suggested previously, will we be able to get the stats via piwik?

PR #2983, opened 2019-02-06
TN-1824 migration to patch incorrect QB (wrong RP) on several QNRs.

pbugni:
The run time is down to about 30 seconds.

### DIFF
--- a/portal/migrations/versions/55469bdd181f_.py
+++ b/portal/migrations/versions/55469bdd181f_.py
@@ -1,0 +1,111 @@
+"""Correct QNR qb_id when assigned to the incorrect research protocol
+
+Revision ID: 55469bdd181f
+Revises: 9026af5fe040
+Create Date: 2019-02-04 13:36:04.709868
+
+"""
+from portal.database import db
+from portal.models.qb_status import QB_Status
+from portal.models.qb_timeline import QBT, invalidate_users_QBT
+from portal.models.questionnaire_bank import QuestionnaireBank
+from portal.models.questionnaire_response import QuestionnaireResponse
+from portal.models.research_protocol import ResearchProtocol
+from portal.models.user import User, active_patients
+from portal.models.overall_status import OverallStatus
+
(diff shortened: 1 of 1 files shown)


portal/migrations/versions/55469bdd181f_.py
```diff
@@ -0,0 +1,111 @@
...
+        rp3_name = "IRONMAN_v3" + qb.name[len('IRONMAN'):]
+        rp3_qb = QuestionnaireBank.query.filter_by(name=rp3_name).one()
+        assert (rp3_qb.research_protocol_id == rp3_id)
+        qb_map[qb.id] = rp3_qb.id
+    return qb_map
+
+
+def upgrade():
+    # Tricky task to locate the errors.  One time expensive process,
+    # loop through patients identifying problems.
```
ivan-c: >Per email "RE: 170-99-003" 2018-01-17 to truenth-dev


If it's a specific problem, can you give a brief description? Not sure the exact issue after looking at referenced email...

pbugni: Added detailed description to the jira issue.

portal/migrations/versions/55469bdd181f_.py
```diff
@@ -0,0 +1,111 @@
...
+                    expected_qb = QuestionnaireBank.query.get(
+                        qb_status._current_indef.qb_id)
+
+                if recorded_qb != expected_qb:
+                    replacement = qb_map.get(recorded_qb.id)
+                    if replacement == expected_qb.id:
+                        # retain the change request, as altering it inline
+                        # interrupts the iteration
+                        fix_map[qnr.id] = expected_qb.id
+                        print(
```
ivan-c: Could you make this an audit log entry instead?

pbugni: funny you should ask!  i had the same thought - see latest commit.

PR #2986, opened 2019-02-06
Fixup project/product config

ivan-c:
* Move env var configs to flask configs
* Move defaults to `config.py`
* Use TrueNTH USA as default config (`PERSISTENCE_DIR`)

NB: `PERSISTENCE_DIR=eproms` is required to maintain an ePROMs configuration


ivan-c:
Tested with both `eproms` and `gil` for `PERSISTENCE_DIR`

PR #2987, opened 2019-02-06
Feature/webpack package babel config fix

achen2401:
noting issues in building JS files when adding code introduced in ES2017 (e.g. dynamic import, async/wait)
Fixes include:
correct the relevant packages included in package.json
adding path in webpack config file to ensure dynamically generated modules will land in the correct location


bin/nodejs-wrapper.sh
```diff
@@ -74,6 +74,10 @@ setup_node_venv "" "$node_venv"
 echo "Activating NodeJS virtual environment..."
 . "${node_venv}/bin/activate"
 
+#removing unused packages not listed in the dependencies list in package.json - this is useful because some package(s) can be removed
+echo "Pruning unused packages..." 
+npm --prefix "${repo_root}/portal" prune --production=false
```
ivan-c: I'd like to keep this script as simple as possible, and unless they're causing problems, I don't think we need to bother cleaning them up. When we build the portal docker images, we're starting from scratch every time, so we don't ever have to worry about previously installed packages.

However, if this is to make your dev ergonomics/process easier, I'm 100% on board, but can I add a flag to `nodejs-wrapper.sh` (eg `nodejs-wrapper.sh -p`) so `prune` is only called if the flag is passed?

PS It would be an extra command to run (and you may already know this :), but I think you should be able to prune the packages installed by `nodejs-wrapper.sh` by running `nodejs-wrapper.sh npm --prefix ./portal prune --production=false`

achen2401: Sure, I will remove the command.  But just to make sure - when you mentioned “starting from scratch “, does that mean the modules in node_modules folder are wipes and re-installed each time? I am just concerned for the scenario, in this case, that JS files are built against syntax-dynamic-import module instead of @babel/plugin-syntax-dynamic-import (note the new name space).

achen2401: Ivan, thank you for the explanation.  Thinking about it a bit more - I don’t anticipate issues actually.. as in the babel config file (.babelrc) - we are specifying newer versions (ones with @babel namespace) of plugins for building (duh).  Sorry about that, wasn’t thinking right.

PR #2993, opened 2019-02-11
Feature/assessment status purge

pbugni:
Adding the common `purge=True` query string option for `/api/patient/<int:patient_id>/assessment-status`

(also took a moment to kill url_prefix in this module for improved grep results)

### DIFF
--- a/portal/views/assessment_engine.py
+++ b/portal/views/assessment_engine.py
@@ -1809,7 +1809,7 @@ def patient_assessment_status(patient_id):
             "BEGIN trace for assessment-status on {}".format(patient_id))
 
     purge = request.args.get('purge', False)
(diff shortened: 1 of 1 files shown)


portal/views/assessment_engine.py
```diff
@@ -1801,6 +1807,10 @@ def patient_assessment_status(patient_id):
     if trace:
         establish_trace(
             "BEGIN trace for assessment-status on {}".format(patient_id))
+
+    purge = request.args.get('purge', False)
+    if purge == 'True':
```
ivan-c: would prefer a case-insensitive check
```suggestion
    if purge.lower() == 'true':
```

PR #2994, opened 2019-02-12
TN-1839 enforce unique constraint on QNR.identifier

pbugni:
Post review, this should now be ready to merge.

@ivan-c you should probably also check out the comments in TN-1882 detailing the issues found from the respective production dbs.

### DIFF
--- a/portal/migrations/versions/a1de3ab4d050_.py
+++ b/portal/migrations/versions/a1de3ab4d050_.py
@@ -12,8 +12,7 @@ from portal.dict_tools import dict_compare
 from portal.models.audit import Audit
 from portal.models.identifier import Identifier
 from portal.models.questionnaire_response import QuestionnaireResponse
-from portal.models.role import Role, ROLE
-from portal.models.user import User, UserRoles
+from portal.models.user import User
 
 # revision identifiers, used by Alembic.
 revision = 'a1de3ab4d050'
@@ -21,18 +20,13 @@ down_revision = '55469bdd181f'
 
 
 def admin_id():
-    """Look up an admin ID"""
-    admin = User.query.filter_by(email='bob25mary@gmail.com').first()
-    admin = admin or User.query.join(UserRoles).join(Role).filter(
-        Role.id == UserRoles.role_id).filter(
-        UserRoles.user_id == User.id).filter(
-        Role.name == ROLE.ADMIN.value).first()
-    return admin.id
+    sys = User.query.filter_by(email='__system__').first()
+    return sys.id
 
 
 def diff_docs(doc1, doc2):
     """Print details of two differing QNR documents"""
-    added, removed, modified, same = dict_compare(doc1, doc2)
+    added, removed, modified, _ = dict_compare(doc1, doc2)
     assert not added
     assert not removed
     assert not set(modified) - set(['authored', 'group'])
--- a/portal/migrations/versions/a1de3ab4d050_.py
+++ b/portal/migrations/versions/a1de3ab4d050_.py
@@ -1,0 +1,117 @@
+"""Remove duplicate QNR.identifiers
+
+Revision ID: a1de3ab4d050
+Revises: 55469bdd181f
+Create Date: 2019-02-11 16:48:55.332527
+
+"""
+import json
+from sqlalchemy import func
+from portal.database import db
+from portal.dict_tools import dict_compare
+from portal.models.audit import Audit
+from portal.models.identifier import Identifier
+from portal.models.questionnaire_response import QuestionnaireResponse
+from portal.models.role import Role, ROLE
+from portal.models.user import User, UserRoles
+
+# revision identifiers, used by Alembic.
+revision = 'a1de3ab4d050'
+down_revision = '55469bdd181f'
+
+
+def admin_id():
+    """Look up an admin ID"""
+    admin = User.query.filter_by(email='bob25mary@gmail.com').first()
+    admin = admin or User.query.join(UserRoles).join(Role).filter(
+        Role.id == UserRoles.role_id).filter(
+        UserRoles.user_id == User.id).filter(
+        Role.name == ROLE.ADMIN.value).first()
+    return admin.id
+
+
+def diff_docs(doc1, doc2):
+    """Print details of two differing QNR documents"""
+    added, removed, modified, same = dict_compare(doc1, doc2)
+    assert not added
+    assert not removed
+    assert not set(modified) - set(['authored', 'group'])
+    if modified:
+        if len(doc1['group']) != len(doc2['group']):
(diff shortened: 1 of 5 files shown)


portal/views/assessment_engine.py
```diff
@@ -847,14 +848,20 @@ def assessment_update(patient_id):
...
+    except (TypeError, KeyError):
+        abort(400, "Ill formed or missing 'identifier', see FHIR spec")
     try:
         existing_qnr = QuestionnaireResponse.query.filter(
-            QuestionnaireResponse.document["identifier"]
-            == updated_qnr["identifier"]
-        ).one()
-    # except NoResultException:
+            QuestionnaireResponse.document['identifier']['system']
+            == json.dumps(identifier.system)).filter(
```
ivan-c: Could you move this `filter(` to its own line so it's obvious it's chained off the previous `filter()`?

portal/views/assessment_engine.py
```diff
@@ -1345,6 +1359,24 @@ def assessment_add(patient_id):
...
+    if 'identifier' in request.json:
+        # Confirm it's unique, or raise 409
+        try:
+            identifier = Identifier.from_fhir(request.json['identifier'])
+        except (TypeError, KeyError):
+            abort(400, "Ill formed 'identifier', see FHIR spec")
+
+        existing_qnr = QuestionnaireResponse.query.filter(
+            QuestionnaireResponse.document['identifier']['system']
+            == json.dumps(identifier.system)).filter(
```
ivan-c: See above

tests/test_assessment_engine.py
```diff
@@ -55,6 +56,46 @@ def test_submit_invalid_assessment(self):
...
+        identifier = Identifier(system='https://unique.org', value='abc123')
+        data = swagger_spec['definitions']['QuestionnaireResponse']['example']
+        data['identifier'] = identifier.as_fhir()
+
+        self.promote_user(role_name=ROLE.PATIENT.value)
+        self.login()
+        response = self.client.post(
+            '/api/patient/{}/assessment'.format(TEST_USER_ID),
+            content_type='application/json',
+            data=json.dumps(data),
```
ivan-c: You can pass JSON data (and omit the `content_type='application/json'` header) with `json=data`

[See](http://flask.pocoo.org/docs/1.0/changelog/#version-1-0)
>Add json keyword argument for the test client request methods. This will dump the given object as JSON and set the appropriate content type

portal/views/assessment_engine.py
```diff
@@ -847,14 +848,20 @@ def assessment_update(patient_id):
             'valid': True,
         })
 
-    # Todo: enforce identifier uniqueness at initial submission
+    try:
+        identifier = Identifier.from_fhir(updated_qnr.get('identifier'))
+    except (TypeError, KeyError):
+        abort(400, "Ill formed or missing 'identifier', see FHIR spec")
```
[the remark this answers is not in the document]
pbugni: @ivan-c  Not on the PUT view, as I understand it.  Don't we require an identifier to be able to update (PUT) one?

portal/views/assessment_engine.py
pbugni: those are all over the code base, easy to parse once you get used to reading them :)

tests/test_assessment_engine.py
pbugni: thanks - neat enhancement

portal/models/identifier.py
```diff
@@ -40,6 +40,11 @@ def value(self, value):
 
     @classmethod
     def from_fhir(cls, data):
+        if not data or not all((data.get('system'), data.get('value'))):
+            raise ValueError(
+                "Ill formed 'identifier'; requires both 'system' and "
```
ivan-c: I don't think this requirement is strictly necessary (resources that break it may still be FHIR compliant).

Do we have partial QuestionnaireResponses lacking `identifier`s?

portal/views/assessment_engine.py
```diff
@@ -847,17 +848,24 @@ def assessment_update(patient_id):
...
+        current_app.logger.warning(
+            "attempted update on QuestionnaireResponse with unknown "
+            "identifier {}".format(identifier))
         abort(404, "existing QuestionnaireResponse not found")
+    if len(existing_qnr) > 1:
+        msg = ("can't update; multiple QuestionnaireResponses found with "
+               "identifier {}".format(identifier))
+        current_app.logger.warning(msg)
+        abort(409, msg)
     else:
```
ivan-c: I think this `else` is pointless if the `if` exits (`abort()`s)

tests/test_assessment_engine.py
```diff
@@ -207,15 +240,11 @@ def test_assessments_bundle(self):
...
-            data=json.dumps(example_data),
-        )
+            json=example_data)
         assert upload.status_code == 200
 
         response = self.client.get(
-            '/api/patient/assessment?instrument_id={}'.format(instrument_id),
-            content_type='application/json',
-        )
+            '/api/patient/assessment?instrument_id={}'.format(instrument_id))
```
ivan-c:
```suggestion
            '/api/patient/assessment', query_string={'instrument_id':instrument_id})
```

tests/test_assessment_engine.py
```diff
@@ -104,10 +152,7 @@ def test_submit_assessment_for_qb(self):
         self.promote_user(role_name=ROLE.PATIENT.value)
         self.login()
         response = self.client.post(
-            '/api/patient/{}/assessment'.format(TEST_USER_ID),
-            content_type='application/json',
-            data=json.dumps(data),
-        )
+            '/api/patient/{}/assessment'.format(TEST_USER_ID), json=data)
```
ivan-c: :1st_place_medal:

portal/migrations/versions/a1de3ab4d050_.py
```diff
@@ -0,0 +1,118 @@
...
+from portal.models.user import User, UserRoles
+
+# revision identifiers, used by Alembic.
+revision = 'a1de3ab4d050'
+down_revision = '55469bdd181f'
+
+
+def admin_id():
+    """Look up an admin ID"""
+    admin = User.query.filter_by(email='bob25mary@gmail.com').first()
```
ivan-c: Can we use the `__system__` email for migrations?

portal/models/identifier.py
pbugni: No partials.  Just all the submissions from WiserCare (where they're submitting the patient identifier w/o a system).

portal/migrations/versions/a1de3ab4d050_.py
pbugni: oh!  forgot about that option.  thanks.

PR #2997, opened 2019-02-12
initial queries - fix initial blank page display on loading

achen2401:
found this issue while testing another:  at Initial Queries during registration, one was presented with a blank page before content faded in.
Fix includes:
delay fading out of loading indicator (by adding callback class on document body to indicate as such) to allow loading of body content on callback
create a Utility dedicated function for delayed removing loading indicator

### DIFF
--- a/portal/templates/initial_queries.html
+++ b/portal/templates/initial_queries.html
@@ -1,7 +1,7 @@
 {% extends "layout.html" %}
+{%- block mainclass -%}class="vis-on-callback"{%- endblock -%}
 {% block main %}
 {% from "flask_user/_macros.html" import back_btn %}
-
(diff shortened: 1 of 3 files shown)


portal/templates/initial_queries.html
```diff
@@ -1,7 +1,7 @@
 {% extends "layout.html" %}
+{%- block mainclass -%}class="vis-on-callback"{%- endblock -%}
```
ivan-c: Would it make sense to use the [VueJS `data` object](https://vuejs.org/v2/guide/state-management.html#Simple-State-Management-from-Scratch) to store this state instead of the DOM, or would that be overkill (eg VueJS not used in this page)?

achen2401: yeah, that is a good idea.  but you are right VueJS is not used on this page yet but I plan on converting it to using it if I have downtime.

PR #3001, opened 2019-02-13
Feature/pep479

pbugni:
Without these changes, can't run tests or flask on python 3.7

See [pep479](https://www.python.org/dev/peps/pep-0479) for the details.


pbugni:
Turns out `celery` isn't python 3.7 compatible just yet:  https://github.com/celery/celery/issues/5049#issuecomment-420921636
Rolling back to fix the failing tests.

PR #3021, opened 2019-02-27
Update default (development) Liferay config values

ivan-c:
Retire old stg-lr7 domain in favor of cms-stage

### DIFF
--- a/portal/config/config.py
+++ b/portal/config/config.py
@@ -162,9 +162,9 @@ class BaseConfig(object):
     DEFAULT_LOCALE = 'en_US'
     FILE_UPLOAD_DIR = os.environ.get('FILE_UPLOAD_DIR', 'uploads')
 
-    LR_ORIGIN = os.environ.get('LR_ORIGIN', 'https://stg-lr7.us.truenth.org')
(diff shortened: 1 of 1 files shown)


portal/config/config.py
```diff
@@ -162,9 +162,9 @@ class BaseConfig(object):
     DEFAULT_LOCALE = 'en_US'
     FILE_UPLOAD_DIR = os.environ.get('FILE_UPLOAD_DIR', 'uploads')
 
-    LR_ORIGIN = os.environ.get('LR_ORIGIN', 'https://stg-lr7.us.truenth.org')
-    LR_GROUP = os.environ.get('LR_GROUP', 20145)
-    LR_FOLDER_ST = os.environ.get('LR_FOLDER_ST', 35564)
+    LR_ORIGIN = os.environ.get('LR_ORIGIN', 'https://cms-stage.us.truenth.org')
```
pbugni: does `stage` have a different meaning in the context of LifeRay?  I would expect this default value to point to a LR development server?

ivan-c: I think it may have something to do with the approval/migration process for content. @victorlima02 ?

PR #3022, opened 2019-02-27
TN-1888-TN-1893-new-orgs-protocol

mcjustin:
TN-1888-TN-1893-new-orgs-protocol explicitly set to v3 (note that these were never at v2, except for on dev&test (errantly) )

### DIFF
--- a/portal/config/eproms/Organization.json
+++ b/portal/config/eproms/Organization.json
@@ -249,14 +249,6 @@
       "resourceType": "Organization"
     },
     {
-      "extension": [
-        {
-          "research_protocols": [
-            {"name": "IRONMAN v2"}
-          ],
-          "url": "http://us.truenth.org/identity-codes/research-protocol"
-        }
-      ],
       "id": 20000,
       "identifier": [
         {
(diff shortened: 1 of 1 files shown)


mcjustin:
@pbugni  note that this is a precursor to the changes that were were chatting about (simplifying protocols in the org tree per https://jira.movember.com/browse/TN-1913 )

mcjustin:
mcjustin [4:01 PM]
those are mostly tngr
... which doesn't have protocols anywhere up or down the tree

pbugni [4:02 PM]
okay

mcjustin [4:02 PM]
These two are fake orgs, have been afraid to remove them because we likely have (non-real) users associated w/ them
No RPs for 21220:AU B Child Site 2
No RPs for 21210:AU B Child Site 1

pbugni [4:05 PM]
if they’ve outlived their usefulness, lets purge (but that can wait)

mcjustin [4:05 PM]
only other one is `No RPs for 30100:Southampton` - that's our single "TrueNTH Global" org - like TNGR, no protocols there

mcjustin [4:05 PM]
replied to a thread:
seems like that would be a mess...

those 2 *are* ironman orgs - want me to assign protocols to them?

pbugni [4:09 PM]
if you want the patients to be able to take assessments, yes :slightly_smiling_face: (edited) 

mcjustin [4:09 PM]
I don't care about those patients

pbugni [4:09 PM]
okay, then we can merge this

PR #3032, opened 2019-03-06
pull base requirements.txt file as dev dependency.

pbugni:
Noticed this trick in another example.  Should mean any devs keeping their local builds up via `pip` are down to a single step: `pip install -r requirements.dev.txt`

### DIFF
--- a/requirements.dev.txt
+++ b/requirements.dev.txt
@@ -1,4 +1,4 @@
--r requirements.txt
+--requirement requirements.txt
 
 alabaster==0.7.12         # via sphinx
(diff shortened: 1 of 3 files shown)


requirements.dev.txt
```diff
@@ -1,3 +1,5 @@
+-r requirements.txt
```
ivan-c: Can you use the long-option to match what's in the README?
I tend to prefer long-options (if available) in scripts because they can be self-documenting.

```suggestion
--requirement requirements.txt
```

pbugni: :+1:

PR #3041, opened 2019-03-08
TN-1835 RCT patients

pbugni:
Rework `/patients` logic, for more predictable consumption by admins, intervention_staff and staff.
See #TN-1835 for details.

### DIFF
--- a/portal/models/intervention.py
+++ b/portal/models/intervention.py
@@ -106,6 +106,20 @@ class Intervention(db.Model):
 
         return d
 
+    @staticmethod
+    def rct_ids():
+        """returns list of RCT (randomized control trial) intervention ids"""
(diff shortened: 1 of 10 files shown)


portal/models/intervention.py
```diff
@@ -106,6 +106,20 @@ def as_json(self):
 
         return d
 
+    @staticmethod
+    def rct_ids():
+        """returns list of RCT (randomized control trial) intervention ids"""
+        names = current_app.config.get('RCT_INTERVENTIONS')
+        if not names:
+            return None
+        ids = [i.id for i in Intervention.query.filter(
```
ivan-c: Would you be able to just return the `Intervention` `Query`?

You could `filter()` against `~rct_interventions`,  then you wouldn't have to look up the intervention twice (looking up the ids, then filtering on them).

PR #3048, opened 2019-03-12
TN-1935-IRONMAN-TEST-ORG

mcjustin:
https://jira.movember.com/browse/TN-1935


mcjustin:
> 
> 
> Looks good- NB this will need to branch from `master` instead if we need to update any prod/demo sites before the next release though

@ivan-c  thanks for flagging. They specifically said that they don't need this on prod. Wouldn't be surprised if they do eventually, but next usual release would be soon enough.

PR #3062, opened 2019-03-22
TN-1945 Psa Tracker Graph bug fixes

achen2401:
https://jira.movember.com/browse/TN-1945
address the issue where data points were drawn beyond visual area.
changes include: 
- since the graph is drawn in log scale, need to account for value(s), e.g. < 0.1 that caused invalid log value  - in which case use the minimum allowable log value (0.1)
- grid line needs to be capped at maximum allowable range for PSA tracker values (10000) (otherwise seeing extra line drawn)
- allow edit of treatment date only if it is possible (the section is only shown in profile, where treatment is edited, in Truenth USA)

### DIFF
--- a/portal/static/js/src/components/PsaTracker.vue
+++ b/portal/static/js/src/components/PsaTracker.vue
@@ -455,7 +455,7 @@
                 this.postData();
             },
             checkTreatmentCoreData: function() {
-                tnthAjax.getConfigurationByKey("REQUIRED_CORE_DATA",false, function(data) {
+                tnthAjax.getConfigurationByKey("REQUIRED_CORE_DATA", false, function(data) {
                     if (!data || !data.REQUIRED_CORE_DATA) {
                         return;
                     }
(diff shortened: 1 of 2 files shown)


portal/static/js/src/components/PsaTracker.vue
```diff
@@ -451,6 +454,16 @@
                 }
                 this.postData();
             },
+            checkTreatmentCoreData: function() {
+                tnthAjax.getConfigurationByKey("REQUIRED_CORE_DATA",false, function(data) {
+                    if (!data || !data.REQUIRED_CORE_DATA) {
+                        return;
+                    }
+                    if (data.REQUIRED_CORE_DATA.indexOf("treatment") === -1) {
+                        $("#psaTrackerTreatmentContainer").hide(); //do not allow edit of treatment date if it is not a required core data, because, it is only editable in profile for Truenth where treatment is required
```
ivan-c: I might put the comment on its own line when it's this long (easier to read)...

portal/static/js/src/components/PsaTracker.vue
```diff
@@ -451,6 +454,16 @@
                 }
                 this.postData();
             },
+            checkTreatmentCoreData: function() {
+                tnthAjax.getConfigurationByKey("REQUIRED_CORE_DATA",false, function(data) {
```
ivan-c:
```suggestion
                tnthAjax.getConfigurationByKey("REQUIRED_CORE_DATA", false, function(data) {
```

PR #3067, opened 2019-03-26
Simplify docker-compose syntax

ivan-c:
* Remove shell-style defaults (see below) from docker-compose files
  * Shell-style defaults allow overrides via `.env` and `portal.env` (env files)
  * Now only overrideable via `portal.env`
* Still necessary for some settings that cannot be set from `portal.env`
  * eg `volumes`, `ports`

Shell-style defaults:
```
${FOO:-default value of FOO if not set}
```


ivan-c:
@pbugni 
>Does this mean variable w/o the shell override syntax, defined in .env will be ignored?
Correct, "shell-style variables" (eg `${FOO}`) are required if you want to change values in `docker-compose.yaml` with environment variables specified in the `.env` file


>Do our docs need any updates?
No, we shouldn't need to. I've never had to override any of those defaults, and I can't easily think of a reason to

PR #3070, opened 2019-03-26
TN-1961 hide account creation button from MUSIC staff on patients' list

achen2401:
https://jira.movember.com/browse/TN-1961
Changes include:
hide account creation button on patients list from MUSIC staff
will **not** hide the button for staff that has top org affiliation other than MUSIC

### DIFF
--- a/portal/static/js/src/admin.js
+++ b/portal/static/js/src/admin.js
@@ -466,25 +466,47 @@ import CurrentUser from "./mixins/CurrentUser.js";
                 $("#adminTable .filterControl input").attr("placeholder", i18next.t("Enter Text"));
                 $("#adminTable .filterControl select option[value='']").text(i18next.t("Select"));
             },
-            handleMedidataRave: function (params) {
-                if (!$("#adminTableContainer").hasClass("patient-view")) { //check if this is a patients list
+            isPatientsList: function() {
+                return $("#adminTableContainer").hasClass("patient-view");//check if this is a patients list
+            },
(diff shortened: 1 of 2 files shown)


portal/static/js/src/admin.js
```diff
@@ -466,26 +466,48 @@ import CurrentUser from "./mixins/CurrentUser.js";
                 $("#adminTable .filterControl input").attr("placeholder", i18next.t("Enter Text"));
                 $("#adminTable .filterControl select option[value='']").text(i18next.t("Select"));
             },
-            handleMedidataRave: function (params) {
-                if (!$("#adminTableContainer").hasClass("patient-view")) { //check if this is a patients list
+            isPatientsList: function() {
+                return $("#adminTableContainer").hasClass("patient-view");//check if this is a patients list
+            },
+            setCreateAccountVisByTopOrgSetting: function(setting_name, params) {
```
ivan-c: Can you add a comment about what this generally does?

portal/static/js/src/admin.js
```diff
@@ -498,7 +520,10 @@ import CurrentUser from "./mixins/CurrentUser.js";
                 if (this.isAdminUser()) {
                     return false;
                 }
-                this.handleMedidataRave(); //a function specifically created to handle MedidataRave related stuff
+                //a function specifically created to handle MedidataRave related stuff
```
ivan-c: Can you move these comments (definitely appreciated) into the body of each function, maybe add a bit of detail about the business need?

PR #3073, opened 2019-03-27
Refactor to multistage builds

ivan-c:
* Refactor from [builder pattern to multi-stage builds](https://docs.docker.com/develop/develop-images/multistage-build/)
  * Remove shell scripting/docker-compose orchestration
  * Use `COPY` instead of running `git clone` (at run-time)
    * Generate `.dockerignore` during build from `.gitignore`
* Use `dpkg-buildpackage` hooks
* Update docs


bin/docker-build.sh
```diff
@@ -33,15 +32,23 @@ cp \
     "${root_path}/docker/portal.env.default" \
     "${root_path}/docker/portal.env"
 
-default_compose_file="${root_path}/docker/docker-compose.yaml:${root_path}/docker/docker-compose.build.yaml"
+# Use .gitignore as .dockerignore during build only
```
pbugni: can this not be managed w/ symlinks?

docker/Dockerfile
```diff
@@ -1,4 +1,42 @@
...
+# Squelch superfluous output; apt-get has poor verbosity control
+RUN \
+    apt-get update --quiet > /dev/null && \
+    apt-get install --quiet --quiet --no-install-recommends \
+        devscripts \
+        equivs \
+        git | grep "Setting up" | awk '{print $3 $4}' ORS=' '
+        # only print package names, versions
+
+WORKDIR /root/portal
```
pbugni: Probably a docker standard, but why `/root`?  I'd favor a POSIX structure...

bin/docker-build.sh
ivan-c: Good idea, but unfortunately doesn't work very well in practice.

Ideally we'd commit a symlink (git can track symlinks correctly- [example](https://github.com/ivan-c/keychain-castle/blob/master/home/bin/keychain)) and commit it.
However, git does not honor (actually create) symlinks when you clone a new repo, unless you override a default config (`git config --local core.symlinks true`).

I'd be OK using a symlink (or hardlink) at run-time, but I thought it might be useful to be able to override the generated `.dockerignore` (eg occasionally evaluating if we do need a separate `.dockerignore` file).
Would be happy to make it less "configurable" if we realize `.gitignore` is indeed good enough long-term

docker/Dockerfile
ivan-c: An odd quirk of the `root` user- their [`$HOME` is actually `/root`](http://www.linfo.org/slash_root.html)

I'll try using the `$HOME` env var (ie `$HOME/portal`), which would be a little more flexible (if we wanted to tighten up security and not run as root)

ivan-c: That ended up working, nice improvement!

bin/docker-build.sh
pbugni: as you see fit.  i think too much configuration can be a bad thing, and would vote for the simplicity of a link, however you chose to enable.  we can always change, once the real need for a separate file makes its case.

ivan-c: >too much configuration can be a bad thing

Totally agree. I just wish I knew what I know now- it can be really difficult to predict what should or shouldn't be configurable. I've tended to make things more configurable "just in case" and then remove it later but that might not be the best approach for everything.

PR #3076, opened 2019-03-27
Track Flask-User i18n Strings; Refactor

ivan-c:
* Add `flask_user.pot` to portal's tracked translations
  * Upload POT file and download corresponding PO files from Smartling
* Move i18n util code to `i18n_utils.py`


portal/models/i18n_utils.py
```diff
@@ -302,3 +308,177 @@ def compile_pos():
             "Saved MO file: %s",
             os.path.relpath(mo_filepath, current_app.root_path),
         )
+
+
+def upsert_to_template_file():
+    db_translatables = {}
+    db_translatables.update(get_db_strings())
```
pbugni: what's the value of creating an empty dict just to update it?  why not:
`db_translatables = get_db_strings()`

portal/models/i18n_utils.py
```diff
@@ -302,3 +308,177 @@ def compile_pos():
...
+    if not db_translatables:
+        current_app.logger.warn("no DB strings extracted")
+        return
```
pbugni: this looks like more than a `warn` if it causes premature exit??

portal/models/i18n_utils.py
```diff
@@ -302,3 +308,177 @@ def compile_pos():
...
+                    continue
+                for loc in locations:
+                    potlines.append("# " + loc + "\n")
+                potlines.append("msgid " + entry + "\n")
+                potlines.append("msgstr \"\"\n")
+                potlines.append("\n")
+            potfile.truncate(0)
+            potfile.seek(0)
+            potfile.writelines(potlines)
+    except:
```
pbugni: hurm, i take it, you don't know what exception type to catch?  wouldn't you want to also log the exception?

portal/models/i18n_utils.py
```diff
@@ -302,3 +308,177 @@ def compile_pos():
...
+        exceptionType, exceptionValue, exceptionTraceback = sys.exc_info()
+        sys.exit(
+            "Could not write to translation file!\n ->%s" % (exceptionValue))
+
+
+def fix_references(pot_fpath):
+    """Fix reference comments to remove checkout-specific paths"""
+    # Todo: override PoFileParser._process_comment() to perform this as part of `pybabel extract`
+
+    path_regex = re.compile(r"^#: {}(?P<rel_path>.*):(?P<line>\d+)".format(
```
pbugni: possibly a premature optimization, but why re-compile the regex every time in the loop?

portal/models/i18n_utils.py
ivan-c: I think at one point we didn't have DB strings to translate from TNUSA

portal/models/i18n_utils.py
ivan-c: I added the most likely ones, I'll improve the logging if I end up catching any

portal/models/i18n_utils.py
ivan-c: Not ideal, but I can't think of a good way to keep that function self-contained and not re-compile the regex every time.

Some kind of memoization to cache it? But I think I would need a separate function just to build the regex?

portal/models/i18n_utils.py
ivan-c: That function will likely be re-written at some point- it could be a lot simpler using `polib` instead of manipulating the raw file.

PR #3080, opened 2019-03-28
Migration to correct user_consent situation for deceased patients

pbugni:


portal/migrations/versions/0701782c564d_.py
```diff
@@ -0,0 +1,112 @@
...
+    if user_consents[-1].status == 'suspended' and not(
+            user_consents[-1].options & SEND_REMINDERS_MASK):
+        return
+
+    admin = User.query.filter_by(email='__system__').first()
+
+    log.info("Patching broken consents for {}, with deceased date {}".format(
+        user, dd))
+    keeper = user_consents.pop(0)
+    log.info(" keeping original user_consent {} {} {} {}".format(
```
ivan-c: Probably not worth it now, but if you end up having to correct these a lot, might be worth adding a `__repr__` method to `UserConsent`

PR #3091, opened 2019-04-01
i18n utility fixups

ivan-c:
* Fix upload condition check
* Fix unicode upload issues

### DIFF
--- a/portal/models/i18n.py
+++ b/portal/models/i18n.py
@@ -1,4 +1,5 @@
 """Module for i18n methods and functionality"""
+from __future__ import unicode_literals  # isort:skip
 from future import standard_library  # isort:skip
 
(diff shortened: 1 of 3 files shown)


portal/models/i18n.py
```diff
@@ -1,4 +1,5 @@
 """Module for i18n methods and functionality"""
+from __future__ import unicode_literals  # isort:skip
```
pbugni: There also appear to be a number of unused imports left behind from the utils refactor:
```
from io import BytesIO
import os
import sys
import tempfile

from polib import pofile
import requests```

ivan-c: Done, thanks! Looks like Codacy should've caught those- re-enabled it

PR #3104, opened 2019-04-02
TN-1975 always preserve order of QBT.at values inserting QB_Timeline

pbugni:
New data structure added to maintain order of list by attribute `at` field, named `AtOrderedList`.
The `append` method on `AtOrderedList` enforces order on insertion.
Use when building user's qb_timeline rows to prevent order errors, such as those demonstrated in TN-1975.

NB - no migration yet to seek out any qb_timeline rows needing to be reprocessed.  Waiting to see if another issue's migration during same sprint will cover.

### DIFF
--- a/portal/models/qb_timeline.py
+++ b/portal/models/qb_timeline.py
@@ -86,12 +86,12 @@ class AtOrderedList(list):
 
         """
         if not self.__len__():
-            return super().append(value)
+            return super(AtOrderedList, self).append(value)
 
         # Expecting to build in order; common case new value
         # lands at end.
         if self[-1].at <= value.at:
-            return super().append(value)
+            return super(AtOrderedList, self).append(value)
 
         # Otherwise, walk backwards till new value < existing
         for i, e in reversed(list(enumerate(self))):
--- a/portal/models/qb_timeline.py
+++ b/portal/models/qb_timeline.py
@@ -63,15 +63,17 @@ class QBT(db.Model):
 
 
 class AtOrderedList(list):
-    """Specialize ``list`` to maintain insertion order and ``at`` values
+    """Specialize ``list`` to maintain insertion order and ``at`` attribute
 
     When building up QBTs for a user, we need to maintain both order and
-    sort by ``QBT.at`` values.  As there are rows with identical ``at``
-    values, it can't simply be sorted by a field.
+    sort by the ``QBT.at`` attribute.  As there are rows with identical ``at``
+    values, it can't simply be sorted by a field, as insertion order matters
+    in the case of a tie (as say the ``due`` status needs to precede a
+    ``completed``, which does happen with paper entry).
 
     As the build order matters, continue to add QBTs to the end of the list,
     taking special care to insert those with earlier ``at`` values in
-    the correct place.  Two or more identicle ``at`` values should result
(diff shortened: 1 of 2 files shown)


portal/models/qb_timeline.py
```diff
@@ -62,6 +62,50 @@ def qbd(self):
...
+class AtOrderedList(list):
+    """Specialize ``list`` to maintain insertion order and ``at`` values
+
+    When building up QBTs for a user, we need to maintain both order and
+    sort by ``QBT.at`` values.  As there are rows with identical ``at``
+    values, it can't simply be sorted by a field.
+
+    As the build order matters, continue to add QBTs to the end of the list,
+    taking special care to insert those with earlier ``at`` values in
+    the correct place.  Two or more identicle ``at`` values should result
```
ivan-c:
```suggestion
    the correct place.  Two or more identical ``at`` values should result
```

portal/models/qb_timeline.py
```diff
@@ -62,6 +62,50 @@ def qbd(self):
             recur_id=self.qb_recur_id, qb_id=self.qb_id)
 
 
+class AtOrderedList(list):
+    """Specialize ``list`` to maintain insertion order and ``at`` values
```
ivan-c: `at values` seems too jargony. Something between "QB timeline snapshot creation datetime" and `at` would be ideal

tests/test_qb_timeline.py
```diff
@@ -420,3 +421,89 @@ def test_age(self):
...
+
+
+class HasAt(object):
+    """Mock the sorted portion of QBT for testing AtOrderedList"""
+    def __init__(self, at, label):
+        self.at = at
+        self.label = label
+
+
+n = HasAt(datetime.utcnow(), 'now')
```
ivan-c: Please pick an arbitrary datetime, [non-deterministic tests can be flakey](https://martinfowler.com/articles/nonDeterminism.html#Time)

tests/test_qb_timeline.py
```diff
@@ -420,3 +421,89 @@ def test_age(self):
...
+y = HasAt(n.at - relativedelta(days=1), 'yesterday')
+y2 = HasAt(y.at, 'yesterday again')
+wa = HasAt(n.at - relativedelta(weeks=1), 'week ago')
```
ivan-c: Please use longer variable names. Unless they get long enough to cause PEP8 nags, longer variable names are easier to grep/find/replace

portal/models/qb_timeline.py
```diff
@@ -370,7 +414,7 @@ def update_users_QBT(user_id, invalidate_existing=False):
 
         # As we move forward, capture state at each time point
 
-        pending_qbts = []
+        pending_qbts = AtOrderedList()
```
ivan-c: Why not just define a sorted "view" (function) of the list (eg `.keys()`, `.items()`, `.values()`)?

pbugni: tried to make this clear in the comment - we need to preserve insertion order AND then by `at`.  Using a built in sorted view won't retain this need.

pbugni:
@ivan-c said:
> Had some code style nags, ready to merge otherwise

Committed changes to address.  Please review AYC

PR #3106, opened 2019-04-03
TN-1976 maintain QNR questionnaire bank association on state change

pbugni:
TN-1969 is now in question, at least by me.  I'd like to see this change go out regardless.

### DIFF
--- a/portal/models/questionnaire_response.py
+++ b/portal/models/questionnaire_response.py
@@ -12,7 +12,9 @@ from sqlalchemy.dialects.postgresql import ENUM, JSONB
 from ..database import db
 from ..date_tools import FHIR_datetime
 from ..system_uri import TRUENTH_EXTERNAL_STUDY_SYSTEM
+from .audit import Audit
 from .fhir import bundle_results
+from .questionnaire import Questionnaire
 from .reference import Reference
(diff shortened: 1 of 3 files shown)


portal/models/questionnaire_response.py
```diff
@@ -61,6 +63,65 @@ def __str__(self):
...
+        # clear both until current values are determined
+        self.questionnaire_bank_id, self.qb_iteration = None, None
+
+        authored = FHIR_datetime.parse(self.document['authored'])
+        qn_ref = self.document.get("questionnaire").get("reference")
+        qn_name = qn_ref.split("/")[-1] if qn_ref else None
+        qn = Questionnaire.find_by_name(name=qn_name)
+        qbstatus = QB_Status(self.subject, as_of_date=authored)
+        qbd = qbstatus.current_qbd()
+        if qbd and qn and qn.id in [
```
ivan-c: I try to use tuples or generator comprehensions if I don't have any need for a list. Probably a premature optimization (little less memory), but I don't think there are many downsides.

pbugni: Okay, old habits...but a good suggestion.

PR #3107, opened 2019-04-03
Feature/vue components i18next fixes

achen2401:
https://jira.movember.com/browse/TN-1978
address issue where i18next strings were not being able to be extracted from Vue component by i18next scanner command tool.
Fix made:
moving data consumed by Vue components into separate JS data files, format recognized by i18next scanner.


portal/static/js/src/data/PsaTrackerData.js
```diff
@@ -0,0 +1,97 @@
+export default {
```
ivan-c: Would you be able to move or exclude this from the default string extraction file list since this is TNUSA-specific? No worries if there isn't an easy way- I can manually exclude them in Smartling but it can be error-prone.

achen2401: ah, thanks ivan, yeah, I think there is a way.. let me reorganize these a bit...  will report back when I am done.

achen2401: @ivan-c Ivan, not sure if this is your best way to do this - I broke down the data directory, i.e. common, gil and eproms so TNUSA-specific strings (like those from PSATrackerData.js) won't be extracted by default  (see my changes in i18next_gulpfile.js script. )  keen to hear your feedback on this.

ivan-c: Looks like a great temporary solution, but we'll need to come up with a better workflow for extracting strings in-place.

One possibility you might consider is: using a single exclude filter (see below) for the PSA file, if it makes the include filters simpler.

>! at the beginning of a pattern will negate the match

@achen2401 
I did some quick research and couldn't find anything we could plug in with little effort, could you create a story in Jira (to research/implement a way to extract strings from VueJS components) and I'll add some notes?
Thanks!!

portal/i18next_gulpfile.js
```diff
@@ -103,25 +108,25 @@ function writeJsonFileFromPoFile(locale, messageFilePath, outputFileName) {
  */
 gulp.task("i18next-extraction", ["clean-src"], function() {
   console.log("extracting text and generate json file ...");
-  return i18nextScanner(["static/**/*.{js,html}", "templates/*.html"], "./src/" + nameSpace + ".json");
+  return i18nextScanner([jsSrcPath+"*.{js,html}", jsSrcPath+"components/*.{js,html}", jsSrcPath+"mixins/*.{js,html}", jsSrcPath+"modules/*.{js,html}", jsSrcPath+"data/common/*.{js,html}", "templates/*.html"], "./src/" + nameSpace + ".json");
```
ivan-c: I might split this with one pattern per line, for readability

PR #3108, opened 2019-04-04
WIP TN-1953-IRONMAN-widen-QB-windows

mcjustin:
https://jira.movember.com/browse/TN-1979 impl
https://jira.movember.com/browse/TN-1953 spike

### DIFF
--- a/portal/config/eproms/QuestionnaireBank.json
+++ b/portal/config/eproms/QuestionnaireBank.json
@@ -38,7 +38,7 @@
     },
     {
       "classification": "baseline",
-      "expired": "{\"months\": 2, \"days\": 13}",
+      "expired": "{\"months\": 2, \"days\": 14}",
       "name": "IRONMAN_baseline",
       "overdue": "{\"days\": 1}",
       "questionnaires": [
@@ -88,7 +88,7 @@
     },
     {
       "classification": "baseline",
-      "expired": "{\"months\": 2, \"days\": 13}",
+      "expired": "{\"months\": 2, \"days\": 14}",
       "name": "IRONMAN_v3_baseline",
       "overdue": "{\"days\": 1}",
       "questionnaires": [
@@ -181,9 +181,9 @@
     {
       "classification": "recurring",
       "due": "{\"days\": 0}",
-      "expired": "{\"months\": 3, \"days\": -15}",
+      "expired": "{\"months\": 3, \"days\": -2}",
       "name": "IRONMAN_recurring_3mo_pattern",
-      "overdue": "{\"days\": 15}",
+      "overdue": "{\"days\": 16}",
       "questionnaires": [
         {
           "questionnaire": {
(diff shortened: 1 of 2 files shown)


mcjustin:
Scuttled https://github.com/uwcirg/truenth-portal/pull/3098 in favor of this (prefer to branch of master)

mcjustin:
Per earlier chatter at https://github.com/uwcirg/truenth-portal/pull/3098 
The docs for this are hidden in the [source code](https://github.com/uwcirg/truenth-portal/blob/develop/portal/models/questionnaire_bank.py#L41-L57), I see. Most everything is relative to start. In fact, I think you may want due to be days=0.

Thanks. Hmm, entry.start is currently set to 0 for all... Do you mean relative to entry.recurs.start ? I'll take another swing at this once the windows requirements have settled (hopefully tomorrow)

Correct. Here's the significant code, to determine the "start" of a "recurring QB" (self), where ic == iteration count, and trigger_date is the initial consent:
                start = (trigger_date + RelativeDelta(self.start) +
                         RelativeDelta(recur.start) +
                         (ic * RelativeDelta(recur.cycle_length)))

mcjustin:
@pbugni  re CommunicationRequest - I wasn't clear what date notify_post_qb_start is based on ... those likely need edits.

pbugni:
> @pbugni re CommunicationRequest - I wasn't clear what date notify_post_qb_start is based on ... those likely need edits.

`notifiy_post_qb_start` is in reference to the "calculated" start of any QB, effectively using the same algorithm posted above.  For example, if baseline has `start: days=0`, the start will be the consent date.  If 3 month has `start: days=0` and a recur `start: months=3`, the calculated start of the 3 month QB would be 3 months after the consent.  If you have a reminder on the 3 months with `notify_post_qb_start: days=7`, that would fire 7 days after the start of the QB, which translates to 3 months, 7 days after the consent.

pbugni:
@mcjustin Take a look at this when you get a chance.  Due to dates of different length, I think it may work better to generally use forward only math (months plus days, not minus).  We're really not set up to give them exactly what they want, as "overdue" should be relative to the trigger, not the QB start due to inconsistent months.  Once we publish, you can look at any patient's /timeline to see how this all lines up.

portal/config/eproms/QuestionnaireBank.json
```diff
@@ -180,8 +180,8 @@
     },
     {
       "classification": "recurring",
-      "due": "{\"days\": 14}",
-      "expired": "{\"days\": 29}",
+      "due": "{\"days\": 0}",
+      "expired": "{\"months\": 3, \"days\": -15}",
```
mcjustin: @pbugni  I thought you were preferring additive date math - why not make this consistent w/ line 91 ( "expired": "{\"months\": 2, \"days\": 13}",)

mcjustin: Otherwise, I think this looks good...

pbugni: I was playing with different values - additive and subtraction, trying to find the best combo.  The one checked in is pretty good, but it breaks the logic in `visit_name`, as we only look at the month.  (i.e. 3 month looks like 2 and 6 looks like 5 with this).

mcjustin:
@pbugni  didn't we plan to merge this into develop instead of master, now that we're aiming to include it in the RC this week?

PR #3111, opened 2019-04-05
TN-408 fix rendering of tree view for single parent organization

achen2401:
stumbled on this when testing this story:
https://jira.movember.com/browse/TN-408

Fix the rendering of single parent organizations from API, after adding legend for each:
/api/organization?tree_view=True


pbugni:
Thanks for noticing and fixing that!

PR #3117, opened 2019-04-08
TN-408 fix organization JSON for UCSF urologic clinic

achen2401:
per convo related to this story:
https://jira.movember.com/browse/TN-408
Fix organization JSON for UCSF Urologic Surgical Oncology Clinic so it is the child clinic of UCSF.


achen2401:
@mcjustin @ivan-c Fixing the order of JSON entries seemed to fix the failed tests, at least in my instance, thank you for the advices.

PR #3118, opened 2019-04-09
Tn 1924 eproms pw reset footer

mcjustin:
Known issue: needs translations (brought over from LR?)
Suspect: parent_org variable population?


portal/templates/flask_user/emails/eproms_base_message.html
```diff
@@ -6,3 +6,6 @@
 <p>Thank you,</p>
 <p>The TrueNTH Team</p>
 {% endtrans %}
+{% trans %}
```
ivan-c: I don't think `parent_org` is populated into *these* email templates, only a `User` object (`user`).

To get something similar, you might have to look at `user.organizations[0]` or [`user.first_top_organization()`](https://github.com/uwcirg/truenth-portal/blob/develop/portal/models/user.py#L785-L797) (haven't tested the below code)

...might also cause exceptions for test users (without orgs)
```suggestion
{% trans parent_org=user.first_top_organization() %}
```

mcjustin: Thanks @ivan-c  - I'm going to try the user.first_top_organization() approach.
Based on what I see at https://eproms-dev.cirg.washington.edu/api/user/1465/organization, only one org is at user.organization (indexed at 0), and it's the user's org, not the top parent of that org.

portal/templates/flask_user/emails/eproms_base_message.html
```diff
@@ -6,3 +6,6 @@
 <p>Thank you,</p>
 <p>The TrueNTH Team</p>
 {% endtrans %}
+{% trans parent_org=user.first_top_organization() %}
```
ivan-c: My mistake, this will return the top-level org (ie Ironman/CRV).

You'll just want to use the first org name in the way below:
```suggestion
{% trans parent_org=user.organizations[0].name %}
```
---
`flask shell` example:
```
>>> from portal.models.user import User
>>> User.query.get(1).organizations[0].name
u'Centro de Pesquisa em Oncologia'
```

mcjustin: @ivan-c  thanks for testing! I actually do want the top level org; I have just changed that to reference it's name, tho.

PR #3120, opened 2019-04-09
TN-1946 Dynamically lookup option text from Questionnaires

ivan-c:
* Add convenience property (`QuestionnaireResponse.document_answered`) for getting QNRs with dynamically populated answers
  * Lookup option text from Questionnaire resource referenced in QuestionnaireResponse, if available
* Override default behavior until migration removes superfluous `valueString` answers added by AE


ivan-c:
@pbugni Thanks for taking a look! I pinky-swear I'll add some tests when I rip out the extra `valueString`

PR #3138, opened 2019-04-16
Feature/manual entry consent completion dates

achen2401:
subtask for this story:
https://jira.movember.com/browse/TN-1965
For manual entry of papel questionnaire:
Dealing with the scenario where completion date is the same as the consent date.
Changes include:
- Use consent date/time if completion date is **the same** as consent date
- Use previous consent date/time for comparison if subject has withdrawn
- styling fixes of manual entry modal
- add audit entry for present assessment url (url redirect to assessment engine) so to see what parameters were actually passed

### DIFF
--- a/portal/views/assessment_engine.py
+++ b/portal/views/assessment_engine.py
@@ -1606,9 +1606,8 @@ def present_needed():
         return redirect('/')
 
     url = url_for('.present_assessment', **args)
-    auditable_event("present assessment url {}".format(url),
-                    user_id=current_user().id, subject_id=subject_id,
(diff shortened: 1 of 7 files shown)


portal/views/assessment_engine.py
```diff
@@ -1606,6 +1606,9 @@ def present_needed():
         return redirect('/')
 
     url = url_for('.present_assessment', **args)
+    auditable_event("present assessment url {}".format(url),
```
ivan-c: This seems like debugging info, is it really necessary to save to the audit log?

PR #3139, opened 2019-04-17
Flush qb_timeline for suspended workers, again.

pbugni:


ivan-c:
@pbugni Merge when ready

PR #3140, opened 2019-04-17
Set celery worker prod default

ivan-c:
* Respawn celery workers after 10 tasks to prevent potential memory leaks

### DIFF
--- a/docker/docker-compose.prod.yaml
+++ b/docker/docker-compose.prod.yaml
@@ -17,6 +17,10 @@ services:
 
   celeryworker:
     restart: unless-stopped
+    environment:
+      # Number of tasks a celery worker should run before being recreated
+      # https://docs.celeryproject.org/en/latest/userguide/configuration.html#std:setting-worker_max_tasks_per_child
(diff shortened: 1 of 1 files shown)


docker/docker-compose.prod.yaml
```diff
@@ -17,6 +17,10 @@ services:
 
   celeryworker:
     restart: unless-stopped
+    environment:
+      # Number of tasks a celery worker should run before being recreated
+      # https://docs.celeryproject.org/en/latest/userguide/configuration.html#std:setting-worker_max_tasks_per_child
+      CELERYD_MAX_TASKS_PER_CHILD: 10
```
pbugni: Love this idea, but 10 feels like an arbitrarily low number.  I suppose you're experimenting, but I would expect there's startup costs to each worker ...

ivan-c: We've have/had `10` running on *-test/dev for a few months without issue, but I haven't tried any other values.

I'm tempted to keep what we've been testing against (`10`), then update test/dev and compare memory usage (to demo).

@pbugni Thoughts?

PR #3148, opened 2019-04-18
TN-1426 remove users with admin role from staff list

achen2401:
address https://jira.movember.com/browse/TN-1426
remove admin users from staff list of staff/staff-admin users

changes include:
simply code to exclude admin users in query
remove previous code that did not exclude admin users as it should

### DIFF
--- a/portal/views/staff.py
+++ b/portal/views/staff.py
@@ -113,9 +113,7 @@ def staff_index():
     ot = OrgTree()
     staff_role_id = Role.query.filter(
         Role.name == ROLE.STAFF.value).with_entities(Role.id).first()
-    admin_role_id = Role.query.filter(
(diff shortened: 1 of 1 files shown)


portal/views/staff.py
```diff
@@ -131,21 +128,14 @@ def staff_index():
...
-             UserRoles.role_id.in_([admin_role_id]))
-    ).join(UserOrganization).filter(
-        and_(UserOrganization.user_id == User.id,
-             UserOrganization.organization_id.in_(user_orgs)))
-    admin_list = [u.id for u in admin_staff]
-
     # Gather up all staff belonging to any of the orgs (and their children)
     # NOTE, a change from before, staff admin users can now edit records of
     # other users that have staff OR staff admin role(s)
+    # excluding users with ADMIN role from the list
```
ivan-c: Don't think this comment is necessary

PR #3154, opened 2019-04-22
TN-2015 Translate Staff-initiated password reset

ivan-c:


portal/views/user.py
```diff
@@ -1906,16 +1907,21 @@ def trigger_password_reset_email(user_id):
...
         abort(400, "invalid email address")
-    auditable_event("password reset email triggered for user {}".format(
-        user_id), user_id=current_user().id, subject_id=user_id,
-        context='login')
+
+    try:
+        with force_locale(user.locale_code):
+            user_manager.send_reset_password_email(user.email)
+    except ValueError as e:
+        abort(400, str(e))
```
pbugni: Not that we were logging before, but given the recent "silent" error issues - should we at least log this exceptional event?  400s don't generate emails, as you know...

PR #3169, opened 2019-04-25
Cleanup/standardize docker scripts

ivan-c:
* Standardize common shell vars (script name, repo path etc)
* Use `docker inspect` to get `COMPOSE_PROJECT_NAME` for DB archive names
* `cd` to `docker/` directory instead of setting `COMPOSE_FILE` env var


bin/deploy-docker.sh
```diff
@@ -4,6 +4,9 @@
 # Environment variables passed to this script (eg IMAGE_TAG) will be available to the given docker-compose.yaml file
 
 cmdname="$(basename "$0")"
+bin_path="$(cd "$(dirname "$0")" && pwd)"
```
pbugni: At one point, you suggested killing the `docker` subdir and moving docker files to the root of the project.  Would this be a cleaner approach, now that we're effectively docker committed?

bin/deploy-docker.sh
```diff
@@ -35,28 +38,19 @@ done
...
-    web_image_hash="$(docker-compose images -q web | cut -c1-7)"
-    dump_filename="psql_dump-$(date --iso-8601=seconds)-${web_image_hash}-${COMPOSE_PROJECT_NAME}"
+if [ -n "$BACKUP" ] && [ -n "$(docker-compose ps --quiet db)" ]; then
+    # get COMPOSE_PROJECT_NAME (see .env)
+    compose_project_name="$(
+        docker inspect "$(docker-compose ps --quiet web)" \
+            --format '{{ index .Config.Labels "com.docker.compose.project"}}'
+    )"
+    web_image_hash="$(docker-compose images --quiet web | cut -c1-7)"
+    dump_filename="psql_dump-$(date --iso-8601=seconds)-${web_image_hash}-${compose_project_name}"
```
pbugni: not part of this PR, but why do we write the backup to /tmp/{filename} ?  wouldn't it be better to put in a longer living directory?  (yes, expanding scope due to volume needs, etc., but I'd think we'll want this before using on prod?)

bin/deploy-docker.sh
ivan-c: >At one point, you suggested killing the docker subdir and moving docker files to the root of the project

Some continuous integration services expect a Dockerfile at repo root (Heroku), but we aren't using any at the moment, and that would be the only reason.

>killing the `docker` subdir

I actually prefer a separate `docker/` directory for the docker-specific functionality :). The root of a repo can tend to get crowded with config/dotfiles

>now that we're effectively docker committed

I don't want to give it labels :wink: , but I've tried to make our tooling as free-standing as possible (ie useable independently outside our docker deployment method).

bin/deploy-docker.sh
ivan-c: `/tmp` is one of the few places I should always be able to write to, otherwise this script would require `sudo`.

Maybe I can accept an optional path to save backups to, defaulting to `/tmp` if nothing is passed with `-b`?

pbugni: The rational for using /tmp makes perfect sense.  I'll ask out of band how we're retaining those, no need to work in yet another option parameter.  Fine for this PR.

PR #3171, opened 2019-04-25
TN-1983 Add entry method to adherence report

pbugni:


portal/views/reporting.py
```diff
@@ -284,25 +290,38 @@ def questionnaire_status():
...
-                'user_id', 'study_id', 'status', 'visit', 'site', 'consent']
-            yield ','.join(desired_order) + '\n'  # header row
+            desired_order = (
+                'user_id', 'study_id', 'status', 'visit',
+                'entry_method', 'site', 'consent')
+            yield desired_order  # header row
             for i in items:
-                yield ','.join(
-                    [str(i.get(k, "")) for k in desired_order]) + '\n'
+                if int(i['user_id']) not in (1707, 1726):
```
ivan-c: Leftover debugging code?

portal/views/reporting.py
```diff
@@ -284,25 +290,38 @@ def questionnaire_status():
...
                 Organization.query.get(org_id).name.replace(' ', '-'))
         filename = '{}-{}.csv'.format(base_name, strftime('%Y_%m_%d-%H_%M'))
+
+        si = StringIO()
+        cw = csv.writer(si)
+        cw.writerows(gen(results))
+
         return Response(
-            gen(results),
+            si.getvalue(),
```
ivan-c: I think this no [longer streams the CSV response](http://flask.pocoo.org/docs/1.0/patterns/streaming/), which may be desirable for enormous datasets.

Using the `csv` module would be a great improvement for standardization if we could get it to work in the same way...

portal/views/reporting.py
pbugni: *blush*! thanks for catching that

portal/views/reporting.py
pbugni: Found [a solution|https://stackoverflow.com/a/32609533] to bring back streaming using the csv module.

PR #3172, opened 2019-04-25
TN-2007 remove microseconds from user_consent.acceptance_date fields.

pbugni:


portal/models/user_consent.py
```diff
@@ -38,7 +38,7 @@ class UserConsent(db.Model):
     organization_id = db.Column(
         db.ForeignKey('organizations.id'), nullable=False)
     acceptance_date = db.Column(
-        db.DateTime, default=datetime.utcnow, nullable=False)
+        db.DateTime, default=utcnow_sans_micro, nullable=False)
```
ivan-c: Can you add a comment why this is the default or why microseconds are problematic?

PR #3173, opened 2019-04-26
TN-2030 use less to allow org tree css to be served correctly

achen2401:
https://jira.movember.com/browse/TN-2030

seeing this error in console:

```
Refused to apply style from 'https://eproms-test.cirg.washington.edu/static/css/orgTreeView.css' because its MIME type ('text/html') is not a supported stylesheet MIME type, and strict MIME checking is enabled.
```

giving this fix a try, not sure if it will work.


portal/templates/org_tree.html
```diff
@@ -1,6 +1,6 @@
 {% extends "layout.html" %}
 {% block main %}
-<link rel="stylesheet" href="{{ url_for('static', filename='css/orgTreeView.css') }}">
+<link rel="stylesheet" href="{{ url_for('static', filename='css/orgTreeView.css') }}" type="text/css">
```
ivan-c:
```
type="text/css"
```
is this still needed? [sounds like](https://stackoverflow.com/questions/5409114/is-type-text-css-necessary-in-a-link-tag) it might not be necessary...

PR #3186, opened 2019-05-01
i18next script bug fix

achen2401:
address issue where frontend tranlation json files were not being converted to pot files

- fix is to wait for i18next scanner to finish extracting json text before conversion.  

@ivan-c ivan, let me know if you see issue, it worked in my instance after I made the fixes.


achen2401:
@ivan-c ivan, I **think**  this was an issue but we just didn't catch it. I think the context separator, one of the available options for the i18next converter is the culprit.  Let me know if you see anymore issue.

PR #3190, opened 2019-05-03
Feature/dockerize site persistence

pbugni:
site persistence can't write to read only filesystem (such as when running in docker container).  simply log warning if `site.cfg` is not writable.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,7 +19,7 @@ pip-selfcheck.json
 log
 portal/test_uploads
 docker/portal.env
-docker/docker-compose.pb.yaml
+docker/docker-compose.dev.*.yaml
 
 geckodriver.log
 coverage.xml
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -19,6 +19,7 @@ pip-selfcheck.json
 log
 portal/test_uploads
 docker/portal.env
+docker/docker-compose.pb.yaml
```
ivan-c:
```suggestion
docker/docker-compose.*.yaml
```

pbugni: Hmm, but that would also hide `docker-compose.{dev,flower,prod}.yaml` - should we try `docker-compose.personal.yaml` or a different pattern, such as `docker-compose.dev.*.yaml`?

ivan-c: >Hmm, but that would also hide docker-compose.{dev,flower,prod}.yaml

You would have to use `--force` with `git add` on those files, but I don't think any other change

>should we try docker-compose.personal.yaml or a different pattern, such as docker-compose.dev.*.yaml

Would prefer a wildcard to allow for some flexibility, the second pattern (`docker-compose.dev.*.yaml`) works for me though

PR #3192, opened 2019-05-04
Hotfix original CSV generator function

ivan-c:
Temp fix for below exception:
```
   File "/srv/www/eproms.cirg.washington.edu/portal/env/local/lib/python2.7/site-packages/werkzeug/wsgi.py", line 870, in __next__
     return self._next()
   File "/srv/www/eproms.cirg.washington.edu/portal/env/local/lib/python2.7/site-packages/werkzeug/wrappers.py", line 82, in _iter_encoded
     for item in iterable:
   File "/srv/www/eproms.cirg.washington.edu/portal/portal/views/reporting.py", line 308, in gen
     writer.writerow(desired_order)
 TypeError: unicode argument expected, got 'str'
```

### DIFF
--- a/portal/views/reporting.py
+++ b/portal/views/reporting.py
@@ -305,7 +305,8 @@ def questionnaire_status():
             yield ','.join(desired_order) + '\n'  # header row
             for i in items:
                 yield ','.join(
-                    [str(i.get(k, "")) for k in desired_order]) + '\n'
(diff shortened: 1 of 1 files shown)


pbugni:
@ivan-c & @mcjustin this works in python2, to create tables that align well with the embedded commas.
the csv library is a better solution, but the python2 version of that library can't handle unicode, and the workaround code is ugly - we'll wait till we're on py3 to port over.

PR #3201, opened 2019-05-07
Debian virtualenv workaround

ivan-c:
* Invoke python commands in interpreter-agnostic way (`python -m foo bar` vs `foo bar`) 
* Work around [bug](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=880472) in [debian python wheels ](https://packages.debian.org/stretch/python-pip-whl) with virtualenv
* Install latest version of `virtualenv` locally, ignoring [`virtualenv` (py3k) provided by debian package](https://packages.debian.org/stretch/virtualenv)

### DIFF
--- a/bin/nodejs-wrapper.sh
+++ b/bin/nodejs-wrapper.sh
@@ -31,24 +31,6 @@ setup_python_venv() {
     fi
 
 
-    echo virtualenv debug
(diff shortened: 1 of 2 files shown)


ivan-c:
Closing this until we need it

PR #3208, opened 2019-05-08
rename medidataRave config variables && remove medidataRave config variables to allow editing by default

achen2401:
Remove MedidataRave configuration variable to allow editing by default.
NOTE: **Override** can be applied by instance, as we currently do in Production to prevent editing  of MedidataRave fields there.


portal/config/eproms/site_persistence_file.json
```diff
@@ -21,7 +21,6 @@
         "CUSTOM_PATIENT_DETAIL = True\n",
         "LOCALIZED_AFFILIATE_ORG = 'TrueNTH Global Registry'\n",
         "ORGS_W_CUSTOM_INVITES = ['TrueNTH Global Registry', 'IRONMAN']\n",
-        "MEDIDATA_RAVE_FIELDS = ['deceased', 'study_id', 'consent_status', 'dob', 'org']\n",
```
pbugni: Well, you're not going to like my $.02, but to me, this variable name is undesirable.
1. it's defining which fields to hide/protect
2. it's vendor specific
If it defines a list of buttons/controls/fields to disable or hide - lets call it something like "PROTECTED_FIELDS" and the matching "PROTECTED_ORG"

But not critical, so I'm just submitting as a comment.

ivan-c: Can you add `MEDIDATA_RAVE_ORG` and `MEDIDATA_RAVE_FIELDS` to  `config.py` so that they can be configured via environment variable ([eg `PIWIK_DOMAINS`](https://github.com/uwcirg/truenth-portal/blob/develop/portal/config/config.py#L117-L118))?

ivan-c: >lets call it something like "PROTECTED_FIELDS" and the matching "PROTECTED_ORG"

I like @pbugni s suggestion, and since we have to formalize these configs (add to `config.py`) anyway, can you rename them to something more generic?

portal/config/config.py
```diff
@@ -111,7 +111,8 @@ class BaseConfig(object):
 
     # Medidata integration configuration
     # disable creation and editing of patients when active
-    MEDIDATA_RAVE_ORG = os.environ.get('MEDIDATA_RAVE_ORG')
+    PROTECTED_ORG = os.environ.get('PROTECTED_ORG')
```
ivan-c: Could you add a comment about whether this is a org name/id etc?

portal/config/config.py
```diff
@@ -111,7 +111,8 @@ class BaseConfig(object):
...
+    PROTECTED_FIELDS = os.environ.get('PROTECTED_FIELDS')
```
ivan-c: the values of environment variables are strings and need to be converted to the correct type (if we need something besides a string), vs `.cfg` files which actually are treated the same as `.py` files

```suggestion
    PROTECTED_FIELDS = os.environ['PROTECTED_FIELDS'].split(',') \
        if os.environ.get('PROTECTED_FIELDS') else None
```

PR #3209, opened 2019-05-08
TN-1807 Run Adherence Report as background task

pbugni:
The adherence report (triggered from the /patients view in eproms) is now generated as a background [read: celery] task.
The immediate response is a 202 including a `Location` header, defining where to obtain job status.  You'll note, once it starts, the `state` moves to PROGRESS and includes integer values for `total` and `current` (only updating every 25 atm).  Example job status output:
```
{"current":275,"state":"PROGRESS","total":900}
```
Once the job has completed, the status will show `state`:`SUCCESS`, and then the browser should be redirected to the celery-result URL (just chop the suffix `/status` from the status url).


achen2401:
@pbugni paul, added frontend changes.  let me know if they work for you.  I did test a bit before committing the changes.

portal/views/reporting.py
```diff
@@ -223,105 +218,20 @@ def questionnaire_status():
...
+        'acting_user_id': current_user().id,
+        'include_test_role': request.args.get('include_test_role', False),
+        'org_id': request.args.get('org_id'),
+        'response_format': request.args.get('format', 'json').lower()
+    }
+
+    # Hand the task off to the job queue, and return 202 with URL for
+    # checking the status of the task
+    task = adherence_report_task.apply_async(kwargs=kwargs)
+    return jsonify({}), 202, {'Location': url_for(
```
ivan-c: @achen2401 @pbugni 
I just wanted to add that I confirmed `url_for()` will return an `https` prefixed URL in prod deployments, without `scheme` needing to be explicitly passed.

just FYI I used this logging line to test `request.scheme` during a request
`current_app.logger.debug('scheme: %s', request.scheme)`

portal/celery_worker.py
```diff
@@ -26,26 +26,31 @@ def setup_periodic_tasks(sender, **kwargs):
...
+                                    schedule="0 * * * *", active=True)
+            session.add(test_job)
+            session.commit()
+
+        # add all tasks to Celery
+        logger.info("ScheduledJobs found: {}".format(
+            ScheduledJob.query.count()))
+        for job in session.query(ScheduledJob).all():
+            task = getattr(tasks, job.task, None)
+            if task:
```
ivan-c: Could be inverted (`if not task; continue`) into a guard clause for less nesting

portal/config/config.py
```diff
@@ -53,10 +53,16 @@ class BaseConfig(object):
...
     # Allow Heroku env vars to override most defaults
     # NB: The value of REDIS_URL may change at any point
-    REDIS_URL = os.environ.get(
-        'REDIS_URL',
-        'redis://localhost:6379/0'
-    )
+
+    # We override REDIS_URL when testing now to avoid needing to
+    # also reset the other variables using it as a default below
+    if os.environ.get('TESTING', 'false').lower() == 'true':
```
ivan-c: Would you be able to move this into the [app factory](https://github.com/uwcirg/truenth-portal/blob/develop/portal/factories/app.py#L131), so it can be conditional on `app.testing`, and `os.environ['TESTING']` will only be read once (and set `app.testing`)?

portal/models/qb_timeline.py
```diff
@@ -412,7 +413,7 @@ def attempt_update(user_id, invalidate_existing):
 
         # acquire a multiprocessing lock to prevent multiple requests
         # from duplicating rows during this slow process
-        timeout = current_app.config.get("MULTIPROCESS_LOCK_TIMEOUT")
+        timeout = int(current_app.config.get("MULTIPROCESS_LOCK_TIMEOUT"))
```
ivan-c: Is there a better place to cast this, so it will always be an `int` in `app.config.["MULTIPROCESS_LOCK_TIMEOUT"]`?

portal/models/reporting.py
```diff
@@ -102,6 +106,120 @@ def get_reporting_stats():
...
+            continue
+
+        try:
+            acting_user.check_role('view', other_id=patient.id)
+        except Unauthorized:
+            # simply exclude any patients the user can't view
+            continue
+
+        qb_stats = QB_Status(user=patient, as_of_date=as_of_date)
+        row = {
```
ivan-c: Would it make sense to start to formalize some of the CSVs we're writing as classes?

portal/tasks.py
```diff
@@ -96,6 +95,13 @@ def info():
         current_app.config.get('SERVER_NAME'))
 
 
+@celery.task(bind=True, ignore_result=False, max_retries=3, track_started=True)
```
ivan-c: Isn't `ignore_result=False` the default?
```suggestion
@celery.task(bind=True, max_retries=3, track_started=True)
```

ivan-c: Did you find the `max_retries=3` necessary?

pbugni: will remove - both came along during hair pulling debugging

portal/models/reporting.py
pbugni: err - perhaps.  i think i'll punt on that level of change at this time, especially as this needs serialization support during its redis life.

portal/models/qb_timeline.py
pbugni: I don't know of one.  The default in config.py IS an int - we ran into this when customized in `application.cfg`.  This is the only use of said configuration.  My attempt to recreate the problem fails.

portal/config/config.py
pbugni: I don't believe so.  Can set `app.testing` as you suggest, but we don't yet have an application context when reading the config, so I can't lookup the value of app.testing at this LOC.

ivan-c: >Can set app.testing as you suggest
Sorry, should've been more clear. Flask should set `app.testing` if `app.config['TESTING'] == True`. I'm not sure where/when flask does, but we should probably not modify that behavior.

I was hoping we could key all our configuration directly off of `app.testing` instead of some on `app.config['TESTING']`

PR #3239, opened 2019-05-21
Cleanup string extraction tooling

ivan-c:
* Rewrite POT file generation to idiomatic code, with `polib`
* Fix POT file reference format for static & DB strings
  * Use available object metadata as gettext references (translation context)

### DIFF
--- a/portal/models/i18n.py
+++ b/portal/models/i18n.py
@@ -39,7 +39,6 @@ def get_db_strings():
                 msgid = getattr(entry, field_name)
                 if not msgid:
                     continue
-                msgid = '"{}"'.format(re.sub('"', r'\\"', msgid))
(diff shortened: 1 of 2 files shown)


portal/models/i18n.py
```diff
@@ -16,13 +16,19 @@
...
 from .organization import Organization
+from .overall_status import OverallStatus
 from .questionnaire_bank import QuestionnaireBank, classification_types_enum
 from .research_protocol import ResearchProtocol
 from .role import Role
 from .user import current_user
 
 
 def get_db_strings():
+    """
```
pbugni: As suggested by pep8, we tend to use the [Multi-Line Docstring](https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings) format - so that first summary line is just right of the opening triple quotes.

PR #3244, opened 2019-05-22
TN-2071-New-IRONMAN-UK-org

mcjustin:
https://jira.movember.com/browse/TN-2071

### DIFF
--- a/portal/config/eproms/Organization.json
+++ b/portal/config/eproms/Organization.json
@@ -425,34 +425,6 @@
     {
       "extension": [
         {
-          "timezone": "Europe/London",
-          "url": "http://hl7.org/fhir/StructureDefinition/user-timezone"
-        },
-        {
-          "research_protocols": [
-			{"name": "IRONMAN v3"}
-          ],
-          "url": "http://us.truenth.org/identity-codes/research-protocol"
-        }
-      ],
-      "id": 14644,
-      "identifier": [
-        {
-          "system": "http://pcctc.org/",
-          "use": "secondary",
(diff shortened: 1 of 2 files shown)


mcjustin:
@ivan-c  sorry don't merge, they have a conflicting org ID here

portal/config/eproms/Organization.json
```diff
@@ -1981,6 +1981,34 @@
...
+          "url": "http://hl7.org/fhir/StructureDefinition/user-timezone"
+        },
+        {
+          "research_protocols": [
+			{"name": "IRONMAN v3"}
+          ],
+          "url": "http://us.truenth.org/identity-codes/research-protocol"
+        }
+      ],
+      "id": 1464499,
```
pbugni: Hmm, I get your numbering (to make unique), but this has me wondering if we shouldn't simply delete the old organization in place of the new.  Is there any value in keeping it around?  I get your concerns with test patients, but this should be inconsequential as a new org is taking its place.

In other words, how about we simply update the old organization's name, partOf and other altered bits?

mcjustin: @pbugni  sounds good, will make that change

tests/test_assessment_status.py
```diff
@@ -745,8 +745,9 @@ def test_boundary_in_progress(self):
 
     def test_boundary_recurring_in_progress(self):
         self.login()
-        backdate, nowish = associative_backdate(
```
ivan-c: :+1:

PR #3256, opened 2019-05-28
GIL - portal wrapper loading error message fix

achen2401:
address error message reported when portal wrapper failed to load in GIL: 

`Received error {u'actor': u'user 12392',
 u'message': u'In GIL, error loading nav elements from portal - undefined' ...`

Fix: Provide error status code if response text is not available

### DIFF
--- a/portal/static/js/src/gil.js
+++ b/portal/static/js/src/gil.js
@@ -1339,7 +1339,7 @@ module.exports = utilObj = (function() {
           params.subject_id = $("#currentUserId").val() ? $("#currentUserId").val() : 0;
           params.page_url = window.location.href;
           //don't think we want to translate message sent back to the server here
-          params.message = "In GIL, error loading nav elements from portal - " + xhr.responseText;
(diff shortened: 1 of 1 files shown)


portal/static/js/src/gil.js
```diff
@@ -1339,7 +1339,7 @@ module.exports = utilObj = (function() {
           params.subject_id = $("#currentUserId").val() ? $("#currentUserId").val() : 0;
           params.page_url = window.location.href;
           //don't think we want to translate message sent back to the server here
-          params.message = "In GIL, error loading nav elements from portal - " + xhr.responseText;
+          params.message = "In GIL, error loading nav elements from portal - " + ("status code: "+ xhr.status) + "\nresponse: " + (xhr.responseText?xhr.responseText:"empty");
```
ivan-c: We don't need translations here, but I'd like to work towards a general set of error messages we can reuse for most situations. Ideally we wouldn't use too many placeholders (one-word strings would be difficult/impossible to translate)


eg:

```
"{application_state}: error saving data"
"{application_state}: error loading data"
```

PR #3270, opened 2019-05-31
Hotfix scheduled job load

ivan-c:
* Restore original behavior before #3209


ivan-c:
Tested on my local docker instance, confirmed working

PR #3283, opened 2019-06-05
Improve docs & deploy testing

ivan-c:
* Move inline shell script to standalone script file
  * Add comments, logging
* Add notes on developing with docker


docker/docker-compose.yaml
```diff
@@ -49,7 +49,7 @@ services:
     <<: *service_base
     command: bash -c '
       (wait-for-it --host=redis --port=6379 --strict -- flask set-celery-beat-healthy) &
-      wait-for-it --timeout=120 --host=web --port="$$PORT" --strict --
+      wait-for-it --timeout=120 --host=web --port=$$PORT --strict --
```
pbugni: why do you need a double dollar sign here?

ivan-c: That's the special way to escape `$` in [`docker-compose.yml`
](https://docs.docker.com/compose/compose-file/#variable-substitution)

>You can use a $$ (double-dollar sign) when your configuration needs a literal dollar sign. This also prevents Compose from interpolating a value, so a $$ allows you to refer to environment variables that you don’t want processed by Compose.

PR #3288, opened 2019-06-07
Build front-end code in isolated stage

ivan-c:
* Move front-end build process to separate docker stage
* Change `nodejs-wrapper.sh` to pass through commands if `npm` installed already

Todo:

- [ ] Use Makefile functionality to conditionally build front-end files if (bundles) absent

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,3 +1,29 @@
+FROM debian:buster as frontend
+
+
+ENV \
+    DEBIAN_FRONTEND=noninteractive
+
+WORKDIR "$HOME/portal"
+
+# Copy repo into image
+COPY . .
+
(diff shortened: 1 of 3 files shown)


docker/Dockerfile
```diff
@@ -1,3 +1,17 @@
+FROM node:10 as frontend
+
+WORKDIR "/portal"
+
+# Copy required front-end files only
+# preserve partial checkout structure
+COPY bin/ bin/
```
pbugni: err, this has me scratching my head.  copy `bin/` to `bin/`?  isn't that and the following line a no-op?

ivan-c:
Needs prerequisite work to cleanup build process; closing for now

PR #3291, opened 2019-06-10
TN-2094 fix JS runtime error when sorting null consentdate on patient

achen2401:
address https://jira.movember.com/browse/TN-2094
fix JS runtime error raised when sorting was attempted on consent date field of the patient list - specifically on rows containing empty consent date
specific error thrown `admin.js:78 Error thrown:  Error: TypeError: a.replace is not a function`
The runtime error halted subsequent JS executions.
Fix made: added falsy checks on date variables being compared  before performing string-related operation on each

### DIFF
--- a/portal/static/js/vendor/bootstrap_datatables_extension.js
+++ b/portal/static/js/vendor/bootstrap_datatables_extension.js
@@ -38,6 +38,7 @@ var tnthTables = {
     "dateSorter": function(a, b) {
         a = a || 0;
         b = b || 0;
+        var a_d = 0, b_d = 0;
         /*
          * make sure the string passed in does not have line break element if so it is a possible mult-line text, split it up and use the first item in the resulting array
          */
@@ -49,6 +50,10 @@ var tnthTables = {
             if (ar.length > 0) {
                 a = ar[0];
             }
+            /* note getTime returns the numeric value corresponding to the time for the specified date according to universal time
+            * therefore, can be used for sorting
+            */
+            a_d = (new Date(a)).getTime();
         }
         //NOTE: adding if clause here because b is set to 0 if null/undefined, attempting to perform string specific operations,e.g. regular expression matching as follows, on 0, a numeric variable will result in runtime error
         if (b) {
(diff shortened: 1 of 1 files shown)


portal/static/js/vendor/bootstrap_datatables_extension.js
```diff
@@ -42,15 +42,21 @@ var tnthTables = {
...
-        var br = b.split("\n");
-        if (br.length > 0) {
-            b = br[0];
+        //NOTE: adding if clause here because b is set to 0 if null/undefined, attempting to perform string specific operations,e.g. regular expression matching as follows, on 0, a numeric variable will result in runtime error
+        if (b) {
+            b = b.replace(regex, "\n");
+            var br = b.split("\n");
+            if (br.length > 0) {
+                b = br[0];
+            }
```
pbugni: I see use of both `a` and `b` below w/o similar null checks?

achen2401: thanks, paul, null checks added.

portal/static/js/vendor/bootstrap_datatables_extension.js
```diff
@@ -38,33 +38,39 @@ var tnthTables = {
...
          */
         var regex = /<br\s*[\/]?>/gi;
-        a = a.replace(regex, "\n");
-        b = b.replace(regex, "\n");
-        var ar = a.split("\n");
-        if (ar.length > 0) {
-            a = ar[0];
+        //NOTE: adding if clause here because a is set to 0 if null/undefined, attempting to perform string specific operations,e.g. regular expression matching as follows, on 0, a numeric variable will result in runtime error
+        if (a) {
+            a = a.replace(regex, "\n");
```
ivan-c: Should this logic be run when populating the table data, instead of sorting it? Wouldn't this change only affect the sort order if there's a leading `\n` in a cell?

achen2401: hmm, I believe, at least my understanding is that this part of code is called when the column is actively sorted via either user interaction on the UI or manual triggering , so this in fact _is_ dealing specifically with sorting  - so yes, it only affects sort order and would have been invoked **after** the table data has already been populated.

PR #3312, opened 2019-06-18
Audit Log runtime JS error fix

achen2401:
stumbled on this error while working in my instance:
Runtime JS error, i.e. URI error, malformed URI, generated in audit log display of some user accounts in profile:
example audit entry that throws error: 
`... by SELECT performers.id AS performers_id, performers.reference_txt AS performers_reference_txt, performers.codeable_concept_id AS performers_codeable_concept_id 
FROM performers, observation_performers 
WHERE **%(param_1)s** = observation_performers.observation_id AND performers.id = observation_performers.performer_id with status`  
**%(** when decodeURIComponent function is applied (to decode escaped characters) to it generates the error

try and catch statement is added to catch such error

### DIFF
--- a/portal/static/js/src/profile.js
+++ b/portal/static/js/src/profile.js
@@ -1894,7 +1894,7 @@ export default (function() {
                         $("#profileAuditLogErrorMessage").text(i18next.t("No audit log item found."));
                         return false;
                     }
-                    var ww = $(window).width(), fData = [], len = ((ww < 650) ? 20 : (ww < 800 ? 40 : 80));
+                    var ww = $(window).width(), fData = [], len = ((ww < 650) ? 20 : (ww < 800 ? 40 : 80)), errorMessage="";
                     (data.audits).forEach(function(item) {
                         item.by = item.by.reference || "-";
                         var r = /\d+/g;
@@ -1910,6 +1910,7 @@ export default (function() {
                         } catch(e) { //catch error, output to console e.g. URI malformed error. output error to console for debugging purpose
                             console.log("Error decoding auditing comment " + c);
                             console.log(e);
+                            errorMessage = e.message + "" + e.stack;
                         }
                         item.comment = c.length > len ? (c.substring(0, len + 1) + "<span class='second hide'>" + (c.substr(len + 1)) + "</span><br/><sub onclick='{showText}' class='pointer text-muted'>" + i18next.t("More...") + "</sub>") : item.comment;
                         item.comment = (item.comment).replace("{showText}", "(function (obj) {" +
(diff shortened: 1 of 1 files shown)


ivan-c:
Is this an issue in a jinja template, or AJAX call? What's the full stack trace for this exception?

I'm not 100% sure what's going on, but we may need to call equivalent code (URI decoding) in flask. As far as I know, SQLAlchemy should correctly escape values to prevent SQL syntax errors.
```
... by SELECT performers.id AS performers_id, performers.reference_txt AS performers_reference_txt, performers.codeable_concept_id AS performers_codeable_concept_id 
FROM performers, observation_performers 
WHERE **%(param_1)s** = observation_performers.observation_id AND performers.id = observation_performers.performer_id with status
**%(** 
```

achen2401:
> Is this an issue in a jinja template, or AJAX call? What's the full stack trace for this exception?
> 
> I'm not 100% sure what's going on, but we may need to call equivalent code (URI decoding) in flask. As far as I know, SQLAlchemy should correctly escape values to prevent SQL syntax errors.
> 
> ```
> ... by SELECT performers.id AS performers_id, performers.reference_txt AS performers_reference_txt, performers.codeable_concept_id AS performers_codeable_concept_id 
> FROM performers, observation_performers 
> WHERE **%(param_1)s** = observation_performers.observation_id AND performers.id = observation_performers.performer_id with status
> **%(** 
> ```

the error occurred in Ajax call.  full error stack:
```Uncaught URIError: URI malformed
    at decodeURIComponent (<anonymous>)
    at <anonymous>:1:1```

think the error occurred trying to decode **%(** in the sql statement there. try this in console and will generate the error stack `decodeURIComponent("%(")`

File: portal/static/js/src/profile.js [hunk 1]
```diff
@@ -1904,7 +1904,13 @@ export default (function() {
                         }
                         item.lastUpdated = self.modules.tnthDates.formatDateString(item.lastUpdated, "iso");
                         item.comment = item.comment ? self.escapeHtml(item.comment) : "";
-                        var c = decodeURIComponent(String(item.comment));
+                        var c = String(item.comment);
+                        try {
+                            c = decodeURIComponent(c);
+                        } catch(e) { //catch error, output to console e.g. URI malformed error. output error to console for debugging purpose
+                            console.log("Error decoding auditing comment " + c);
```
pbugni: I like the more robust decode attempt, but worry about relying on the console.  Should the failure case be transmitted back to the portal's error page?

achen2401: thanks, paul, added fix to report error back to server, so at least we are aware of the error.

pbugni:
It would be nice to see where the source of the lack of encoding stems from.  I'd like to keep this change, but perhaps we need to fix the other side as well?

achen2401:
@ivan-c and @pbugni 
Here is the patient account that has such error (you will see the JS error in the console) in Truenth-Test:
`https://truenth-test.cirg.washington.edu/patients/patient_profile/11672`
I believe it is failing on an audit entry that has sql statement as follows in it: `...WHERE %(param_1)s = observation_performers...` 
Note the audit log **is not** rendered currently due to the JS runtime error for that particular patient account

ivan-c:
Sorry for not merging earlier; we can fix the back-end in a follow-up todo

PR #3332, opened 2019-07-03
TN-2130 fix google auth

pbugni:
Attempt minor code change for google auth scopes as per warning message.


pbugni:
After deploying this patch to https://truenth-dev.cirg.washington.edu/ - I and Warren Cheong found it to solve the problem.

Merging w/o further ado.

PR #3346, opened 2019-07-10
TN-2110 migration to mask email of all SR partners and patients.

pbugni:


portal/migrations/versions/c242e22f5a47_.py
```diff
@@ -0,0 +1,81 @@
...
+
+from portal.models.audit import lookup_version
+from portal.models.user import NO_EMAIL_PREFIX
+
+# revision identifiers, used by Alembic.
+revision = 'c242e22f5a47'
+down_revision = '29ddc9e0f61d'
+
+userlist = (
+    "select user_id, email "
```
ivan-c: After the MR 2-3 digit fiasco I've been leaning towards making migrations less disposable/more reusable.

This looks like it might be helpful as a model method, but probably not worth the time

PR #3352, opened 2019-07-10
TN-1773 Switch to python3 for packaged virtual environment

ivan-c:
* Switch to python3-based virtual environment (built by [dh-virtualenv](https://dh-virtualenv.readthedocs.io/en/1.1/howtos.html#building-packages-for-python3))
* Set locale as UTF8 when calling `flask` CLI to compile translations
* Add [todo](https://github.com/uwcirg/truenth-portal/pull/3352/files#diff-d837a1c17de0268bcea321239ddbed47R8) for when [dh-virtualenv becomes python3-default](https://github.com/spotify/dh-virtualenv/issues/274)
  * python2 will remain required for build process only (ie in [`builder` stage](https://github.com/uwcirg/truenth-portal/blob/develop/docker/Dockerfile#L1))

NB: `python` still maps to `python2` [as per debian policy](https://www.debian.org/doc/packaging-manuals/python-policy/ch-python.html#s-interpreter_name)

#### Todo:
[Use new builtin `venv` module for creating debian-packaged virtual environment](https://github.com/uwcirg/truenth-portal/pull/3360)


bin/nodejs-wrapper.sh
```diff
@@ -51,7 +52,7 @@ setup_node_venv() {
     . "${python_venv_path}/bin/activate"
 
     echo "Installing node_env"
-    pip install nodeenv
+    python -m pip install nodeenv
```
pbugni: line 35 of same script uses `python3` vs naked `python` here.  Intentional?  (need comment as to why?)

debian/portal.triggers
```diff
@@ -1,7 +1,7 @@
 # Register interest in Python interpreter changes (Python 2 for now); and
 # don't make the Python package dependent on the virtualenv package
 # processing (noawait)
-interest-noawait /usr/bin/python2.7
+interest-noawait /usr/bin/python3.5
```
pbugni: isn't `/usr/bin/python3` a symbolic link to the desired version 3 installed?

bin/nodejs-wrapper.sh
ivan-c: It's debian policy to keep `python` as `python2` ([and any attempt to change it will break the system](https://stackoverflow.com/questions/43062608/how-to-update-alternatives-to-python-3-without-breaking-apt/46095470#46095470)); this invocation (`python`) is from inside an activated virtual environment, which sets up `python` to point to `python3` as we would hope.

I'll change this line to `python3` to make it explicit. Hopefully debian will eventually switch :)

debian/portal.triggers
ivan-c: I'm not 100% sure why this references a specific version (I copy/pasted this change from [their python3 docs](https://dh-virtualenv.readthedocs.io/en/1.1/howtos.html#building-packages-for-python3)) but it seems like [`deb-triggers`](http://man7.org/linux/man-pages/man5/deb-triggers.5.html) takes `/usr/bin/python3.5` as the `trigger-name`. 

Perhaps there's a pre-defined list `trigger-name` is from, and `trigger-name` is not a binary/symlink?

PR #3353, opened 2019-07-10
Backup/Migration Restore Scripts 

ivan-c:
* Add scripts to restore a containerized portal from backup (user uploads, SQL dump)


jhowe-uw:
Probably add a simple check when dropping the DB that a dump file exists for restoration, or at least an "Are you sure?" before nuking the DB.

ivan-c:
> Probably add a simple check when dropping the DB that a dump file exists for restoration, or at least an "Are you sure?" before nuking the DB.

Thanks for the suggestions!

Script now checks that the given dump is a real non-zero file (`test -s`). Script also prompts unless `-f` is passed

ivan-c:
@pbugni 
> I'd be tempted to expect the sql dump from stdin so we could chain the commands, but that would make it required and handling the uploads separately ill defined.

I did a little digging and it sounds like you can [replicate this behavior in the shell with `/dev/stdin`](https://superuser.com/questions/747884/how-to-write-a-script-that-accepts-input-from-a-file-or-from-stdin/807874#807874). eg:
```
# cat would be replaced with ssh pipe or similar
cat dump.sql | restore-truenth.sh -f -s /dev/stdin
```
I may make `/dev/stdin` the default in some cases, but will have to read more about it

---
> Would it make any more sense to break into two scripts that take stdin if not given an input file parameter? (This shouldn't hold up rolling this out, perhaps a future optimization?)

I may do this before merging; I like keeping the scripts simple and I think it will be worth the extra boilerplate

PR #3357, opened 2019-07-16
Use python3 venv module to create packaged virtual environment

ivan-c:
* Use `venv` module with dh-virtualenv to create debian-packaged virtual environment
  * currently uses `virtualenv` binary
* Switch to new debian (buster) images
  * work around pip bootstrapping issues (see #3201)


ivan-c:
Looks like GH freaked out when I force-pushed new PR here: #3360

PR #3371, opened 2019-07-22
TN-1925 now defaults to en_US, orig request errant. + 3 more stories

mcjustin:


mcjustin:
@ivan-c  all done here now

portal/config/eproms/Organization.json
```diff
@@ -2164,7 +2169,7 @@
         }
       ],
       "id": 28000,
-      "language": "pt_BR",
+      "language": "en_US",
```
ivan-c: I think this should probably be be added as an option (currently just the default for `id=28000`), by adding it to the `extension` array.

ivan-c: I think this should probably be be added as an option (currently just the pre-selected default for `id=28000`/`Brazil (Region/Country Site)`), by adding it to the `extension` array.

ivan-c: I also assume we don't want to enable `pt_BR` region-wide (`id=28000`/`Brazil (Region/Country Site)`)

mcjustin: I *think* what I have is correct here...
PCCTC's aim is to have all Brazilian orgs the same for language: english as the default, with portugese optional. It's a goal, but not an absolute - at some point there may be a brazilian org that wants it otherwise.
Our implementation of org config inheritance is opaque to me, @pbugni  and I have chatted about it a bit. But I think this change is correct & errs on the safe side.

PR #3413, opened 2019-08-13
TN-2202 in py3, dict.keys() returns a non-index-able type.

pbugni:
Any attempt to index the results of a dict.keys() call fails on py3 with:

```TypeError: 'dict_keys' object does not support indexing```

Fix by casting any `dict_keys` result to a `list` before attempting an index.

### DIFF
--- a/portal/models/questionnaire_response.py
+++ b/portal/models/questionnaire_response.py
@@ -225,7 +225,7 @@ class QuestionnaireResponse(db.Model):
             for answer in combined_answers:
 
                 # Add text answer before coded answer
-                if answer.keys()[0] == 'valueCoding':
+                if list(answer.keys())[0] == 'valueCoding':
 
                     # Prefer text looked up from code over sibling valueString answer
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: portal/models/questionnaire_response.py [hunk 1]
```diff
@@ -225,7 +225,7 @@ def document_answered(self):
             for answer in combined_answers:
 
                 # Add text answer before coded answer
-                if answer.keys()[0] == 'valueCoding':
+                if list(answer.keys())[0] == 'valueCoding':
```
ivan-c:
```suggestion
                if next(iter(answer.keys())) == 'valueCoding':
```


A more general (more suitable to iterators) approach might be to use `next()` to get the first value of it as an iterator
```
>>> answer = {'valueCoding': {'code':[]}}
>>> hasattr(answer.keys(), '__iter__')
True
>>> next(iter(answer.keys()))
'valueCoding'
```

PR #3461, opened 2019-09-16
TN-2274 CLI to merge duplicate users as described in TN-2240

pbugni:
Robust, careful process won't allow merge unless several criteria are met.  Asks admin user to verify each step of the process.

### DIFF
--- a/manage.py
+++ b/manage.py
@@ -479,7 +479,7 @@ def merge_users(src_id, tgt_id, actor):
         raise ValueError("Birth dates don't match; can't continue")
     if src_user.auth_providers.count() > 0:
         raise ValueError("extend to include auth_providers")
-    
+
     if src_user.identifiers != tgt_user.identifiers and (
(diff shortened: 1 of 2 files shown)


manage.py
```diff
@@ -447,3 +446,116 @@ def update_qnr_authored(qnr_id, authored, actor):
...
+    tgt_user = User.query.get(tgt_id)
+
+    if not all((src_user, tgt_user)) or (
+            src_user.birthdate != tgt_user.birthdate):
+        raise ValueError("Birth dates don't match; can't continue")
+    if src_user.auth_providers.count() > 0:
+        raise ValueError("extend to include auth_providers")
+
+    if src_user.identifiers != tgt_user.identifiers and (
+            click.confirm("Add identifiers \n\t{} \nto \n\t{}".format(
```
ivan-c: You may be able to fold some of these confirmations into decorators ([using `prompt=True`](http://click.palletsprojects.com/en/5.x/options/#option-prompting)), but would likely lose the granularity

PR #3466, opened 2019-09-18
TN-2194 don't allow future dates in questionnaire response `authored`

pbugni:


achen2401:
@pbugni Paul, I added my frontend fix just now.  Let me know if you experience any issue.

PR #3471, opened 2019-09-20
TN-2249 - Add analytics to menu. I think most of this is in the right…

mcjustin:
… direction. Perhaps access_strategies should be used - but that's for cards/tiles only, I think. Will add to eproms in a later sprint.

### DIFF
--- a/portal/templates/portal_wrapper.html
+++ b/portal/templates/portal_wrapper.html
@@ -79,6 +79,11 @@
                             <li><a href="{{PORTAL}}/scheduled_jobs">{{ _("Scheduled Jobs") }}</a></li>
                             <li><a href="{{PORTAL}}/settings">{{ _("Settings") }}</a></li>
                             {% endif %}
+                            <!-- fixme use the new 'analytics' intervention's 'link_url' -->
+                            <!-- fixme consider intervention.display_for_user(user) ? -->
+                            {% if user and (user.has_role(ROLE.STAFF.value) or user.has_role(ROLE.STAFF_ADMIN.value)  or user.has_role(ROLE.ANALYST.value) or user.has_role(ADMIN.value)) %}
+                            <li class="resources-menu-item"><a href="{{PORTAL}}/?fixme=intervention_link_url">{{ _("Analytics") }}</a></li>
+                            {% endif %}
                             {% if user and user.has_role(ROLE.RESEARCHER.value) %}
                             <li><a href="{{PORTAL}}/research">{{ _("Research Data") }}</a></li>
                             {% endif %}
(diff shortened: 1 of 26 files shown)


portal/templates/portal_wrapper.html
```diff
@@ -79,6 +79,11 @@
                             <li><a href="{{PORTAL}}/scheduled_jobs">{{ _("Scheduled Jobs") }}</a></li>
                             <li><a href="{{PORTAL}}/settings">{{ _("Settings") }}</a></li>
                             {% endif %}
+                            <!-- fixme use the new 'analytics' intervention's 'link_url' -->
+                            <!-- fixme consider intervention.display_for_user(user) ? -->
+                            {% if user and (user.has_role(ROLE.STAFF.value) or user.has_role(ROLE.STAFF_ADMIN.value)  or user.has_role(ROLE.ANALYST.value) or user.has_role(ADMIN.value)) %}
```
ivan-c: Missing `ROLE` here...
```suggestion
                            {% if user and (user.has_role(ROLE.STAFF.value) or user.has_role(ROLE.STAFF_ADMIN.value)  or user.has_role(ROLE.ANALYST.value) or user.has_role(ROLE.ADMIN.value)) %}
```

ivan-c: This might be a little more succinct...
```suggestion
                            {% if user and (any((user.has_role(r.value) for r in (ROLE.ADMIN, ROLE.ANALYST, ROLE.STAFF_ADMIN, ROLE.STAFF)))) %}
```

portal/templates/portal_wrapper.html
```diff
@@ -79,6 +79,11 @@
                             <li><a href="{{PORTAL}}/scheduled_jobs">{{ _("Scheduled Jobs") }}</a></li>
                             <li><a href="{{PORTAL}}/settings">{{ _("Settings") }}</a></li>
                             {% endif %}
+                            <!-- fixme use the new 'analytics' intervention's 'link_url' -->
+                            <!-- fixme consider intervention.display_for_user(user) ? -->
```
pbugni: I would not suggest `display_for_user()` - too expensive, and designed for patients, for which this doesn't apply

pbugni:
@mcjustin give this a look.  You should simply need to add the analytics client like any other, and give it a well defined `link_url` to see it in the menu.

mcjustin:
> 
> 
> @mcjustin give this a look. You should simply need to add the analytics client like any other, and give it a well defined `link_url` to see it in the menu.

Great, thanks @pbugni

PR #3474, opened 2019-09-24
TN-2289 modify user.has_role() to check for role in list of roles as …

pbugni:
…well as current behaior

### DIFF
--- a/portal/models/user.py
+++ b/portal/models/user.py
@@ -1176,10 +1176,10 @@ class User(db.Model, UserMixin):
 
             """
             if (
-                not (acting_user.has_role(ROLE.ADMIN.value) or
-                     acting_user.has_role(ROLE.APPLICATION_DEVELOPER.value))
-                and (acting_user.has_role(ROLE.STAFF.value)
-                     or acting_user.has_role(ROLE.STAFF_ADMIN.value))
+                not acting_user.has_role((
+                    ROLE.ADMIN.value, ROLE.APPLICATION_DEVELOPER.value))
+                and acting_user.has_role((
+                    ROLE.STAFF.value, ROLE.STAFF_ADMIN.value))
                 and user.id == acting_user.id
             ):
                 raise ValueError(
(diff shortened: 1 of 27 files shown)


portal/models/user.py
```diff
@@ -1671,8 +1670,13 @@ def check_role(self, permission, other_id):
 
         abort(401, "Inadequate role for {} of {}".format(permission, other_id))
 
-    def has_role(self, role_name):
-        return role_name in [r.name for r in self.roles]
+    def has_role(self, role_name_or_list):
```
ivan-c: I think if you used [arg unpacking(?)](http://book.pythontips.com/en/latest/args_and_kwargs.html#usage-of-args) you wouldn't need to do any introspection (when called with individual roles) or modify any existing method calls
```suggestion
    def has_role(self, *role_names):
```

Official python docs:
[Arbitrary Argument Lists](https://docs.python.org/dev/tutorial/controlflow.html#arbitrary-argument-lists)

portal/models/user.py
```diff
@@ -1671,8 +1670,13 @@ def check_role(self, permission, other_id):
...
+        """Given role (or list of) returns true if user has at least one"""
+        if isinstance(role_name_or_list, str):
+            return role_name_or_list in [r.name for r in self.roles]
+        for item in role_name_or_list:
+            if item in [r.name for r in self.roles]:
```
ivan-c: might as well define that inline list as a variable, defined at the beginning of the method; easier to read and doesn't re-create a list of the user's roles for every given role

pbugni: well, for those of us who love list comprehension, i'm not sure it's more legible, but i've plugged in a set in the name of "efficiency" - hard to imagine it'll matter with such a small 'n'

portal/models/user.py
pbugni: excellent suggestion, @ivan-c  - haven't used that trick in years, and seem to have overlooked it.

PR #3485, opened 2019-10-01
fix JS error when parsing audit log (followup fix)

achen2401:
address JS runtime error as seen when parsing audit log.  Example:  https://truenth-test.cirg.washington.edu/patients/patient_profile/11764
Followup fix to this fix:
https://github.com/uwcirg/truenth-portal/commit/1a15e8672bf3

### DIFF
--- a/portal/static/js/src/profile.js
+++ b/portal/static/js/src/profile.js
@@ -1910,22 +1910,25 @@ export default (function() {
                 self.initUserRoles();
             },
             initAuditLogSection: function() {
-                var self = this;
+                var self = this, errorMessage = "";
                 $("#profileAuditLogTable").html(Utility.getLoaderHTML());
                 $("#profileAuditLogTable .loading-message-indicator").show();
                 this.modules.tnthAjax.auditLog(this.subjectId, {useWorker:true}, function(data) {
                     if (data.error) {
-                        $("#profileAuditLogTable").html("");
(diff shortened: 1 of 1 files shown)


portal/static/js/src/profile.js
```diff
@@ -1910,22 +1910,25 @@ export default (function() {
...
-                var self = this;
+                var self = this, errorMessage = "";
                 $("#profileAuditLogTable").html(Utility.getLoaderHTML());
                 $("#profileAuditLogTable .loading-message-indicator").show();
                 this.modules.tnthAjax.auditLog(this.subjectId, {useWorker:true}, function(data) {
                     if (data.error) {
-                        $("#profileAuditLogTable").html("");
-                        $("#profileAuditLogErrorMessage").text(i18next.t("Problem retrieving audit log from server."));
-                        return false;
+                        errorMessage = i18next.t("Problem retrieving audit log from server.");
```
ivan-c:
```suggestion
                        errorMessage = i18next.t("Error retrieving data from server");
```

portal/static/js/src/profile.js
```diff
@@ -1910,22 +1910,25 @@ export default (function() {
...
                     }
                     if (!data.audits || data.audits.length === 0) {
+                        errorMessage = i18next.t("No audit log item found.");
```
ivan-c: I think this is an error message we can reuse...
```suggestion
                        errorMessage = i18next.t("no data returned");
```

PR #3487, opened 2019-10-02
TN-2193 indefinite QB in adherence report

pbugni:
Add "none of the above" Questionnaire Bank, and extended classification enum with "other".
TN-2303 Prevent repeated QNR->QB lookup with marker for n/a.
TN-2304 Add indefinite to adherence report

### DIFF
--- a/manage.py
+++ b/manage.py
@@ -34,6 +34,7 @@ from portal.models.i18n_utils import (
 from portal.models.intervention import add_static_interventions
 from portal.models.organization import add_static_organization
 from portal.models.qb_timeline import invalidate_users_QBT
+from portal.models.questionnaire_bank import add_static_questionnaire_bank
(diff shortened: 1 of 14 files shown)


pbugni:
> Looks good!
> 
> Question- why not change the QB lookup code to tolerate null QBs when attempting to find the QB for a QuestionnaireResponse?(instead of assigning QNRs to a "special" QB)

The lookup code does tolerate null QBs - but we need a clear way to distinguish those QNRs that have yet to be sync'd with the appropriate QBs, and those that just don't have one.

Anytime we want to *force* a (re)evaluation of which QB a given QNR should be attached to, we set `QNR->qb_id = None` and then the next cache / lookup will evaluate.  Post of a new QNR, or change of consent or new procedure or ... all trigger such invalidation.

PR #3490, opened 2019-10-04
TN-2251-eproms-analytics-intervention

mcjustin:
https://jira.movember.com/browse/TN-2251

### DIFF
--- a/portal/config/eproms/Intervention.json
+++ b/portal/config/eproms/Intervention.json
@@ -98,6 +98,13 @@
       "public_access": false,
       "subscribed_events": 1,
       "resourceType": "Intervention"
+    },
+    {
+      "description": "Analytics",
+      "name": "analytics",
+      "public_access": false,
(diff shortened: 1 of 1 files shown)


portal/config/eproms/Intervention.json
```diff
@@ -98,6 +98,13 @@
       "public_access": false,
       "subscribed_events": 1,
       "resourceType": "Intervention"
+    },
+    {
+      "description": "Analytics",
+      "name": "analytics",
+      "public_access": false,
+      "subscribed_events": 1,
```
pbugni: Not a problem if you're ignoring, but did you intend to subscribe these "interventions" to events such as logout?

ivan-c: @mcjustin correct me if I'm wrong, but I believe all cPRO interventions listen for event callbacks- specifically for the logout event

PR #3499, opened 2019-10-08
TN-2314 add second "low_priority" celery worker queue

pbugni:
Adds a second celery worker, dedicated to the "low_priroty" queue (namely, any celery tasks that are not directly user or health check related).

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -36,6 +36,11 @@ unit_tests:
     TOXENV: py3
   <<: *unit_test_definition
 
+unit_tests_celery_background:
+  variables:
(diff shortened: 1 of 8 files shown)


pbugni:
@ivan-c I don't get why the unit-tests are failing with the second celery worker and queue.  Using the app and scheduled jobs, it looks to be well behaved, I *think* this is a symptom of our test environment, but that's not very comforting.
Worth noting, /celery-test hits the default queue and /celery-info hits the low_priority one, for hand experimentation.

pbugni:
@ivan-c can we merge this now?

PR #3506, opened 2019-10-15
TN-2298 remove other story link from LE main page

achen2401:
Per latest feedback to this story:
https://jira.movember.com/browse/TN-2298

- remove "READ OTHER STORIES" link from LE main page
- styling fix of button layout

### DIFF
--- a/portal/gil/static/js/lived_experience.js
+++ b/portal/gil/static/js/lived_experience.js
@@ -17,9 +17,6 @@ $(document).ready(function() {
       });
     }
     if ($("main").attr("data-section") === "livedexperience") {
-      $("#lnReadMoreStory").addClass("disabled").on("click", function(e) {
-          e.preventDefault();
-          return false;
-      });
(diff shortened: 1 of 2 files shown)


portal/gil/static/js/lived_experience.js
```diff
@@ -17,9 +17,6 @@ $(document).ready(function() {
       });
     }
     if ($("main").attr("data-section") === "livedexperience") {
-      $("#lnReadMoreStory").addClass("disabled").on("click", function(e) {
-          e.preventDefault();
-          return false;
-      });
+      $("#lnReadMoreStory").hide();
```
ivan-c: Why not hide this element with CSS? Seems like it would be quicker/less finicky

achen2401: ivan, sorry, didn't see this comment until now.  I will just make another PR for this.  agree.

PR #3512, opened 2019-10-22
TN-2330 catch use of future dates in QNR submission

pbugni:
Report error w/ 400 status on authored validation exception.
Also, give 1 min buffer for external system clock drift

### DIFF
--- a/portal/views/assessment_engine.py
+++ b/portal/views/assessment_engine.py
@@ -3,7 +3,6 @@ from datetime import datetime
 
 from flask import (
     Blueprint,
-    Response,
     abort,
     current_app,
     flash,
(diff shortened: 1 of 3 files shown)


portal/views/assessment_engine.py
```diff
@@ -814,11 +811,13 @@ def assessment_update(patient_id):
...
         QuestionnaireResponse.validate_document(updated_qnr)
-    except jsonschema.ValidationError as e:
+        QuestionnaireResponse.validate_authored(
+            FHIR_datetime.parse(updated_qnr.get('authored')))
+    except (jsonschema.ValidationError, NoFutureDates) as e:
         return jsonify({
             'ok': False,
-            'message': e.message,
-            'reference': e.schema,
+            'message': str(e),
```
ivan-c: Surprised you had to cast this now, if not before...

pbugni: @ivan-c  Correct - this was an overlooked python2-3 required change.  Apparently this catch block hasn't been hit.?!

PR #3517, opened 2019-10-22
Split celery fixtured/unfixtured jobs

ivan-c:
Add tox environment and GitlabCI job for tests requiring celery running in background during tests

#### Todo
- [ ] Add conditional `pytest.skip` check to each file instead of `--ignore` (in `tox.ini`)
- [ ] Move as many tests as possible to "fixture" tox environment

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -34,7 +34,8 @@ def app(request):
 
 
 @pytest.fixture(scope='session')
-def initialized_db():
+def initialized_db(app):
+    """Create database schema"""
     db.create_all()
 
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,6 +1,7 @@
 # test plugin
 # https://docs.pytest.org/en/latest/writing_plugins.html#conftest-py-plugins
 import pytest
+from portal.database import db
 from portal.factories.app import create_app
 from portal.factories.celery import create_celery
 
@@ -30,6 +31,11 @@ def app(request):
 
     request.addfinalizer(teardown)
     return app_
+
+
+@pytest.fixture(scope='session')
+def initialized_db():
+    db.create_all()
 
 
 @pytest.fixture(scope='session')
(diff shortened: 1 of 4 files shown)


tests/conftest.py
```diff
@@ -32,6 +33,11 @@ def teardown():
     return app_
 
 
+@pytest.fixture(scope='session')
+def initialized_db():
+    db.create_all()
```
pbugni: I expect you'll either want to extend this to add the static concepts et al, or add an alternative fixture to do so.
See https://github.com/uwcirg/truenth-portal/blob/develop/tests/__init__.py#L497 - many scenarios will require that static data to be resident in the db.

ivan-c: Yup, we'll definitely have to add those setup steps as fixtures. it was an open question in my mind whether to create small fixtures that only test the bare minimum, or larger ones. I'm leaning towards creating minimal fixtures, but I'm worried they will be harder to maintain than fewer, larger fixtures.

ivan-c:
@pbugni tagged you for a followup review; LMK if you think any of the Todos I have in the PR description should be part of the PR

tests/test_celery.py
```diff
@@ -22,6 +23,7 @@ def test_celery_add(celery_app, celery_worker, client):
     assert response.json['result'] == x + y
 
 
+@pytest.mark.skip(reason="locking up with celery fixtures")
```
pbugni: this did work with the celery fixtures, provided (https://github.com/uwcirg/truenth-portal/blob/feature/multi-celery-queues_fixture-job/tests/conftest.py#L54) both queues are in place.

ivan-c: This PR is based on yours (I branched from your branch) so it should include that change. I [saw multiple failures](https://gitlab.com/uwcirg/truenth-portal/-/jobs/331143964) on this test and since it wasn't actually a test I skipped it. 

I'm not sure why it would fail now, but maybe we need to reduce the fixture scope after I added more tests to the fixtured test suite?

pbugni: Not clear to me either.  It does require the `low_priority` queue, but so do several other tests that now pass.
I'm okay with keeping the skip in there for now.

PR #3527, opened 2019-10-31
Feature branch

waynelai1228:
I don't know if the .css .json and .pot files should go into this pull request; they seem to be auto-generated.
Other than that, I checked the other files and they are good.


waynelai1228:
Done deleting every unneeded files

PR #3532, opened 2019-11-05
Feature/migrate to pytest fixture

waynelai1228:
changed three files in the tests/ directory!


waynelai1228:
app_logger is passed into test functions instead of app

PR #3537, opened 2019-11-08
Resolve Pytest Conflicts

waynelai1228:
update the requirement.dev.txt to reflect the latest pytest version
update the deprecated usage for error message in test_date_tools.py
update the deprecated usage for error message in test_organization.pyw

### DIFF
--- a/requirements.dev.txt
+++ b/requirements.dev.txt
@@ -17,7 +17,7 @@ psycopg2-binary==2.8.3
 py==1.8.0                 # via tox
 pygments==2.4.2           # via sphinx
 pyparsing==2.4.1.1          # via packaging
-pytest==4.6.3
+pytest==5.2.2
 pytest-flask==0.15.0
 selenium==3.141.0
 snowballstemmer==2.0.0    # via sphinx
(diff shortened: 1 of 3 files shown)


pbugni:
Hmm, I don't fault @waynelai1228 as it's a library choice, but I'm surprised to see the string form (calling str(e)) of these exceptions no longer provides the exception text - that's the documented mechanism to get the details.

This looks to be a "feature" of py.test.  For example, this works as expected outside a test context:
```
>>> from werkzeug.exceptions import BadRequest
>>> try:
...   raise BadRequest('the reason')
... except BadRequest as e:
...   str(e)
... 
'400 Bad Request: the reason'
>>> try:                            
...   raise BadRequest('the reason')
... except BadRequest as e:
...   str(e.value)
... 
  File "<stdin>", line 2, in <module>
werkzeug.exceptions.BadRequest: 400 Bad Request: the reason

During handling of the above exception, another exception occurred:

  File "<stdin>", line 4, in <module>
AttributeError: 'BadRequest' object has no attribute 'value'
```

Means we have different ways of extracting the string in test code from elsewhere.  *sigh*, guess we can live with it.

PR #3538, opened 2019-11-08
Resolve Validators Conflict

waynelai1228:
update the requirement.txt to reflect the latest validators version
remove the deprecated attribute require_tld in validators.url() for portal/views/client.py
remove the deprecated test require_tld for test_auth.py


waynelai1228:
> @waynelai1228 can you evaluate the remaining `require_tld` in `portal/views/client.py`?
@pbugni There are two separate validators in `portal/views/client.py`.
The one causing the problems is `validators.url` (bumped to 0.14.0)
The other validators is from WTForm, which still has require_tld field.

PR #3545, opened 2019-11-13
Add missing prod block for `celeryworkerslow`

pbugni:


docker/docker-compose.prod.yaml
```diff
@@ -35,6 +35,21 @@ services:
           --app portal.celery_worker.celery
           --loglevel debug
 
+  celeryworkerslow:
+    restart: unless-stopped
+    environment:
+      # Number of tasks a celery worker should run before being recreated
+      # https://docs.celeryproject.org/en/latest/userguide/configuration.html#std:setting-worker_max_tasks_per_child
+      CELERYD_MAX_TASKS_PER_CHILD: 10
```
pbugni: I wonder about the value of this, in both celeryworker* blocks.  There must be a cost to restarting them, and 10 feels very small to me.

ivan-c: Agreed, maybe something like 100? I spot-checked a caching job (on flower) and it took ~6 seconds to complete

docker/docker-compose.prod.yaml
```diff
@@ -35,6 +35,21 @@ services:
...
+    # Set lower CPU priority to prevent blocking web service
+    cap_add:
+      - SYS_NICE
+    command:
+      nice -n 15
+        celery worker
+          --app portal.celery_worker.celery
```
ivan-c: I think you need to name the queue here...
```suggestion
          --app portal.celery_worker.celery
          --queue low_priority
```

PR #3552, opened 2019-11-19
Feature/portal external asset api update

achen2401:
address https://jira.movember.com/browse/TN-2374:

Modify portal code to use the updated LR api that accepts new parameters to return HTML content for a LR item

@pbugni @ivan-c Your feedback on this is much appreciated.  I hope the changes are in the right direction.


portal/views/external_assets.py
```diff
@@ -2,31 +2,79 @@
...
     params = {
         'uuid': uuid,
         'version': 'latest',
     }
-    return requests.get(url, params=params).json()['asset']
+    response, error_msg = get_request(url, params)
+    if error_msg:
+        return error_msg
+    if isinstance(response, dict):
+        if not response['asset']:
```
pbugni: same, `if 'asset' not in response':`

portal/views/external_assets.py
```diff
@@ -2,31 +2,79 @@
...
+
+
+def get_request(url, params=None):
+
+    error_msg = ''
+    try:
+        response = time_request(url, params).json()
+    except ValueError:  # raised when no json is available in response
+        if response.status_code == 200:
+            return response.text
```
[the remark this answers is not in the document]
achen2401: thank you, paul, addressed, also to your other feedbacks below

portal/views/external_assets.py
```diff
@@ -2,31 +2,79 @@
...
+        else:
+            error_msg = (
+                "Could not retrieve remote content - " "{} {}".format(
+                    response.status_code, response.reason))
+    except ConnectionError:
+        error_msg = (
+                    "Could not retrieve remove content - Server could not be "
+                    "reached")
+    if error_msg:
+        current_app.logger.error(error_msg + ": {}".format(url))
```
ivan-c: I don't think this Liferay-specific error handling code is strictly necessary. I would say the `ValueError` exception bubbled up as a `500`, would be enough logging (we log all exceptions), but the Liferay URL is essential for debugging.
Maybe just catch the `ValueError`, call `current_app.logger.error()` then `reraise` the caught exception?

Serving valid content is the responsibility of Liferay and I would prefer we don't normalize Liferay's failures to deliver content properly.

portal/views/external_assets.py
```diff
@@ -2,31 +2,79 @@
...
         'version': 'latest',
     }
-    return requests.get(
-        url, params=params).json()['asset']
+    response, error_msg = get_request(url, params)
+    if error_msg:
+        return error_msg
+    if isinstance(response, dict):
+        if 'results' not in response or not len(response['results']):
+            abort(404, 'Remote content not found for tag {}'.format(tag))
```
ivan-c: I don't think we should change the Liferay response code if our eventual goal is to have the frontend retrieve content directly from Liferay via AJAX.

If we want `404`s we should ask Victor to change the Liferay API ("find first content item by tag") to `404` on no content item found

portal/views/external_assets.py
pbugni: It is a slippery slope - I think we'd do well to fail noisily when there are problems.
It is very helpful to include any details in the exception text that might save time tracking down the source.

portal/views/external_assets.py
pbugni: Agreed.

PR #3554, opened 2019-11-19
Update deprecated sqlalchemy code

waynelai1228:
will look at other places to see if deprecated code still exist

### DIFF
--- a/tests/test_scheduled_job.py
+++ b/tests/test_scheduled_job.py
@@ -9,6 +9,7 @@ from portal.extensions import db
 from portal.models.role import ROLE
 from portal.models.scheduled_job import ScheduledJob
 from portal.tasks import test as test_task
+from sqlalchemy.orm import session
 from tests import TestCase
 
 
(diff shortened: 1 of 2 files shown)


tests/test_scheduled_job.py
```diff
@@ -175,4 +176,4 @@ def test_job_trigger(self):
         resp = self.client.post('/api/scheduled_job/{}/trigger'.format(job.id))
         assert resp.status_code == 200
 
-        db.session.close_all()
+        session.close_all_sessions()
```
pbugni: I would expect `db.session` is referencing the same symbol.  Did it not work with `db.session.close_all_sessions`?

waynelai1228: db.session.close_all_sessions does not work because db.session is a Session object (sqlalchemy.orm.session.Session) which does not have close_all_sessions method.
The close_all_sessions method is in at sqlalchemy.orm.session.close_all_sessions.

waynelai1228:
@ivan-c I was thinking maybe I could find some other warnings, but I currently am not seeing anymore.

PR #3557, opened 2019-11-20
Fixing utcnow() issues

waynelai1228:


portal/date_tools.py
```diff
@@ -85,7 +85,7 @@ def parse(data, error_subject=None, none_safe=False):
     @staticmethod
     def now():
         """Generates a FHIR compliant datetime string for current moment"""
-        return datetime.utcnow().isoformat() + 'Z'
+        return datetime.now(tz=timezone.utc).isoformat() + 'Z'
```
pbugni: @waynelai1228 these two produce slightly different results, as the new one includes a timezone offset.  the `Z` stands for zulu time - I expect you'll not want both.  Using a python shell here, I can see the differences:
```
>>> from datetime import date, datetime, timezone
>>> datetime.now(tz=timezone.utc).isoformat()
'2019-11-20T18:29:48.595737+00:00'
>>> datetime.utcnow().isoformat()
'2019-11-20T18:30:00.627613' 
```

waynelai1228: @pbugni 
Fixed!

PR #3561, opened 2019-11-25
TN-2371 Cleanup DSP Factors QuestionnaireResponse

ivan-c:
Add migration to cleanup DSP Factors QuestionnaireResponses


portal/migrations/versions/908731b85112_.py
```diff
@@ -0,0 +1,85 @@
+from alembic import op
+import sqlalchemy as sa
+from sqlalchemy.orm import sessionmaker
+
+from portal.models.questionnaire_response import QuestionnaireResponse
+
+
+"""empty message
```
pbugni: If you put the comment block at the top, with "empty message" filled in, it'll show when running the upgrade.
We should probably reorder the template - I just haven't taken the 12 seconds to do so...

portal/migrations/versions/908731b85112_.py
```diff
@@ -0,0 +1,85 @@
...
+            ("questionnaire", "reference")
+        ].astext.endswith(instrument_id)
+    ).order_by(QuestionnaireResponse.id)
+
+    for qnr in questionnaire_responses:
+        qnr_json = fixup_dsp_factors(qnr.document)
+        # "Reset" QNR to save updated data
+        # Todo: fix JSONB mutation detection
+        # See https://bugs.launchpad.net/fuel/+bug/1482658
+        qnr.document = {}
```
pbugni: I've had good luck with `copy.deepcopy()` - see 69af4bd9ac9a_.py

PR #3566, opened 2019-11-26
TN-1620, TN-2142 extensions in qnr bundles

pbugni:
Adding FHIR extensions in the QNR bundles to support TN-1620 (visit name) and TN-2142 (partially complete once in-progress QNRs are expired).


portal/models/questionnaire_response.py
```diff
@@ -261,6 +267,40 @@ def document_answered(self):
...
+        No place within the FHIR spec to associate 'visit name' nor a
+        'status' as per business rules (i.e. 'in-progress' becomes
+        'partially completed' once the associated QB expires).  Use FHIR
+        `extension`s to pass these fields to clients like the front end
+
+        @returns list of FHIR extensions for this instance, typically one for `visit_name`
+        """
+        from .qb_timeline import expires  # avoid cycle
+        results = []
+        if self.questionnaire_bank_id is not None:
```
ivan-c: slightly less indented if this were a guard clause (`if self.questionnaire_bank_id is None:;return []`)

PR #3578, opened 2019-12-10
TN-2287 correct date validation handling for manual entry

achen2401:
Address https://jira.movember.com/browse/TN-2287
Issue: 
In entering completion date for Paper Prom, submission button still remained disabled **after** correcting an invalid date entry.

Changes include

- clear error message after incorrect date entry has been corrected
- add bootstrap error css class to input validation
- remove redundant code for validating date

### DIFF
--- a/portal/static/js/src/profile.js
+++ b/portal/static/js/src/profile.js
@@ -1791,6 +1791,12 @@ export default (function() {
                     this.manualEntry.completionDate = this.manualEntry.consentDate;
                 }
             },
+            resetManualEntryFormValidationError: function() {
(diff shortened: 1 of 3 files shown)


portal/static/js/src/profile.js
```diff
@@ -1791,6 +1791,12 @@ export default (function() {
                     this.manualEntry.completionDate = this.manualEntry.consentDate;
                 }
             },
+            resetManualEntryFormValidationError: function() {
+                this.manualEntry.errorMessage = "";
```
ivan-c: Would `null` make a better default?

PR #3579, opened 2019-12-10
TN-2417 provide ugly workaround to failed query

pbugni:
Hopefully a temporary workaround, run a second query for the entire json document if the query for the json path field fails.

### DIFF
--- a/portal/models/questionnaire_response.py
+++ b/portal/models/questionnaire_response.py
@@ -340,15 +340,41 @@ class QNR_results(object):
             ).filter(
                 QuestionnaireResponse.qb_iteration == self.qb_iteration)
         self._qnrs = []
+        requery = False
         for qnr in query:
+            if qnr.instrument_id is None:
+                current_app.logger.warning(
+                    "QNR {} query returned no instrument".format(qnr.id))
+                instrument = None
+                requery = True
+            else:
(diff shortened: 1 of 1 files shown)


pbugni:
@ivan-c should we include a rollback of yesterdays SQLAlchemy downgrade attempt, as that didn't prove to be the problem?

ivan-c:
> @ivan-c should we include a rollback of yesterdays SQLAlchemy downgrade attempt, as that didn't prove to be the problem?

I'd prefer to minimize change to `master` until we can figure this out. I'll merge what we have in master into develop and restore Wayne's changes there; they'll become part of the next release.

pbugni:
> > @ivan-c should we include a rollback of yesterdays SQLAlchemy downgrade attempt, as that didn't prove to be the problem?
> 
> I'd prefer to minimize change to `master` until we can figure this out. I'll merge what we have in master into develop and restore Wayne's changes there; they'll become part of the next release.

Sure, sounds wise.

pbugni:
> Looks good, thanks for branching from master!
> 
> Is there anything else (metadata, system state) we would want to log? Nothing comes to mind, but I'm not sure we'll be able to reproduce with QNR.id only

I really don't know what else to log.  We can't reproduce at all, but at least we'll have a pointer to the db row that behaved poorly.

PR #3581, opened 2019-12-11
TN-2417  prevent mutations on cached query objects 

pbugni:


portal/models/questionnaire_response.py
```diff
@@ -230,19 +231,26 @@ def validate_document(document):
...
+        session cached objects.
+
         """
-        Return a QuestionnaireResponse populated with text answers based on codes in valueCoding
-        """
-        instrument_id = self.document['questionnaire']['reference'].split('/')[-1]
+        instrument_id = self.document['questionnaire']['reference'].split(
+            '/')[-1]
         questionnaire = Questionnaire.find_by_name(name=instrument_id)
+        document = copy.deepcopy(self.document)
```
ivan-c: Not sure how much of a change it would be, but could be more performant if the `deepcopy` happened after the check for a Questionnaire, since we won't need to modify it anyway.

pbugni: not true - look at how it's used in the view function below - the return value gets modified regardless of the questionnaire.

PR #3582, opened 2019-12-12
TN-2287 followup fixes - display error when completion date is not within window

achen2401:
Followup fixes for this story:  https://jira.movember.com/browse/TN-2287

Noting where testing that when completion date entered was not within expected window,  error message **did not** display as would have expected - only seeing submission button being disabled, which can be confusing to user.

Fixes include:
-  display error message when completion date for paper manual entry is not entered within a valid PROM window
- remove loading indicator when there is error
- reset error message to null instead of empty string when clearing error, fix based on previous feedback


achen2401:
will re-open after more thorough testing

PR #3583, opened 2019-12-12
TN-2405 Handle more than two Research Protocols

pbugni:
Extended logic to handle more than two research protocols using generators.
Caught details previously overlooked for transitioning during protocol change if user had started but only completed questionnaires belonging to both RPs for the applicable visit.

### DIFF
--- a/portal/models/qb_timeline.py
+++ b/portal/models/qb_timeline.py
@@ -380,17 +380,18 @@ def ordered_qbs(user, classification=None):
                 # if however, the user has already posted QNRs for the
                 # currentRP one of two things needs to happen:
                 #
-                #   1. if *any* of the QNRs are for instruments which
-                #      deterministically (i.e over RPs {v2,v3}, "epic23" vs
-                #      "epic26" deterministically define the QB where as
-                #      "eortc" is non-deterministic, as it belongs to both)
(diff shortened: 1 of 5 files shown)


portal/models/qb_timeline.py
```diff
@@ -295,86 +360,118 @@ def ordered_qbs(user, classification=None):
...
+            if nextRPD and curRPD.retired < users_expiration:
+                # if there's a nextRP and curRP is retired before the
+                # user's QB for the curRP expires, we look to transition
+                # the user to the nextRP.
+                #
+                # if however, the user has already posted QNRs for the
+                # currentRP one of two things needs to happen:
+                #
+                #   1. if *any* of the QNRs are for instruments which
+                #      deterministically (i.e over RPs {v2,v3}, "epic23" vs
```
ivan-c: looks like *deterministically* used twice...

pbugni: right, i can't seem to wordsmith this one.  i added that on @mcjustin's request that i clarify what deterministic meant in this context.
do you have a suggestion, or should we call it good enough?

pbugni: @ivan-c  Reworked as a footnote, better?

PR #3584, opened 2019-12-12
TN-2420-3rdof3NewIronmanOrgs

mcjustin:
https://jira.movember.com/browse/TN-2420

### DIFF
--- a/portal/config/eproms/Organization.json
+++ b/portal/config/eproms/Organization.json
@@ -1827,7 +1827,7 @@
     {
       "extension": [
         {
-          "timezone": "America/Denver",
+          "timezone": "America/Edmonton",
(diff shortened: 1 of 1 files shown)


portal/config/eproms/Organization.json
```diff
@@ -1824,6 +1824,34 @@
       },
       "resourceType": "Organization"
     },
+    {
+      "extension": [
+        {
+          "timezone": "America/Denver",
```
ivan-c: I think this [timezone](https://en.wikipedia.org/wiki/List_of_tz_database_time_zones) is technically more accurate, but makes no real difference
```suggestion
          "timezone": "America/Edmonton",
```

PR #3586, opened 2019-12-13
Seconds since last /ping

pbugni:
Implement simple usage tracker to simplify checks since last call to /ping was recorded.

### DIFF
--- a/portal/usage_monitor.py
+++ b/portal/usage_monitor.py
@@ -1,0 +1,21 @@
+"""API to record and track seconds since last request"""
+from flask import current_app
+import redis
+
+
+def _connection():
+    # Documented to be pooled at the library level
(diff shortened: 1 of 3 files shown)


pbugni:
It's a start.  We may need to call `mark_usage()` from additional views, if '/ping' isn't adequate?

portal/usage_monitor.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+def mark_usage():
+    conn = _connection()
+    conn.set(name=KEY, value="ping")
+
+
+def last_usage():
+    conn = _connection()
+    return conn.object(key=KEY, infotype='idletime')
```
ivan-c: neat! didn't know [this](https://redis.io/commands/object) was surfaced by redis (thought you would have to record/manage timestamps)

PR #3591, opened 2019-12-16
TN-2386 include List-Unsubscribe outbound email header.

pbugni:


portal/models/message.py
```diff
@@ -139,6 +151,7 @@ def send_message(self, cc_address=None):
         """
         message = Message(
             subject=self.subject,
+            extra_headers=extra_headers(),
```
ivan-c: Will these headers only be added to reminder emails, or all emails from the system (eg flask-user emails)?
(I think only reminder emails, but want to be sure...)

portal/models/message.py
```diff
@@ -79,6 +79,18 @@ def log_message(message, app):
...
+def extra_headers():
+    """Function to deliver any configured extra_headers for outbound email
+
+    Returns an empty dict, or one including any desired extra_headers, such
+    as a ``List-Unsubscribe`` header.
+
+    """
+    # TN-2386 Include List-Unsubscribe header to improve spam scores
+    return {'List-Unsubscribe': "<mailto:{}?subject=unsubscribe>".format(
+        current_app.config['CONTACT_SENDTO_EMAIL'])}
```
ivan-c: NB `CONTACT_SENDTO_EMAIL` defaults to `None`

May want to omit inclusion of this header if unconfigured, otherwise every system (including dev) will require this config value (or trigger an exception on send).

PR #3592, opened 2019-12-16
Feature/convert unittest functions to fixture

waynelai1228:
Awaiting resolution for test_create_account

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -346,7 +346,7 @@ def add_user(app, initialized_db):
 
 @pytest.fixture
 def service_user(initialize_static, test_user):
-    sponsor = test_user
+    sponsor = db.session.merge(test_user)
     service_user = sponsor.add_service_account()
     with SessionScope(db):
         db.session.add(service_user)
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -75,6 +75,38 @@ def shallow_org_tree():
     OrgTree.invalidate_cache()
 
 
+@pytest.fixture(scope="function")
+def deepen_org_tree(shallow_org_tree):
+    """Create deeper tree when test needs it"""
+    org_l2 = Organization(id=1002, name='l2', partOf_id=102)
+    org_l3_1 = Organization(id=10031, name='l3_1', partOf_id=1002)
+    org_l3_2 = Organization(id=10032, name='l3_2', partOf_id=1002)
+    with SessionScope(db):
+        [db.session.add(org) for org in (org_l2, org_l3_1, org_l3_2)]
+        db.session.commit()
+    # As orgs were just added, make sure they're loaded on next OrgTree call
+    OrgTree.invalidate_cache()
+
+    yield
+
+    # After using this fixture, invalidate OrgTree so subsequent tests
+    # don't find stale entries, namely dead references to the orgs created
+    # above.
+    OrgTree.invalidate_cache()
+
+
+@pytest.fixture
+def prep_org_w_identifier():
+    o = Organization(name='test org')
+    i = Identifier(system=US_NPI, value='123-45')
+    o.identifiers.append(i)
+    with SessionScope(db):
+        db.session.add(o)
+        db.session.commit()
+    o = db.session.merge(o)
+    return o
+
+
 def calc_date_params(backdate, setdate):
     """
     Returns the calculated date given user's request
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -323,7 +323,7 @@ def initialized_patient(app, add_user, initialized_db, shallow_org_tree):
 @pytest.fixture
 def add_user(app, initialized_db):
     def add_user(
-            username, first_name="", last_name="",
+            username, first_name="James", last_name="Nguyen",
             image_url=None, password="fakePa$$", email=None):
         """Create a user and add to test db, and return it"""
         # Hash the password
@@ -578,9 +578,6 @@ def login(initialize_static, app, client, music_org):
             oauth_info=None,
             follow_redirects=True
     ):
-
-        app.config.from_object(TestConfig)
-
         if not oauth_info:
             oauth_info = {'user_id': user_id}
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -19,7 +19,11 @@ from portal.models.codeable_concept import CodeableConcept
 from portal.models.coding import Coding
 from portal.models.encounter import Encounter
 from portal.models.intervention import add_static_interventions
-from portal.models.organization import Organization, OrgTree, add_static_organization
+from portal.models.organization import (
+        Organization, 
+        OrgTree, 
+        add_static_organization
+)
 from portal.models.procedure import Procedure
 from portal.models.qb_timeline import invalidate_users_QBT
 from portal.models.relationship import add_static_relationships
@@ -33,7 +37,7 @@ from portal.models.user_consent import (
     UserConsent,
 )
 from portal.system_uri import SNOMED
-from tests import TEST_USER_ID 
+from tests import TEST_USER_ID
 
 
 def pytest_addoption(parser):
@@ -382,7 +386,7 @@ def login(initialize_static, app, client, music_org):
         app.config.from_object(TestConfig)
 
         if not oauth_info:
-            oauth_info= {'user_id': user_id} 
+            oauth_info = {'user_id': user_id}
 
         return client.get(
                 'test/oauth',
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -106,16 +106,16 @@ def calc_date_params(backdate, setdate):
     return acceptance_date
 
 
-@pytest.fixture(scope="session")
+@pytest.fixture
 def initialize_static(initialized_db):
-    def initialize_static():
-        add_static_concepts(only_quick=True)
-        add_static_interventions()
-        add_static_organization()
-        add_static_relationships()
-        add_static_roles()
-        db.session.commit()
-    return initialize_static
+    add_static_concepts(only_quick=True)
+    add_static_interventions()
+    add_static_organization()
+    add_static_relationships()
+    add_static_roles()
+    db.session.commit()
+
+    yield
 
 
 @pytest.fixture(scope="session")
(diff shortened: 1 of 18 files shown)


tests/conftest.py
```diff
@@ -36,6 +36,15 @@ def pytest_addoption(parser):
     )
 
 
+def setUp():
```
pbugni: `setUp` seems like a pretty generic term.  you're specifically initializing the db static content here - i'd rename to something that's more appropriate.

I don't have a problem with it being a function that the fixtures call, but there needs to be a mechanism to avoid calling it twice, should a test require multiple fixtures, each of which call this function.  Does the fixture mechanism handle this?

tests/conftest.py
```diff
@@ -176,6 +177,19 @@ def add_service_user(sponsor=None):
     return add_service_user
 
 
+@pytest.fixture
+def add_music_org():
+    def add_music_org():
```
pbugni: @waynelai1228 I'm curious why you return a function from this fixture, rather than simply executing it.  Do the tests *need* to control when the music org is added?  Seems simpler to just have it run when a test names this fixture as a parameter, no?

pbugni: second question - is there anything about the "music" organization that is needed in the tests for which this fixture is used?  not trying to be overly pedantic, just wanting to avoid rolling too many business rules into the tests and code.  if a generic org will do, lets use that.

waynelai1228: Yes, you are right, I was just following the previous format and didn't notice I can run it directly.

waynelai1228: The music organization appeared in quite a significant amount of tests for login, I am still trying to find the best formula for this. The organization itself seems like the test-specific generic organization.

tests/conftest.py
```diff
@@ -16,6 +49,75 @@ def pytest_addoption(parser):
...
+                "`associative_backdate` and pass in `setdate` param")
+        acceptance_date = datetime.utcnow() - backdate
+    else:
+        acceptance_date = datetime.utcnow()
+    return acceptance_date
+
+
+@pytest.fixture(scope="session")
+def initialize_static(initialized_db):
+    def initialize_static():
```
pbugni: @waynelai1228 this and a number of the fixtures are still returning a closure, a function defined within the fixture, rather than just executing code as the fixture.

This means the test making use of the fixture must call the function rather than the typical usage pattern, where a named fixture has been executed and is ready by the time it enters the test function body.

Can you please remove the nested functions from the fixtures unless there is a need to do it this way?

waynelai1228: Yep, working on it.

tests/conftest.py
```diff
@@ -41,11 +143,24 @@ def app_logger(app):
 
 
 @pytest.fixture(scope='session')
-def initialized_db(app):
+def initialized_db(app, request):
     """Create database schema"""
+    db.drop_all()
     db.create_all()
```
pbugni: @waynelai1228 we shouldn't need a separate `teardown_db` fixture, especially as it's error prone to rely on all users of `initialized_db` to include the pair of fixtures.
If a fixture `yield`s , after the test is done, it'll run the code below the yield.  You should therefore be able to make the body of `teardown_db` part of `initialized_db`, just below a yield statement.

tests/conftest.py
```diff
@@ -16,6 +49,75 @@ def pytest_addoption(parser):
...
+@pytest.fixture
+def initialize_static(initialized_db):
+    add_static_concepts(only_quick=True)
+    add_static_interventions()
+    add_static_organization()
+    add_static_relationships()
+    add_static_roles()
+    db.session.commit()
+
+    yield
```
pbugni: The point of a `yield` in a fixture, is to mark a point of return, so after the test runs, context pops back to the fixture for clean up.  It's likely to be confusing to leave a yield at the end of a fixture without cleanup code following.  See https://pytest.readthedocs.io/en/2.9.1/yieldfixture.html for more info.

tests/conftest.py
```diff
@@ -71,3 +182,394 @@ def celery_app(app):
...
+        if email is not None:
+            test_user.email = email
+        with SessionScope(db):
+            db.session.add(test_user)
+            db.session.commit()
+        test_user = db.session.merge(test_user)
+        # Avoid testing cached/stale data
+        invalidate_users_QBT(test_user.id)
+        return test_user
+    yield add_user
```
pbugni: i believe you want `return add_user` not yield.

tests/conftest.py
```diff
@@ -71,3 +182,394 @@ def celery_app(app):
...
+
+@pytest.fixture
+def service_user(initialize_static, test_user):
+    sponsor = test_user
+    service_user = sponsor.add_service_account()
+    with SessionScope(db):
+        db.session.add(service_user)
+        db.session.commit()
+
+    yield db.session.merge(service_user)
```
pbugni: another s/yield/return/

tests/conftest.py
```diff
@@ -71,3 +182,394 @@ def celery_app(app):
...
+        with SessionScope(db):
+            db.session.add(tou)
+            db.session.add(privacy)
+            db.session.add(consent)
+            db.session.commit()
+
+        # Invalidate org tree cache, in case orgs are added by other
+        # tests.  W/o doing so, the new orgs aren't in the orgtree
+        OrgTree.invalidate_cache()
+    yield bless_with_basics
```
pbugni: s/yield/return/

tests/conftest.py
```diff
@@ -71,3 +182,394 @@ def celery_app(app):
...
+          happening at given time
+    
+        """
+        audit = Audit(user_id=TEST_USER_ID, subject_id=TEST_USER_ID)
+        for cc in CC.BIOPSY, CC.PCaDIAG, CC.PCaLocalized:
+            get_user(TEST_USER_ID).save_observation(
+                codeable_concept=cc, value_quantity=CC.TRUE_VALUE,
+                audit=audit, status='preliminary', issued=calc_date_params(
+                    backdate=backdate, setdate=setdate))
+    yield add_required_clinical_data
```
pbugni: s/yield/return/

tests/conftest.py
```diff
@@ -71,3 +182,394 @@ def celery_app(app):
...
+
+
+@pytest.fixture
+def system_user(add_user, promote_user):
+    """create and return system user expected for some tasks """
+    sysusername = '__system__'
+    if not User.query.filter_by(username=sysusername).first():
+        sys_user = add_user(sysusername, 'System', 'Admin')
+    promote_user(sys_user, ROLE.ADMIN.value)
+    yield sys_user
```
pbugni: s/yield/return/

tests/conftest.py
```diff
@@ -71,3 +182,394 @@ def celery_app(app):
...
+@pytest.fixture
+def music_org():
+    music_org = Organization(
+        name="Michigan Urological Surgery Improvement Collaborative"
+             " (MUSIC)")
+    with SessionScope(db):
+        db.session.add(music_org)
+        db.session.commit()
+    music_org = db.session.merge(music_org)
+    yield music_org
```
pbugni: s/yield/return

tests/conftest.py
```diff
@@ -71,3 +182,394 @@ def celery_app(app):
...
+            db.session.commit()
+            enc = db.session.merge(enc)
+            procedure.code = code
+            procedure.user = db.session.merge(test_user)
+            procedure.start_time = setdate or datetime.utcnow()
+            procedure.end_time = datetime.utcnow()
+            procedure.encounter = enc
+            db.session.add(procedure)
+            db.session.commit()
+    yield add_procedure
```
pbugni: s/yield/return/

tests/conftest.py
```diff
@@ -71,3 +182,394 @@ def celery_app(app):
...
+          happening at given time
+
+        """
+        audit = Audit(user_id=TEST_USER_ID, subject_id=TEST_USER_ID)
+        for cc in CC.BIOPSY, CC.PCaDIAG, CC.PCaLocalized:
+            get_user(TEST_USER_ID).save_observation(
+                codeable_concept=cc, value_quantity=CC.TRUE_VALUE,
+                audit=audit, status='preliminary', issued=calc_date_params(
+                    backdate=backdate, setdate=setdate))
+    yield add_required_clinical_data
```
pbugni: s/yield/return/

tests/test_auth.py
```diff
@@ -19,452 +19,451 @@
...
+
+
+def test_local_login_valid_username_and_password(test_auth_user, local_login):
+    test_auth_user = db.session.merge(test_auth_user)
+    """login through the login form"""
+    # Attempt to login with valid creds
+    response = local_login(test_auth_user.email, 'Password1')
+
+    # Validate login was successful
+    assert response.status_code is 200
```
pbugni: I typically only use the keyword `is` for comparison w/ `None`.  I think `==` makes more sense here.

tests/conftest.py
```diff
@@ -16,6 +49,75 @@ def pytest_addoption(parser):
...
+@pytest.fixture
+def initialize_static(initialized_db):
+    add_static_concepts(only_quick=True)
+    add_static_interventions()
+    add_static_organization()
+    add_static_relationships()
+    add_static_roles()
+    db.session.commit()
+
+    return
```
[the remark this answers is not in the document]
waynelai1228: Hm... this must be an unsuccessful attempt to make initialize_static effective that was overlooked when editing :D

tests/test_portal.py
```diff
@@ -16,311 +17,399 @@
...
+
+
+def test_report_error(test_user_login, client):
+    params = {
+        'subject_id': 112,
+        'page_url': '/not/real',
+        'message': 'creative test string'
+    }
+    response = client.get('/report-error?{}'.format(
+        urllib.parse.urlencode(params)))
```
ivan-c, line 229: I think using `query_string` will automatically encode the params

tests/organization-example-gastro.json
```diff
@@ -1,6 +1,6 @@
 {
   "resourceType": "Organization",
-  "id": "1",
+  "id": "2",
```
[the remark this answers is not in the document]
waynelai1228: The id of "1" cause conflict errors (two organisations with id of "1") in particular tests (after the change in the commit), changing it to "2" avoids those errors.
I asked for Paul's (@pbugni) opinion and he thought it is not a bad idea to change the id to something else that would be unlikely to invoke conflicts.

pbugni:
Replaced by #3721 - closing this out.

PR #3600, opened 2019-12-19
More edge cases handled in Research Protocol advancement

pbugni:
Now handling two edge cases previously overlooked:
1. if the RP being replaced doesn't include all the "visits" found in the next RP, extend logic to look ahead and step back when appropriate.
2. if the RP being replaced runs dry before the next RP starts, force a transition and pick up the new.

Also added another Test organization with all 3 IRONMAN RPs defined, with old retirement dates so we can test additional scenarios.

### DIFF
--- a/portal/models/qb_timeline.py
+++ b/portal/models/qb_timeline.py
@@ -302,8 +302,9 @@ def cur_next_rp_gen(user, classification, trigger_date):
             next_rp, next_retired = sorted_rps[-1]
         else:
             next_rp = None
-        trace("current RP {}, next RP {}".format(
-            current_rp.id, next_rp.id if next_rp else 'None'))
+        trace("current RP '{}'({}), next RP '{}'({})".format(
+            current_rp.name, current_rp.id, next_rp.name
+            if next_rp else 'None', next_rp.id if next_rp else 'None'))
 
         curRPD = RPD(
             rp=current_rp,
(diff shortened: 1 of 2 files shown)


portal/models/qb_timeline.py
```diff
@@ -315,11 +316,131 @@ def qbds_for_rp(rp, classification, trigger_date):
...
+        if self.cur_qbd:
+            self.cur_start = calc_and_adjust_start(
+                user=self.user, qbd=self.cur_qbd, initial_trigger=self.td)
+            self.cur_exp = calc_and_adjust_expired(
+                user=self.user, qbd=self.cur_qbd, initial_trigger=self.td)
+
+        self.nxt_qbd, self.nxt_start = None, None
+        if self.nxt_rpd:
+            self.nxt_qbd = next(self.nxt_rpd.qbds, None)
+        if self.nxt_qbd:
```
ivan-c: might be good to invert into a guard clause, to remove some of the below nesting

portal/models/qb_timeline.py
```diff
@@ -359,20 +469,22 @@ def nextQBD(qbds_gen):
...
+    else:
+        trace("initial trigger date {}".format(td))
 
-    rp_walker = cur_next_rp_gen(user, classification, td)
-    curRPD, nextRPD = next(rp_walker, (None, None))
+    rp_flyweight = RP_flyweight(
+        user=user, trigger_date=td, classification=classification)
 
-    if curRPD:
+    if rp_flyweight.cur_rpd:
```
ivan-c: might be cleaner to exit early if no RP found...

pbugni: that would miss the else, where we pull in intervention (i.e. not RP) tied QBs.

portal/models/qb_timeline.py
pbugni: yes, looks like i'm too late... already merged :)

PR #3601, opened 2019-12-19
guard against missing keys in migration

pbugni:
this broke on my system as well.  does this guard alter your expected migration?

### DIFF
--- a/portal/migrations/versions/4737a563e3f9_.py
+++ b/portal/migrations/versions/4737a563e3f9_.py
@@ -34,7 +34,7 @@ def fixup_comorb(questionnaire_response_json, comorb_options):
     target_question = qnr_json_copy['group']['question'][0]
     coded_answers = []
     for answer in target_question['answer']:
-        if 'valueString' not in answer:
+        if 'valueCoding' in answer:
             continue
         coded_answer = {'valueCoding': {
             # lookup code from valueString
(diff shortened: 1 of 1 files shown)


portal/migrations/versions/4737a563e3f9_.py
```diff
@@ -34,6 +34,8 @@ def fixup_comorb(questionnaire_response_json, comorb_options):
     target_question = qnr_json_copy['group']['question'][0]
     coded_answers = []
     for answer in target_question['answer']:
+        if 'valueString' not in answer:
+            continue
```
ivan-c: since the "fixed" answers will be pairs of valueCoding and valueString (maintained for legacy reasons), this will "double encode" answers that have already been fixed in migration.

it would be safest to return the original JSON if any valueCoding is found (none were used in the comorb instrument before it was fixed), implying the target QNR is already fixed.

pbugni: would you mind patching this up?

portal/migrations/versions/4737a563e3f9_.py
```diff
@@ -34,6 +34,8 @@ def fixup_comorb(questionnaire_response_json, comorb_options):
     target_question = qnr_json_copy['group']['question'][0]
     coded_answers = []
     for answer in target_question['answer']:
+        if 'valueString' not in answer:
```
ivan-c:
```suggestion
        if 'valueCoding' in answer:
```

portal/migrations/versions/4737a563e3f9_.py
ivan-c:
```suggestion
            return questionnaire_response_json
```

PR #3605, opened 2019-12-21
Deploy Script Activity Check

ivan-c:
* Add inactivity check to deploy script by default
  * Override with new force (`-f`) flag
* Allow configuration of `LOG_LEVEL` by environment variable

### DIFF
--- a/bin/deploy-docker.sh
+++ b/bin/deploy-docker.sh
@@ -65,7 +65,13 @@ fi
 if [ "$FORCE" != true ]; then
     # how long a system must be inactive before deploying
     min_inactivity_seconds=$((15 * 60))
+
     inactive_seconds="$(docker-compose exec --env LOG_LEVEL=info web flask last-usage | tr -d '[[:space:]]')"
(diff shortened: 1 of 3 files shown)


bin/deploy-docker.sh
```diff
@@ -59,6 +62,24 @@ if [ -z "$NO_PULL" ]; then
     docker-compose pull
 fi
 
+if [ "$FORCE" != true ]; then
+    # how long a system must be inactive before deploying
+    min_inactivity_seconds=$((15 * 60))
+
+    inactive_seconds="$(docker-compose exec --env LOG_LEVEL=info web flask last-usage | tr -d '[[:space:]]')"
+    if ! echo "$inactive_seconds" | grep --quiet '^-*[[:digit:]]\+$'; then
+        echo "error reading inactivity time"
```
pbugni: Would it be helpful to also echo the `inactive_seconds` string in event of error?  Thinking of a case where `docker-compose` is unhappy or whatnot, may ease debugging.

PR #3617, opened 2020-01-06
TN-2450 

pbugni:
only proceed with filtering answers if the given answers follow the expected format, pairs of (`valueCoding`, `valueString`) keys.

### DIFF
--- a/portal/models/questionnaire_response.py
+++ b/portal/models/questionnaire_response.py
@@ -661,11 +661,9 @@ def consolidate_answer_pairs(answers):
 
     answer_types = [list(a.keys())[0] for a in answers]
 
-    # Exit early if assumptions not met
-    if (
-        len(answers) % 2 or
-        answer_types.count('valueCoding') != answer_types.count('valueString')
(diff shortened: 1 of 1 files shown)


portal/models/questionnaire_response.py
```diff
@@ -661,11 +661,9 @@ def consolidate_answer_pairs(answers):
...
     answer_types = [list(a.keys())[0] for a in answers]
 
-    # Exit early if assumptions not met
-    if (
-        len(answers) % 2 or
-        answer_types.count('valueCoding') != answer_types.count('valueString')
-    ):
+    # Exit early if assumptions not met, i.e. ordered pairs of
+    # 'valueCoding' followed by 'valueString' keys
+    if answer_types != ['valueCoding', 'valueString'] * int(len(answers)/2):
```
ivan-c: Ah, clever fix!

ivan-c: As discussed in slack, this order needs to be reversed

PR #3619, opened 2020-01-07
Feature/replace flask dogpile lib

pbugni:


portal/models/qb_timeline.py
```diff
@@ -816,15 +826,18 @@ def update(self, value):
         return stringform
 
 
-@dogpile_cache.region('assessment_cache_region')
+@cache.memoize(timeout=60*60*2)
```
ivan-c: Can you add a human-readable version as a comment eg  `2 hours in seconds`

PR #3621, opened 2020-01-10
leverage caching

pbugni:
Profiling /patients, found some low hanging fruit to cache.  Took relatively low risk options only to avoid need to invalidate on data changes (i.e. organizations, questionnaire banks, research protocols)

### DIFF
--- a/portal/cache.py
+++ b/portal/cache.py
@@ -1,9 +1,22 @@
+from flask import request, url_for
 from flask_caching import Cache
-
 
 # Simple init, replaced during app configuration
 cache = Cache(config={'CACHE_TYPE': 'simple'})
 
-
 # Human readable constants, values in seconds, for cache timeouts
 TWO_HOURS = 2*60*60
+FIVE_MINS = 5*60
+
+
+def request_args_in_key():
+    """Set as `key_prefix=request_args_in_key` for safe view caching
+
+    By default, cache.cached() uses only the request path as the key.
+    For any views returning different results with query string parameters,
+    need to include those parameters in the cache key
+
+    :return: key for caching
+
+    """
+    return url_for(request.endpoint, **request.args)
--- a/portal/cache.py
+++ b/portal/cache.py
@@ -1,8 +1,8 @@
 from flask import request, url_for
 from flask_caching import Cache
 
-# Simple init, replaced during app configuration
-cache = Cache(config={'CACHE_TYPE': 'simple'})
+# Configured during app configuration
+cache = Cache()
 
 # Human readable constants, values in seconds, for cache timeouts
 TWO_HOURS = 2*60*60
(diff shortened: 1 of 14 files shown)


portal/cache.py
```diff
@@ -1,9 +1,22 @@
...
 TWO_HOURS = 2*60*60
+FIVE_MINS = 5*60
+
+
+def request_args_in_key():
+    """Set as `key_prefix=request_args_in_key` for safe view caching
+
+    By default, cache.cached() uses only the request path as the key.
+    For any views returning different results with query string parameters,
+    need to include those parameters in the cache key
```
ivan-c: @achen2401 FYI we may need to remove `cache: false` from some jQuery AJAX calls for this to take effect.

As I understand it, we should generally be able to rely on the browser doing the "right thing" with caching.

Our static content has an 8 hour `Expires` header, but otherwise we do not add any headers to instruct the browser to cache requests, so it should make a fresh request every time.

---
See [`jQuery.ajax()`](https://api.jquery.com/jquery.ajax/)
```
It works by appending "_={timestamp}" to the GET parameters
```

portal/factories/app.py
```diff
@@ -233,6 +233,11 @@ def configure_blueprints(app, blueprints):
 
 def configure_logging(app):  # pragma: no cover
     """Configure logging."""
+    # Avoid duplicate logging.  Multiple handlers implies logging
```
ivan-c: Nice!! :star2:

portal/views/organization.py
```diff
@@ -30,6 +31,7 @@
 @org_api.route('/organization')
 @crossdomain()
 @oauth.require_oauth()
```
ivan-c: Might be out of scope for this PR, but do we actually need to require auth for this view?

portal/views/portal.py
```diff
@@ -136,6 +137,15 @@ def report_slow_queries(response):
         SQLALCHEMY_RECORD_QUERIES = True
 
     """
+    if current_app.config.get("LOG_SLOW_RESPONSES"):
```
ivan-c: Would it make sense to move this to a decorator or separate function? Small amount of code, but might be nice to separate benchmarking code from mission-critical functionality

portal/factories/app.py
pbugni: i KNOW!  about !@$ing time we found that one!

portal/views/organization.py
pbugni: agreed that it's out of scope.  doesn't seem like there's a legit reason to require it.

portal/views/portal.py
pbugni: that whole function is just for that, debugging slowness...

portal/cache.py
pbugni: @achen2401 an especially good one to test is any calls you make to /api/organization - there should be a notable difference in timing when getting the cached version.

PR #3623, opened 2020-01-16
TN-2462-MoveRoyalBrisbaneToOtherParent

mcjustin:
https://jira.movember.com/browse/TN-2462

May also need to delete org 21700 - TBD.  <= WE DID DECIDE TO DO THIS 2019-01-17


mcjustin:
Theresa asked that we move forward w/ deleting the old parent - done now, so this is ready for final review @ivan-c

mcjustin:
Noting that there are no prod users associated w/ the deleted org 21700. That may not be the case on other systems (demo/test etc), so we may get alerts

PR #3624, opened 2020-01-16
TN-1538 add DOB to /patients.  remove phone numbers.

pbugni:


portal/templates/admin/patients_by_org.html
```diff
@@ -57,9 +57,8 @@ <h4 class="tnth-headline">{% trans %}Patient List{% endtrans %}</h4>
                   <th data-field="username" data-sortable="true" data-visible="false">{{ _("Username") }}</th>
                   <th data-field="firstname" data-sortable="true" data-class="firstname-field" data-filter-control="input">{{ _("First Name") }}</th>
                   <th data-field="lastname" data-sortable="true" data-class="lastname-field" data-filter-control="input">{{ _("Last Name") }}</th>
+                  <th data-field="birthdate" data-sortable="true" data-class="birthdate-field" data-filter-control="input">{{ _("Date of Birth") }}</th>
                   <th data-field="email" data-sortable="true" data-class="email-field" data-filter-control="input">{{ _("Email") }}</th>
-                  <th data-field="phone" data-sortable="true" data-visible="false" data-width="10%" data-class="phone-field" data-filter-control="input">{{ _("Cell") }}</th>
```
mcjustin: @pbugni  I'd noted in the story: the table needs to remain unchanged for IRONMAN users - that's the pain in this request, from the dev POV.
https://jira.movember.com/browse/TN-1538?focusedCommentId=243611&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-243611

pbugni: yes, I saw that, but I took the opportunity to remove the phone numbers, as they're so rarely used and seem out of place in this, not to mention the performance gain by dropping them.

can we run with this, see if it's actually missed by anyone?

the IRONMAN users will have their table preferences in place, so most (that don't already have phone numbers included) won't see any change.

achen2401: thank you, paul, I am all for performance gain, but defer to you guys on whether to keep the phone number field.

pbugni:
Spoke w/ @mcjustin on slack - he's okay with this.  Will update story with details.

PR #3625, opened 2020-01-17
TN-2464 multi RP transitions pre start

pbugni:
in case of multiple RPs that were retired before a user's initial trigger, need to advance RPs multiple times in the transition process.

### DIFF
--- a/tests/test_qb_timeline.py
+++ b/tests/test_qb_timeline.py
@@ -383,9 +383,13 @@ class TestQbTimeline(TestQuestionnaireBank):
             now=now, backdate=relativedelta(months=7))
         back14, nowish = associative_backdate(
             now=now, backdate=relativedelta(months=14))
+        ahead36, nowish = associative_backdate(
+            now=now, backdate=relativedelta(months=-36))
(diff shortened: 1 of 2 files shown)


tests/test_qb_timeline.py
```diff
@@ -403,6 +407,36 @@ def test_change_before_start_rp(self):
...
+        self.consent_with_org(org_id=org_id, setdate=back1)
+        user = db.session.merge(self.test_user)
+        gen = ordered_qbs(user)
+
+        # expect everything in v5
+        expect_baseline = next(gen)
+        assert visit_name(expect_baseline) == 'Baseline'
+        assert (
+            expect_baseline.questionnaire_bank.research_protocol.name == 'v5')
+        for n in (3, 6, 9, 15, 18, 21, 30):
```
mcjustin: Noting that v5 is every 3 months, and goes out to 60 months, so tests probably should, also

pbugni: well, yes.  but these tests are using bootstrap functions to capture only an isolated detail under test (and often don't match current config - such a change would break and updating that config would be messy).  there are others testing the persistence load and recur functionality, also in isolation.

PR #3635, opened 2020-01-30
TN-2447 reduce fields requested from external IdPs

pbugni:
FB has added additional permission barriers for attributes like user_birthdate and user_gender.

As suggested in the comments of TN-2447, reduce the request to external IdPs - no longer request either birthdate or gender.


pbugni:
This is now in a working state for FB and google, both new accounts and existing access on truenth-test.

PR #3639, opened 2020-02-04
TN-2379 add semaphore locks to expensive user driven reporting tasks.

pbugni:
Two background, user driven reporting tasks are now constrained to a single running occurrence with a semaphore lock.

- on ePROMs from the /patients screen, the far right "Export Report Data" button
- from menu (with researcher role), Research Data -> Export Questionnaire Data


portal/static/js/src/admin.js
```diff
@@ -175,7 +175,7 @@ import CurrentUser from "./mixins/CurrentUser.js";
                 $(".exportReport__status").removeClass("active");
                 if (options.error) {
                     this.updateProgressDisplay("", "");
-                    $(".exportReport__error .message").html("Request to export report data failed.");
+                    $(".exportReport__error .message").html(`Request to export report data failed.${options.message?"<br/>"+options.message: ""}`);
```
ivan-c: Just learned about these format strings! :D

PR #3640, opened 2020-02-05
Force str(Message) in logging to eliminate bogus error log text.

pbugni:
There's a comment in the code:
```            # This should never happen, alert if it does ```

generating emails with the following useless content:

```Unable to generate audit log for email <class 'flask_mail.Message'>```

Although `format()` is documented to automatically call an object's __str__ method, it doesn't look to be functioning from the log statement.  Force the call to improve our visibility on this situation.

### DIFF
--- a/portal/models/message.py
+++ b/portal/models/message.py
@@ -180,8 +180,7 @@ class EmailMessage(db.Model):
         else:
             # This should never happen, alert if it does
             current_app.logger.error(
-                "Unable to generate audit log for email {}".format(
-                    str(Message)))
+                "Unable to generate audit log for email: %s", str(Message))
         # If an exception was raised when attempting the send, re-raise now
         # for clients to manage
         if exc:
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: portal/models/message.py [hunk 1]
```diff
@@ -180,7 +180,8 @@ def send_message(self, cc_address=None):
         else:
             # This should never happen, alert if it does
             current_app.logger.error(
-                "Unable to generate audit log for email {}".format(Message))
+                "Unable to generate audit log for email {}".format(
+                    str(Message)))
```
ivan-c: I think loggers use printf-type formatting by default. maybe the reason it's not casting `Message`?
eg:
```
current_app.logger.error("Unable to generate audit log for email %s", Message)

```

portal/models/message.py
```diff
@@ -180,7 +180,7 @@ def send_message(self, cc_address=None):
         else:
             # This should never happen, alert if it does
             current_app.logger.error(
-                "Unable to generate audit log for email {}".format(Message))
+                "Unable to generate audit log for email: %s", str(Message))
```
ivan-c: just curious, did you find that explicit cast to `str` necessary? I too thought it was automatically cast when you use string formatting

PR #3647, opened 2020-02-13
Test/truenth usa logo update

achen2401:
reference: https://jira.movember.com/browse/TN-2182
Logo updates for **TrueNTH USA only**
Changes include:

- using config to swamp in logo image source for portal wrapper banner
- logo changes in both unauthenticated and authenticated pages in TrueNTH USA
- styling fixes related to logo updates


portal/config/config.py
```diff
@@ -229,6 +229,11 @@ class BaseConfig(object):
         'uwcirg.github.io',
     ]
 
+    BANNER_LOGOS = [
```
ivan-c: Will there ever be more than 2 images?

Would it make sense to name the images based on the `PERSISTENCE_DIR` config name (ie `gil` or `eproms`)? (so we wouldn't need an extra config)

PR #3665, opened 2020-03-31
Allow for slightly mismatched RP schedules

pbugni:
As RP v5 shifted start dates back a couple weeks, need to accommodate in the look ahead logic when determining if it's time to transition to the next RP.

### DIFF
--- a/portal/models/qb_timeline.py
+++ b/portal/models/qb_timeline.py
@@ -384,7 +384,9 @@ class RP_flyweight(object):
                 trace("Finished cur RP with remaining QBs in next")
                 self.cur_start = self.nxt_start
                 self.transition()
-            elif self.cur_start != self.nxt_start:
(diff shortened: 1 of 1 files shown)


portal/models/qb_timeline.py
```diff
@@ -384,7 +384,9 @@ def next_qbd(self):
                 trace("Finished cur RP with remaining QBs in next")
                 self.cur_start = self.nxt_start
                 self.transition()
-            elif self.cur_start != self.nxt_start:
+            elif self.cur_start > self.nxt_start + relativedelta(months=1):
```
ivan-c: A bit worried about patients falling between RPs, but I'm sure you've thought of it :)

pbugni: I'm more that a bit worried, this overly complex set of requirements is difficult to describe in code.  We're shifting designs soon.  Hopefully it will be easier to maintain.

PR #3693, opened 2020-05-27
TN-2627 encounter api & allow pre-registered users in via url authentication

pbugni:
NB - to enable, set config value `ENABLE_URL_AUTHENTICATED` to a non false value.

- TN-2636: Add endpoint for front-end to acquire logged in user's current encounter.  The `auth_method` will define features that should be hidden from the user, such as making profile changes when `auth_method == 'url_authenticated'`
 
- TN-2635: Entry via /access/<token> now allows EPROMs users direct access, providing they've registered and finished initial queries.  The encounter will have the aforementioned `auth_method == 'url_authenticated'`

- TN-2647: Add css-class "portal-weak-auth-disabled" to the completed assessment tile, so the front end can disable or redirect to login, when the user's current encounter has `auth_method == 'url_authenticated'`

- TN-2648: Consider `url_authenticated` when evaluating authorization for all API endpoints

### DIFF
--- a/portal/models/user.py
+++ b/portal/models/user.py
@@ -1634,7 +1634,9 @@ class User(db.Model, UserMixin):
             if ot.at_or_below_ids(org.id, [org_id]):
                 return True
 
-    def check_role(self, permission, other_id):
+    def check_role(
+            self, permission, other_id,
+            allow_on_url_authenticated_encounters=False):
         """check user for adequate role
 
         if user is an admin or a service account, grant carte blanche
@@ -1644,8 +1646,18 @@ class User(db.Model, UserMixin):
         returns true if permission should be granted, raises 404 if the
         other_id can't be found, otherwise raise a 401
 
+        NB - a user with "url_authenticated" as their current encounter's
+        auth_type will NOT have any access, unless specifically requested
+        via the "allow_on_url_authenticated_encounters" parameter
+
         """
         assert (permission in ('view', 'edit'))  # limit vocab for now
+        if (
+                not allow_on_url_authenticated_encounters and
+                self.current_encounter.auth_method == 'url_authenticated'):
+            abort(401, "inadequate auth_method: {}".format(
+                self.current_encounter.auth_method))
+
         if self.id == other_id:
             return True
         try:
@@ -1895,16 +1907,11 @@ def current_user():
     return None
 
 
-def get_user(uid):
-    if uid:
-        return User.query.get(uid)
+def unchecked_get_user(uid, allow_deleted=False):
+    """direct access to user by id - does NOT include authorization check
 
-
-def get_user_or_abort(uid, allow_deleted=False):
-    """Wraps `get_user` and raises error if not found
-
-    Safe to call with path or parameter info.  Confirms integer value before
-    attempting lookup.
+    Clients should typically use `get_user()` unless there's need to
+    access without authorization check, say prior to login.
 
     :param uid: integer value for user id to look up
     :param allow_deleted: set true to allow access to deleted users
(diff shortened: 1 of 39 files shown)


portal/views/user.py
```diff
@@ -957,6 +957,89 @@ def delete_user_consents(user_id):
     return jsonify(message="ok")
 
 
+@user_api.route('/user/<int:user_id>/encounter', methods=('GET',))
```
achen2401, line 960: @pbugni Paul, is the `user_id` required if only current user is expected?  I think that the frontend will need to get the user id (e.g. maybe call to /api/me) before it calls this api endpoint, just thinking that it probably doesn't need to make that extra call to get the user id.

portal/templates/portal_wrapper.html
```diff
@@ -94,7 +94,7 @@
                          <button id="tnthUserBtn" type="button" class="tnth-nav-btn tnth-white tnth-dropdown-toggle" data-toggle="tnth-dropdown">
                             {% if enable_links %}&nbsp;&nbsp; {{ _("MENU") }} <i class="fa fa-bars"></i>{% endif %}
                             <img class="profile-img" src="{{movember_profile}}" width="50" height="50" alt="{{_('Profile image')}}"/>
-                            <span class="welcome-text"><em>{{ _("Welcome") }}</em>{% if user.display_name != "Anonymous" %}, {{ user.display_name }}{% endif %}</span>
+                            <span class="welcome-text portal-weak-auth-hide"><em>{{ _("Welcome") }}</em>{% if user.display_name != "Anonymous" %}, {{ user.display_name }}{% endif %}</span>
```
ivan-c, line 97: Would a check for `user.has_role('write_only')` be more appropriate than checking a symptom of that role?

portal/views/assessment_engine.py
```diff
@@ -1623,6 +1618,11 @@ def present_needed():
         return redirect('/')
 
     url = url_for('.present_assessment', **args)
+
+    encounter = current_user().current_encounter
+    if encounter.auth_method == "url_authenticated":
+        return redirect('/confirm-identity?redirect_url={}'.format(url))
```
ivan-c, line 1624: I've found redirects really hard to debug, would it make sense to add a logging line here before redirecting?

portal/models/user.py
```diff
@@ -1643,8 +1647,19 @@ def check_role(self, permission, other_id):
...
 
+        NB - a user with "url_authenticated" as their current encounter's
+        auth_type will NOT have any access, unless specifically requested
+        via the "allow_on_url_authenticated_encounters" parameter
+
         """
         assert (permission in ('view', 'edit'))  # limit vocab for now
+        if (
+                not allow_on_url_authenticated_encounters and
+                current_app.config.get('ENABLE_URL_AUTHENTICATED') and
```
ivan-c, line 1658: Is there a `os.environ.get('ENABLE_URL_AUTHENTICATED')` call in [config.py](https://github.com/uwcirg/truenth-portal/blob/develop/portal/config/config.py) for this? (if we want to config via docker)

portal/views/assessment_engine.py
achen2401: thanks, ivan, added

portal/templates/portal_wrapper.html
achen2401: think the issue here that we don't check the user's role, but checking the authentication method, i.e. url_authenticated, in the user's encounter instead

portal/models/user.py
pbugni: i expect this to become part of eproms' site_persistence config, but given they might want to toggle before a release cycle - yes, we'll need that!

portal/templates/portal_wrapper.html
pbugni: @achen2401 is correct.  roles are generally static, as opposed to an encounter level auth method.

portal/models/user.py
ivan-c: Added in 90156217c91cb1b151bc6834f99ff6f2b22256fc

portal/templates/portal_wrapper.html
ivan-c: Ah, will the check in the view also work in the template?:
`user.current_encounter.auth_method == 'url_authenticated'`

achen2401: thanks, ivan, I think it works, although, in this case we want to hide the relevant display `user.current_encounter.auth_method != 'url_authenticated'`

PR #3695, opened 2020-05-28
URL-authenticated - disabling relevant frontend elements

achen2401:
address : https://jira.movember.com/browse/TN-2645

Changes include:

- frontend code change to check authentication method via which user is authenticated
- disabling UI elements, i.e. profile menu links and links with css class, 'portal-weak-auth-disabled', if user encounter indicates a "url_authenticated" auth method

**NOTE:** This is dependent on changes in this PR: #3693


pbugni:
@achen2401 looks good.  i'll recommend you hold-up this PR till we add the css-class and front end needs as that's part of the same issue?

PR #3696, opened 2020-06-02
Feature/url authenticated user confirmation

achen2401:
Address: https://jira.movember.com/browse/TN-2628
prompting **URL-authenticated** user to verify identity before continuing to take survey

Changes include:

- redirect URL-authenticated user to page presenting dialog asking user to confirm identity
- clicking Yes on the dialog will record an audit event and redirect user to survey
- clicking No will redirect user back to home page


portal/static/js/flask_user/ConfirmIdentity.js
```diff
@@ -0,0 +1,27 @@
...
+    $("#btnConfirmIdentity").on("click", function(e) {
+        e.stopPropagation();
+        /*
+         * record an audit event before redirecting user to survey
+         */
+        $.ajax({
+            type: "POST",
+            url: "/api/auditlog",
+            data: {
+                "message": "User confirmed identity as survey taker"
```
pbugni, line 17: I'll note, the user & subject of the audit will be pulled from the current user.

PR #3721, opened 2020-07-24
Fixturize Unittests

ivan-c:
* Squash commits of #3592
* Include both old unittests and new fixtures until merge conflicts worked around

### Todo:
- [x] Fix conflicts between new fixtures and old unittests
- [x] Remove old unittests when suitable fixture exists
- [x] Remove fixtures that return functions, only as necessary (for tests to pass)


pbugni:
@ivan-c We're at a passing state - there's clearly room for improvement, but I'd love to accept what we have and move on.
Tests passing, doing one more white space only commit addressing the remaining pep8 warnings.

PR #3725, opened 2020-07-28
Reuse valid encounters during login flow, if found

pbugni:
Due to oauth initiation happening via API invocation, many of the fail safe calls to `login()` needlessly generate a new encounter, and previously occasionally included the incorrect assumption on `auth_method`.

Now reusing the current_encounter for a user on a call to login, if it's found to be valid and of the requested auth_method.

Should also resolve a bug in the `url_authenticated` flow, where any oauth protected API call from an external intervention (i.e. the AssessmentEngine) incorrectly promoted the auth_method.

### DIFF
--- a/portal/models/encounter.py
+++ b/portal/models/encounter.py
@@ -172,6 +172,10 @@ def initiate_encounter(user, auth_method):
     if user.has_role(ROLE.SERVICE.value):
         auth_method = 'service_token_authenticated'
 
+    # If the auth_method is unknown, fall back to failsafe
+    if auth_method is None:
+        auth_method = 'failsafe'
+
     # Initiate new as directed
     encounter = Encounter(
         status='in-progress', auth_method=auth_method,
--- a/portal/models/login.py
+++ b/portal/models/login.py
@@ -28,14 +28,13 @@ def login_user(user, auth_method=None):
         generate_failsafe_if_missing=False)
     if not active_encounter:
         current_app.logger.debug(
-            "No current encounter found for user {}, initiate as {}".format(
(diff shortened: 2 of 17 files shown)


portal/models/encounter.py
```diff
@@ -172,6 +172,10 @@ def initiate_encounter(user, auth_method):
     if user.has_role(ROLE.SERVICE.value):
         auth_method = 'service_token_authenticated'
 
+    # If the auth_method is unknown, fall back to failsafe
+    if auth_method is None:
+        auth_method = 'failsafe'
```
ivan-c, line 177: Should this be something like "unknown"?

portal/models/login.py
```diff
@@ -10,16 +10,33 @@
...
 
     """
     if not _call_or_get(flask_login_current_user.is_authenticated):
         flask_user_login(user)
-    initiate_encounter(user, auth_method)
+
+    # Reuse active encounter if available, and a new auth_method wasn't provided
+    active_encounter = user.current_encounter(generate_failsafe_if_missing=False)
+    if not active_encounter:
+        current_app.logger.debug(
```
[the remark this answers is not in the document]
pbugni: thanks for the reminder.  will correct.

portal/models/encounter.py
pbugni: it's part of an enumeration - and in the db - so that would require a migration and other work.  i don't feel it's that semantically different - not worth the effort.  prefer to stick w/ fail safe, which we use when the situation is "unknown".

PR #3739, opened 2020-08-05
TN-2653 Extend UserConsent and related APIs to support ResearchStudy

pbugni:


portal/models/research_study.py
```diff
@@ -42,3 +42,21 @@ def update_from_fhir(self, data):
...
+
+def add_static_research_studies():
+    """Seed database with default static research studies
+
+    Idempotent - run anytime to pick up any new relationships in existing dbs
+
+    """
+    base = {
+      "id": 0,
+      "title": "Base Study",
```
ivan-c, line 55: From an i18n angle, will these titles be viewable by staff or patients?

ie do we need to add them to the [existing static string extraction code](https://github.com/uwcirg/truenth-portal/blob/develop/portal/models/i18n.py#L23-L37)

pbugni: ugh - yes, they may, as part of TN-2634.

PR #3747, opened 2020-08-14
Feature/substudy consent ui update

achen2401:
address: https://jira.movember.com/browse/TN-2634
add a new entry for substudy consent in the Consent History table in the profile page.


portal/static/js/src/data/common/consts.js
```diff
@@ -0,0 +1,2 @@
+export var EPROMS_SUBSTUDY_ID = 1;
+export var EPROMS_SUBSTUDY_TITLE = i18next.t("IRONMAN sub study");
```
ivan-c, line 2: >sub study

Is this the preferred format? I've seen `sub-study` and `substudy`, just wanted to double check...

achen2401: thank you, Ivan. I will confirm with Cam.

achen2401: noting here that placeholder text will change per feedback.  A new story will be created to address this.

PR #3754, opened 2020-08-27
TN-2740 research study in research protocol

pbugni:
To wire up organizations for sub-studies, research protocols are now associated with a research study.


tests/conftest.py
```diff
@@ -149,12 +149,19 @@ def calc_date_params(backdate, setdate):
     return acceptance_date
 
 
+@pytest.fixture
+def initialized_with_research_study(initialized_db):
```
ivan-c, line 153: Would it make sense for this fixture to return a reference (SQLA queryset?) to the data that's been added, so tests that depend on the fixture can introspect the fixture?

tests/test_research_protocol.py
```diff
@@ -1,78 +1,78 @@
...
+        db.session.add(rp)
+        db.session.commit()
+    rp = db.session.merge(rp)
+
+    assert rp.id
+    assert rp.created_at
+
+
+def test_rp_as_json(initialized_with_research_study):
+    rp = ResearchProtocol(name="test_rp", research_study_id=0)
```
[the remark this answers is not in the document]
pbugni: not that one, as the point of that test is the JSON instantiation - but you're point is solid.  made another fixture and improved the rest of the tests in that module.

tests/conftest.py
pbugni: complicated :)  that one is referencing a known static value - but we do return live objects where they're useful to the tests.

PR #3769, opened 2020-09-15
TN-2670-ironman_ss-qnr-edit-2

mcjustin:
https://jira.movember.com/browse/TN-2670 - removes two questions near the end; note that this results in the final question's index being decremented by 2.


ivan-c (approved):
Looks good!

ivan-c:
@pbugni will this in-place edit of a Questionnaire be noticed by site-persistence, or would we need to clear the old one (or use a migration) by hand?

mcjustin:
@pbugni  says that it should be fine w/out a migration... will merge and try on test

PR #3770, opened 2020-09-16
TN-2747 completed or in-process indefinite overlooked on RP change

pbugni:
The indefinite case it special - we need to consider any work done on previous research protocols which may come from a different questionnaire bank.

### DIFF
--- a/portal/models/questionnaire_response.py
+++ b/portal/models/questionnaire_response.py
@@ -506,14 +506,26 @@ class QNR_results(object):
 
 
 class QNR_indef_results(QNR_results):
-    """Specialized for indefinite QB"""
+    """Specialized for indefinite QB
+
+    Indefinite is special, in that once done - don't offer again, even if
+    the protocol changes.
+
+    """
 
     def __init__(self, user, qb_id):
         self.user = user
+        # qb_id is the current indef qb - irrelevant if done in previous
+        self.qb_id = qb_id
+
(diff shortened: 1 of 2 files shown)


mcjustin (approved):

portal/models/questionnaire_response.py
```diff
@@ -523,16 +535,48 @@ def __init__(self, user, qb_id):
...
+        """Return set of completed Questionnaire results for Indefinite"""
+        # ignore the qb_id - as prior protocol versions picked up also count
+        return {
+            qnr.instrument for qnr in self.qnrs
+            if qnr.status == "completed"}
+
+    def partial_qs(self, qb_id, iteration):
+        """Return set of partial Questionnaire results for Indefinite"""
+        # ignore the qb_id - as prior protocol versions picked up also count
+        return {
```
ivan-c, line 559: Should these return lists or filtered Queryset for consistency?

ivan-c (approved):
Looks good! Had a minor question but no blockers

portal/models/questionnaire_response.py
pbugni: no - sigh - it's a bit of a mess.  there are comments elsewhere explaining how required returns a list as it's ordered (only makes sense for non indef, but you get the idea) where the others return sets for easy set method checks.

PR #3771, opened 2020-09-17
TN-2563 Significant refactor towards supporting concurrent QBs

pbugni:
Significant refactor towards supporting concurrent questionnaire banks assigned to a user, as per research study.
In general, this means passing or including research study id logic in all the questionnaire bank and assessment plumbing.

A few TODOs remain noting unfinished work, that's beyond the immediate scope of this PR.


portal/views/portal.py
```diff
@@ -845,6 +855,14 @@ def settings():
             locale_code=user.locale_code)
 
     if form.patient_id.data and form.timestamp.data:
+        # TODO: is this in use?  would change all a user's QNR authored
+        # dates - guessing it is dead.  Remove soon unless needed
+        msg = (
```
ivan-c, line 860: Is this the form to back-date QuestionnaireResponses? If so, might still be helpful for testing

ivan-c (approved):
Looks great! No obvious issues, but may have missed the forest for the trees :evergreen_tree:

portal/views/portal.py
pbugni: that was the original intended use - @mcjustin used it for exactly that.  since then we've added the ability to easily change a users consent date.

PR #3773, opened 2020-09-24
TN-2765 backend changes to enable patient, clinician association.

pbugni:
Includes model changes and an api to obtain the list of clinicians the current staff user has access to @ /api/clinician


ivan-c (approved):
Looks good!

PR #3774, opened 2020-09-24
Update IRONMAN research study names (EMPRO, and clarify global/main)

mcjustin:


mcjustin:
Redundant with https://github.com/uwcirg/truenth-portal/pull/3772

PR #3781, opened 2020-09-29
TN-2768 extend organization API for research study needs

pbugni:
The /api/organization/<org_id> endpoint now takes a query string parameter to include inherited information, such as a research protocol defined not at the requested org level, but somewhere up the org tree.

Research Protocols now include research_study_id in their serialized form.

Together, the front end can now call `/api/organization/<org_id>?include_inherited_attributes=true` to check for research study enrollment at any level.

### DIFF
--- a/portal/models/extension.py
+++ b/portal/models/extension.py
@@ -26,16 +26,22 @@ class CCExtension(Extension, metaclass=ABCMeta):
     def children(self):  # pragma: no cover
         pass
 
-    def as_fhir(self, include_empties=True):
+    def as_fhir(self, include_empties=True, include_inherited=False):
         if self.children.count():
             return {
                 'url': self.extension_url,
                 'valueCodeableConcept': {
                     'coding': [c.as_fhir() for c in self.children]}
             }
-        # Return valid empty if none are currently defined and requested
         if include_empties:
+            # Return valid empty if none are currently defined and requested
             return {'url': self.extension_url}
+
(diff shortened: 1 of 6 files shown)


portal/models/extension.py
```diff
@@ -26,16 +26,22 @@ class CCExtension(Extension, metaclass=ABCMeta):
...
                 'valueCodeableConcept': {
                     'coding': [c.as_fhir() for c in self.children]}
             }
-        # Return valid empty if none are currently defined and requested
         if include_empties:
+            # Return valid empty if none are currently defined and requested
             return {'url': self.extension_url}
+
+        if include_inherited:
+            # TODO in the organization instance, need a `parent()`
```
mcjustin, line 41: Not clear to me the impact of this remaining a TODO...

mcjustin (approved):
Naive question, otherwise makes sense to me

portal/models/extension.py
pbugni: This method is located in the `CCExtension` class, which is responsible for serialization of CodeableConcepts.  Its only use in organizations is `LocaleExtension`.  In other words, until this todo is tackled, an organization's locales will only show in the FHIR format if they're defined at the requested org's level - we're not yet supporting `include_inherited_attributes` for organization locales.

achen2401 (approved):
looks great, thank you, Paul!

PR #3784, opened 2020-09-30
Treating clinician section in profile

achen2401:
Address https://jira.movember.com/browse/TN-2713
Add a Treating Clinician section in the patient profile page for a patient that is part of the sub-study
Changes include:
- adding new Treating Clinician in template and corresponding frontend code to fill Clinician select list via ajax call to clinician service
- add check to display section only if the subject is part of the sub-study (he should have consented to the sub-study)
- bug fix to make sure when saving clinician data, previous careProvider data for organization is not replaced

### DIFF
--- a/portal/static/js/src/modules/TnthAjax.js
+++ b/portal/static/js/src/modules/TnthAjax.js
@@ -296,6 +296,24 @@ export default { /*global $ */
             }
         });
     },
+    "getResearchStudies": function(userId, params, callback) {
+        callback = callback || function() {};
+        if (!userId) {
+            callback({error: i18next.t("User id is required.")});
+            return false;
+        }
+        this.sendRequest("/api/user/" + userId + "/research_study", "GET", userId, params, function(data) {
+            if (data) {
+                if (!data.error) {
+                    callback(data);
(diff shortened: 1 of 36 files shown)


pbugni (approved):
Looks tight - well done.

portal/static/js/src/modules/TnthAjax.js
```diff
@@ -272,6 +272,60 @@ export default { /*global $ */
...
+            callback({error: i18next.t("User id is required.")});
+            return false;
+        }
+        this.sendRequest("/api/user/" + userId + "/research_study", "GET", userId, params, function(data) {
+            if (data) {
+                if (!data.error) {
+                    callback(data);
+                    return true;
+                } else {
+                    callback({"error": i18next.t("Server error occurred retrieving research study information.")});
```
ivan-c, line 311: Are you able to reuse an existing "server error" error string here? I'd prefer not to have a error message (and 10+ translations) for every API we implement

ivan-c (approved):
Looks good! Minor i18n nitpick but ready to merge

PR #3786, opened 2020-10-01
TN-2691 rewire assessment engine view / card html generation

pbugni:
The requirements for Assessment Engine cards have evolved beyond the intent of intervention strategies.  To implement the complicated designs which require state checks known only to the portal (not the intervention), use standard view and template tool sets.

This PR does not alter the end user's perspective, simply ports from the abusive use of an intervention access check to update database records with HTML for each user, to a view for the assessment engine.  Note, only available on EPROMs.

This will enable implementation of the sub-study changes needed in a subsequent PR.


portal/eproms/views.py
```diff
@@ -72,6 +73,75 @@ def landing():
...
+    now = datetime.utcnow()
+
+    # TODO handle research study id.  Patient must be done with study 0
+    #  before seeing any study 1 work.
+    research_study_id = 0
+    assessment_status = QB_Status(
+        user=user,
+        research_study_id=research_study_id,
+        as_of_date=now)
+    current_app.logger.debug("{}".format(assessment_status))
```
ivan-c, line 105: leftover debugging?

```suggestion
    current_app.logger.debug("%s", assessment_status)
```

portal/migrations/versions/1675585e7031_.py
```diff
@@ -0,0 +1,30 @@
...
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
+revision = '1675585e7031'
+down_revision = '6395e14d5ee5'
+
+
+def upgrade():
+    """Assessment Engine has its own view, no longer pushing HTMl into db
```
ivan-c, line 18:
```suggestion
    """Assessment Engine has its own view, no longer pushing HTML into db
```

portal/views/clinician.py
```diff
@@ -17,15 +17,18 @@
...
-            tuple(limit_to_orgs))).with_entities(
-            User.id, User.first_name, User.last_name)
+        Role.name == ROLE.CLINICIAN.value).with_entities(
+        User.id, User.first_name, User.last_name)
+
+    limit_to_orgs = org_restriction_by_role(acting_user, None)
+    if limit_to_orgs:
+        query = query.join(UserOrganization).filter(
+            UserOrganization.user_id == User.id).filter(
+            UserOrganization.organization_id.in_(tuple(limit_to_orgs)))
```
ivan-c, line 30: Why tuple and not a set here?

tests/test_intervention.py
```diff
@@ -577,24 +577,15 @@ def test_card_html_update(
...
-        db.session.add(strat)
-        db.session.commit()
-    user, ae, metastatic_org = map(
-        db.session.merge, (test_user, ae, metastatic_org))
+    # Fetch home page, which should include Assessment Engine "card"
+    results = client.get("/home")
 
     # without completing an assessment, card_html should include username
-    assert user.display_name in ae.display_for_user(user).card_html
+    assert bytes(test_user.display_name, 'utf-8') in results.data
```
ivan-c, line 588: surprised `results.data` isn't a string...

ivan-c (approved):
Looks great! had some questions and nitpicks but no blockers

portal/views/clinician.py
pbugni: old required SQL alchemy syntax.  looks like it's now supported, thanks.

tests/test_intervention.py
pbugni: yup.  joys of py3, i couldn't find a simpler way to look for a "string" in the resulting HTML - open to suggestions :)

PR #3791, opened 2020-10-06
TN-2773 clinician name in patient list

pbugni:
The patient query for sub-study patients now populates a `user.clinician` attribute.    It will have a value of `None` when the patient does not have an assigned clinician.


achen2401 (approved):
thanks, Paul

PR #3803, opened 2020-10-15
TN-2541-empro-qb-for-patient-temp

mcjustin:
Since this QB is blocking the ability to administer the empro patient questionnaire and thus test things like trigger evaluation, our temporary approach for this story is to make the empro patient QB contingent on the empro consent date only (removing the complex requirement to also factor in the state of the ironman global QB).

### DIFF
--- a/portal/config/eproms/QuestionnaireBank.json
+++ b/portal/config/eproms/QuestionnaireBank.json
@@ -720,6 +720,29 @@
     },
     {
       "resourceType": "QuestionnaireBank",
+      "classification": "baseline",
+      "name": "ironman_ss_baseline",
+      "start": "{\"days\": 0}",
+      "due": "{\"days\": 0}",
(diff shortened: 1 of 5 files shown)


mcjustin:
@pbugni  you might double-check my config there, it's always tricky. Related slack convo here https://cirg.slack.com/archives/D2TK044J1/p1584052177017800

mcjustin:
@ivan-c  any idea why this pipeline failed? I attempted to 'Retry', but got "An error occurred while making the request." (my page was fresh)

pbugni:
> @ivan-c any idea why this pipeline failed? I attempted to 'Retry', but got "An error occurred while making the request." (my page was fresh)

This is an intermittent error we've seen before, as our docs link into external sources - looks up again, but nothing we can do short of removing our documentation links to werkzeug.

PR #3805, opened 2020-10-19
TN-2794 don't insert skipped visits if posted results tie user to older RP

pbugni:
Complicated situation discovered thanks to an assertion in the reporting code.

If a user happens to have posted QNRs that deterministically tie them to the older RP (say they completed `ironmisc` on v3 versus `ironmisc_v5`) AND the timing is such that the older protocol (say v3) didn't include a visit (such as month 33), do NOT insert that skipped visit into their timeline, as we can't move back and forth between protocols.

### DIFF
--- a/portal/models/qb_timeline.py
+++ b/portal/models/qb_timeline.py
@@ -534,6 +534,11 @@ def ordered_qbs(user, classification=None):
                 period_instruments = set(
                     [q.instrument for q in qnrs_for_period])
                 if not transition_now and period_instruments & cur_only:
+                    # Posted results tie user to the old RP; clear skipped
+                    # state as it's unavailable when results from the "next"
+                    # exist.
+                    rp_flyweight.skipped_nxt_start = None
+
                     # Don't transition yet, as definitive work on the old
                     # (current) RP has already been posted, UNLESS ...
                     if period_instruments & next_only:
--- a/portal/models/qbd.py
+++ b/portal/models/qbd.py
@@ -71,3 +71,12 @@ class QBD(object):
         results['iteration'] = self.iteration
         results['visit'] = visit_name(self)
         return results
+
+    def __repr__(self):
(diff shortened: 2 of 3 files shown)


mcjustin:
> 
> 
> Complicated situation discovered thanks to an assertion in the reporting code.
> 
> If a user happens to have posted QNRs that deterministically tie them to the older RP (say they completed `ironmisc` on v3 versus `ironmisc_v5`) AND the timing is such that the older protocol (say v3) didn't include a visit (such as month 33), do NOT insert that skipped visit into their timeline, as we can't move back and forth between protocols.

@pbugni to clarify an edge case: staff would not be able to add a "paper" (backdated) month 33 in this case, correct? It's not a very realistic use case, and I don't think that's a blocker - just looking for clarity.

pbugni:
> > Complicated situation discovered thanks to an assertion in the reporting code.
> > If a user happens to have posted QNRs that deterministically tie them to the older RP (say they completed `ironmisc` on v3 versus `ironmisc_v5`) AND the timing is such that the older protocol (say v3) didn't include a visit (such as month 33), do NOT insert that skipped visit into their timeline, as we can't move back and forth between protocols.
> 
> @pbugni to clarify an edge case: staff would not be able to add a "paper" (backdated) month 33 in this case, correct? It's not a very realistic use case, and I don't think that's a blocker - just looking for clarity.

Correct - once this is published, you may review the /timeline for the patient.  As they submitted deterministic results on v3 for month 36, they transition to v5 as of month 39.  No month 33 in v3, therefore a paper submission for month 33 would not get an associated QB.

mcjustin:
Just noting that I added some of these business rules to https://promowiki.movember.com/display/IR/Questionnaire+Schedule#QuestionnaireSchedule-RelatedBusinessRules

mcjustin (approved):

PR #3807, opened 2020-10-20
Fixup constraint migration

ivan-c:
* Remove both versions of unique email constraint, if any exist


pbugni (approved):
Good find, not sure how i missed the obvious!

PR #3812, opened 2020-10-23
Bug fix for eproms staff resources page links

achen2401:
address  https://jira.movember.com/browse/TN-2813
Links in EPROMS resources page link to incorrect LR content

Issue stems from code relying on getting the topic tag for the work instruction based on its index within the asset tags array - the index can change however, as tags in the array can change its order

Fix includes:
-  filter for only topic tag in the asset tags array so that the template can use the correct tag to build the url that displays the LR content

### DIFF
--- a/portal/eproms/views.py
+++ b/portal/eproms/views.py
@@ -360,7 +360,7 @@ def resources():
         for asset in results:
             if 'demo' in asset['tags']:
                 demo_content.append(asset_by_uuid(asset['uuid']))
-            # topic tag for the work instruction
+            # filter for topic tag for the work instruction
             asset['topics'] = [tag for tag in asset['tags'] if tag not in (
                 asset_title, 'individual work instruction')]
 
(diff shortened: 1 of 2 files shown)


mcjustin (approved):

pbugni (approved):

portal/eproms/views.py
```diff
@@ -352,14 +352,17 @@ def resources():
     org = user.first_top_organization()
     if not org:
         abort(400, 'user must belong to an organization')
-    resources_data = get_any_tag_data(
-        '{} work instruction'.format(org.name.lower()))
+    asset_title = '{} work instruction'.format(org.name.lower())
```
ivan-c, line 355: should this string be translated?
`'{} work instruction'`

ivan-c (approved):
Had a question about i18n, ready to merge otherwise

portal/eproms/views.py
achen2401: I believe Diana mentioned that no needs for it to be translated:  https://mo-programs.slack.com/archives/C6JFH0C4A/p1558306641000700

PR #3813, opened 2020-10-26
TN-2781 Integrate SDC service

ivan-c:
TN-2781
* Integrate SDC at service level

### DIFF
--- a/portal/tasks.py
+++ b/portal/tasks.py
@@ -24,6 +24,7 @@ from .factories.app import create_app
 from .factories.celery import create_celery
 from .models.communication import Communication
 from .models.communication_request import queue_outstanding_messages
+from .models.observation import Observation
 from .models.qb_status import QB_Status
 from .models.qb_timeline import invalidate_users_QBT, update_users_QBT
 from .models.reporting import (
@@ -382,6 +383,7 @@ def extract_observations(questionnaire_response_id):
     obs_bundle = response.json()
 
     for obs in obs_bundle['entry']:
-        observation = Observation.update_from_fhir(obs)
+        observation = Observation()
+        observation.update_from_fhir(obs)
         db.session.add(observation)
     db.session.commit()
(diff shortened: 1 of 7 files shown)


portal/tasks.py
```diff
@@ -356,3 +356,32 @@ def celery_beat_health_check(**kwargs):
...
+    #from .trigger_states.empro_states import enter_user_trigger_critical_section
+    qnr = QuestionnaireResponse.query.get(questionnaire_response_id)
+
+    # TODO enable after merging
+    #enter_user_trigger_critical_section(user_id=qnr.subject_id)
+
+    qnr_json = qnr.as_sdc_fhir()
+
+    SDC_BASE_URL = current_app.config['SDC_BASE_URL']
+    response = post(f"{SDC_BASE_URL}/$extract", json=qnr_json)
```
pbugni, line 379: this looks like a blocking call?  worth a comment?

pbugni (approved):
awesome progress

portal/views/assessment_engine.py
```diff
@@ -1575,6 +1575,10 @@ def assessment_add(patient_id):
     db.session.commit()
     questionnaire_response.assign_qb_relationship(
         acting_user_id=current_user().id)
+
+    from ..tasks import extract_observations_task
```
pbugni, line 1579: shouldn't we limit this expensive call to QNRs that include the `sdc-questionnaire-observationExtract` extension?

portal/views/assessment_engine.py
```diff
@@ -1575,6 +1575,10 @@ def assessment_add(patient_id):
...
+    extract_observations_task.apply_async(questionnaire_response_id=questionnaire_response.id)
```
pbugni, line 1580: don't we need to do this in the PUT / update QNR call as well?

ivan-c: Done! Thanks for the reminder

pbugni (approved):

portal/views/assessment_engine.py
ivan-c: Thanks, as discussed added a hack and a TODO so we don't forget :crossed_fingers:

PR #3814, opened 2020-10-26
TN-2719 trigger states / engine

pbugni:
Implementation of EPROMs sub-study trigger engine.

Includes an API for the front-end to query at any time to get a user's trigger status:
```/api/user/<int:user_id>/triggers```

As well as a function that wraps a REDIS lock per user during the trigger evaluation process (including SDC / celery jobs, etc.) via:
```portal.trigger_states.empro_states.enter_user_trigger_critical_section(user_id)```

TODOs mark multiple integration points awaiting TN-2781 implementation.

### CODE REVIEW

File: portal/models/questionnaire_response.py [hunk 1]
```diff
@@ -906,3 +919,33 @@ def author_role(row_data, qnr_data):
                         row_data['option_text'] = strip_tags(
                             answer['valueCoding'].get('text', None))
                 yield {k: v for k, v in row_data.items() if v is not None}
+
+
+def first_last_like_qnr(qnr):
+    """Specialized lookup function to return similar QNRs
+
+    :param qnr: reference QNR - look for first and most recent QNRs similar
+      to the one provided.  That is owned by the same subject and QB.
```
mcjustin, line 928: @pbugni  Since we have two QB's for the EMPRO patient questionnaire now (baseline and recurring), don't you need a TODO here to accommodate?

portal/trigger_states/empro_domains.py
```diff
@@ -0,0 +1,116 @@
...
+                    self._triggers.add('hard')
+
+    def eval(self):
+        """Use instance data to evaluate trigger state for domain"""
+        # consider caching - for now, recalc when called
+        self._triggers.clear()
+
+        # check if current data includes hard trigger
+        for value in self.current_answers.values():
+            if value >= 4:
```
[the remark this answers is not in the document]
pbugni: hm, i had the impression the SDC was going to deliver a 1-5 normalized index.  we have questions w/ only 4 options in this Questionnaire?

portal/models/questionnaire_response.py
pbugni: d'OH!  right you are, I'll refactor that to fit.

portal/trigger_states/empro_domains.py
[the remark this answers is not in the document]
mcjustin: We do indeed have one question (social isolation) which has only 4 options, not 5. Can't speak to the SDC though, I'll defer to @ivan-c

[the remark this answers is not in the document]
ivan-c: Justin is correct- The SDC service will only return one of the below 4 Observations. I could add another coding to the Questionnaire options (to extract) to indicate soft/hard trigger, but there may be other ways...
```
{
    "text": "During the past week: Has your physical condition or medical treatment interfered with your social activities?",
    "type": "choice"
    "code": [
    {
        "code": "social_isolation",
        "display": "Social isolation",
        "system": "http://us.truenth.org/observation"
    }
    ],
    "linkId": "ironman_ss.20",
    "option": [
        {
            "valueCoding": {
            "code": "ironman_ss.20.1",
            "display": "Not at all"
            }
        },
        {
            "valueCoding": {
            "code": "ironman_ss.20.2",
            "display": "A little"
            }
        },
        {
            "valueCoding": {
            "code": "ironman_ss.20.3",
            "display": "Quite a bit"
            }
        },
        {
            "valueCoding": {
            "code": "ironman_ss.20.4",
            "display": "Very much"
            }
        }
    ]
}
```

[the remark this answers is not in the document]
ivan-c: This was discussed yesterday, most severe and second most severe Observations will be tagged with an extension on the `valueCoding` indicating them as `penultimate` and `last`

portal/trigger_states/models.py
```diff
@@ -0,0 +1,56 @@
...
+    NB only rows with `state` = `processed` include triggers.
+
+    """
+    __tablename__ = 'trigger_states'
+    id = db.Column(db.Integer, primary_key=True)
+    user_id = db.Column(db.ForeignKey('users.id'), nullable=False, index=True)
+    state = db.Column('state', trigger_state_enum, nullable=False, index=True)
+    timestamp = db.Column(
+        db.DateTime, default=datetime.utcnow, nullable=False, index=True)
+    triggers = db.Column(JSONB)
```
ivan-c, line 27: I think this is a list of FHIR references to QuestionnaireResponse resources- Why not a join table to the associated QuestionnaireResponses (or Observations)? Seems like this will be a commonly looked up value

pbugni: I've been waiting on the clarity, but agree that association with a QNR for a row with triggers makes the most sense.  A bit puzzled how to best handle the non-real-world situation where a user has multiple completed QNRs for a single "visit".

portal/models/questionnaire_response.py
pbugni: resolved via https://github.com/uwcirg/truenth-portal/pull/3814/commits/8fbef49d0fac2919324c1ea55552ed3a9b7ec593

portal/trigger_states/empro_states.py
```diff
@@ -0,0 +1,392 @@
...
+    from ..date_tools import FHIR_datetime
+
+    def send_n_report(em, context, record):
+        """Send email, append success/fail w/ context to record"""
+        result = {'context': context, 'timestamp': FHIR_datetime.now()}
+        try:
+            em.send_message()
+            result['email_message_id'] = em.id
+            record.append(result)
+        except SMTPRecipientsRefused as e:
```
ivan-c, line 238: Should we handle that exception in the [logging module](https://github.com/uwcirg/truenth-portal/blob/develop/portal/logs.py#L61-L64)?

pbugni: As we're catching here - shouldn't ever hit.  No, that clause in the logging module is a backdoor to break out of a cmd line execution.

PR #3857, opened 2020-11-24
Feature/ui submit test questionnaire data

achen2401:
Per internal meeting, adding UI Test button on patient home page to allow **manual** submission of main study questionnaire responses.

Changes include:
- make questionnaire data submission button available **ONLY** when main study assessment is available (status of due or overdue).  NOTE: Only main study questionnaire data will be submitted.  User will still need to go through the sub-study questionnaire (for sub-study subject)
- the button is ONLY available in the Test environment
- the button is activate via the combination key press of Ctrl, Shift and "d" letter keys
- make portal JS code buildable


mcjustin (approved):
Looks good to me. I'm not sure about the hotkey requirement, but let see how it goes!

ivan-c (approved):

PR #3858, opened 2020-11-24
remove date range restriction for consent datepicker

achen2401:
Remove date range constraint on the allowable dates to be selected from date picker for consent date.
Note again: datepicker for consent date is only available in the test environment


mcjustin (approved):
@achen2401  you didn't ask me to review, but this looks right to me - just removing the date restriction.

PR #3863, opened 2020-12-01
Display sub-study tailored content based on trigger state readiness

achen2401:
Frontend code changes are needed as back-end processing of triggers is asynchronous.
Changes include:

- Add frontend code - to that for display in post-questionnaire thank you modal and to that for sub-study tailored content for domain topics - to poll the readiness of trigger information (trigger information is considered ready to consume if the trigger state is processed, triggered o resolved.
- the frontend code will poll a maximum of 5 tries (beyond that seems too long for pleasant user experience)
- fix bug related to loading video content
- styling updates


portal/eproms_substudy_tailored_content/src/js/Base.js
```diff
@@ -202,24 +202,50 @@ export default {
...
-                    this.processDefaultDomainContent();
+                    processedData = JSON.parse(response[0].value);
                 } catch(e) {
                     //log error to console
                     this.setErrorMessage(`Error parsing trigger data ${e}`);
                 }
+
+                if (processedData &&
+                    params.retryAttempt < params.maxTryAttempts &&
+                    //if the trigger data has not been processed, try again until maximum number of attempts has been reached
```
pbugni, line 229: I may have overlooked it, but you should only bother to poll IF the state is "inprocess".  Checking against `processedTriggerStates` would also wait if the user is `unstarted` or `due`.

pbugni (approved):

mcjustin (approved):

portal/eproms_substudy_tailored_content/src/js/Base.js
achen2401: right, the check won't be initiated if the triggers have been processed.

ivan-c (approved):

PR #3865, opened 2020-12-02
TN-2858 enable model level `keep_unmentioned` for site_persistence

pbugni:
With the addition of the SDC generating novel codings from observations after scoring QNRs, the site_persisted type `Codings` can no longer be purged of values not mentioned in the `Coding.json` file.

Now, each ModelDetails tuple includes a keep_unmentioned override, should the type need to be retained regardless of the global setting.


ivan-c (approved):
:tada:

PR #3866, opened 2020-12-03
Refactor / clean-up sub-study eligibility checks

pbugni:
Determine patient's research study eligibility in consistent manner.
Prevent qb_timeline row additions prior to patient's eligibility.
Provide bootstrap safe lookup.

Previously, sub-study qb_timeline rows were erroneously added for patients w/o a clinician.  This should be run on test dbs after merge.  (No, not worth a migration IMO)
`delete from qb_timeline where user_id in (select id from users where clinician_id is null and id in (select distinct(user_id) from qb_timeline where research_study_id = 1));`

### DIFF
--- a/portal/eproms/views.py
+++ b/portal/eproms/views.py
@@ -96,7 +96,7 @@ def assessment_engine_view(user):
     from ..models.research_study import BASE_RS_ID, EMPRO_RS_ID, ResearchStudy
     now = datetime.utcnow()
 
-    research_study_status = patient_research_study_status(user)
+    research_study_status = patient_research_study_status(user, False)
     assessment_status = QB_Status(
         user=user,
         research_study_id=BASE_RS_ID,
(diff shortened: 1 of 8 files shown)


achen2401 (approved):

portal/eproms/views.py
```diff
@@ -96,7 +96,7 @@ def assessment_engine_view(user):
     from ..models.research_study import BASE_RS_ID, EMPRO_RS_ID, ResearchStudy
     now = datetime.utcnow()
 
-    research_study_status = patient_research_study_status(user)
+    research_study_status = patient_research_study_status(user, False)
```
ivan-c, line 99: flags are a little more obvious with keyword args
```suggestion
    research_study_status = patient_research_study_status(user, ignore_QB_status=False)
```

ivan-c (approved):
Looks great! Minor nitpick, but ready to merge

PR #3872, opened 2020-12-07
remove testing coverage as it's breaking build

pbugni:
mysterious upstream change breaking build when dependent on `pytest_cov` - remove for time being; bring back once resolved.


pbugni:
wrong - deleting.

PR #3885, opened 2020-12-14
TN-2577 Add flask-session override for SameSite cookie settings

ivan-c:
TN-2577

* Add flask-session override class ([base class](https://github.com/fengsp/flask-session/blob/0.3.2/flask_session/sessions.py))
* Include cookies in cross-site requests (eg portal banner) with [`SameSite=None`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie/SameSite#None)
* Update dependencies as necessary

### DIFF
--- a/portal/session.py
+++ b/portal/session.py
@@ -57,7 +57,7 @@ class BaseRedisSessionInterface(RedisSessionInterface):
         )
 
 
-class SameSiteSession(Session):
+class RedisSameSiteSession(Session):
     """Extends flask-session to passthrough SESSION_COOKIE_SAMESITE flask config"""
     def _get_interface(self, app):
         config = app.config.copy()
@@ -66,15 +66,6 @@ class SameSiteSession(Session):
         config.setdefault('SESSION_USE_SIGNER', False)
         config.setdefault('SESSION_KEY_PREFIX', 'session:')
         config.setdefault('SESSION_REDIS', None)
-        config.setdefault('SESSION_MEMCACHED', None)
-        config.setdefault('SESSION_FILE_DIR', os.path.join(os.getcwd(), 'flask_session'))
-        config.setdefault('SESSION_FILE_THRESHOLD', 500)
-        config.setdefault('SESSION_FILE_MODE', 384)
-        config.setdefault('SESSION_MONGODB', None)
-        config.setdefault('SESSION_MONGODB_DB', 'flask_session')
-        config.setdefault('SESSION_MONGODB_COLLECT', 'sessions')
-        config.setdefault('SESSION_SQLALCHEMY', None)
-        config.setdefault('SESSION_SQLALCHEMY_TABLE', 'sessions')
         session_interface = BaseRedisSessionInterface(
             config['SESSION_REDIS'],
             config['SESSION_KEY_PREFIX'],
--- a/portal/session.py
+++ b/portal/session.py
@@ -1,0 +1,83 @@
+import os
+
+from flask_session import Session, RedisSessionInterface
+from flask_session.sessions import total_seconds
+
+class BaseRedisSessionInterface(RedisSessionInterface):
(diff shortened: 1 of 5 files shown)


portal/session.py
```diff
@@ -0,0 +1,84 @@
...
+from flask_session.sessions import total_seconds
+
+
+class BaseRedisSessionInterface(RedisSessionInterface):
+    """Override RedisSessionInterface save_session() method to passthrough flask config"""
+
+    def save_session(self, app, session, response):
+        domain = self.get_cookie_domain(app)
+        path = self.get_cookie_path(app)
+        if not session:
```
pbugni, line 13: is this `not` a typo?  w/i a negated test for session, you check if it's modified?

requirements.txt
```diff
@@ -60,5 +60,5 @@ swagger-spec-validator==2.4.3
 urllib3==1.25.7             # via requests
 validators==0.14.0 # pyup: <=0.10.1 # pin until require_tld supported again
 vine==1.3.0               # via amqp
-werkzeug==0.16.0          # via flask
+werkzeug==1.0.1           # via flask
```
pbugni, line 63: 👍

portal/session.py
```diff
@@ -0,0 +1,84 @@
...
+        config.setdefault('SESSION_TYPE', 'null')
+        config.setdefault('SESSION_PERMANENT', True)
+        config.setdefault('SESSION_USE_SIGNER', False)
+        config.setdefault('SESSION_KEY_PREFIX', 'session:')
+        config.setdefault('SESSION_REDIS', None)
+        config.setdefault('SESSION_MEMCACHED', None)
+        config.setdefault('SESSION_FILE_DIR', os.path.join(os.getcwd(), 'flask_session'))
+        config.setdefault('SESSION_FILE_THRESHOLD', 500)
+        config.setdefault('SESSION_FILE_MODE', 384)
+        config.setdefault('SESSION_MONGODB', None)
```
pbugni, line 73: Looks like all permutations of SESSION storage are covered.  Is this just in case we adopt a different session type, or because it's also done in the base class, or ?

My concern is maintenance - I'd hope we can just inherit the defaults, surprised we'd need to call out so many?

portal/session.py
ivan-c: I'll preface that most of the code in these overridden methods is copied directly from the parent class they extend: [`RedisSessionInterface`](https://github.com/fengsp/flask-session/blob/0.3.2/flask_session/sessions.py#L90-L173)

If I understand correctly, `session.modified` tracks if the incoming session data differs from the existing session data.
This nested condition handles the case where the session data has been cleared, but hadn't been reflected in the server-side persistent storage (redis) yet.

requirements.txt
ivan-c: I try to minimize surprise breakage by sneaking a few dependency updates into each release, but we haven't had one recently, so we're playing a bit of catch up here :tomato:

portal/session.py
ivan-c: >Looks like all permutations of SESSION storage are covered. Is this just in case we adopt a different session type, or because it's also done in the base class, or ?

Yup, I just copied too much from the [base class](https://github.com/fengsp/flask-session/blob/0.3.2/flask_session/__init__.py#L63-L109), when we only really care about supporting redis. I'll go ahead and remove references to other server-side session backends

ivan-c:
@mcjustin @pbugni FYI I'm running into issues with [`flask-oauthlib`](https://github.com/lepture/flask-oauthlib) which relies on an older version of a dependency (werkzeug) we need to update to support the new cookie values for `SameSite`

I may have to fork `flask-oauthlib` in the short term.

achen2401 (approved):

pbugni (approved):
Thank you!

PR #3886, opened 2020-12-15
TN-2869 Hotfix: correct RP v3 -> v5 overlapping QB transition problems

pbugni:
v3 -> v5 research protocol transition exposed a number of problems with overlapping QBs.

Previously, configurations were carefully managed to prevent any given visit from overlapping with another.  However the transition from v3 to v5 breaks this in a number of ways.

The current evaluation process for the set of questionnaire banks that apply to a user work on a series of generators that manage the applicable list of research protocols (RP), and the respective associated QBs.  When to transition and what visits are safe to include must also check with the available results.  The generator model doesn't always have available scope to check state of filed results, or to look ahead or back to check for status.

This PR includes a number of enhanced checks and hacks such as:
- qb_timeline included visits should be removed due to overlap with existing results
- QNR results fit better in the first or second QB when the authored date straddles multiple
- rolling RPs forward 2x for an organization with 3 RPs all defined before new patient is added (and preventing this logic from incorrectly rolling forward too far the indefinite case)
- report when QNR.authored != QNR.document['authored'] (as it leads to mis-classified QNRs)
- correct associations and status when a visit is removed due to overlap and posted results.


portal/models/qb_timeline.py
```diff
@@ -559,7 +570,13 @@ def ordered_qbs(user, classification=None):
                         if q.instrument not in common:
                             continue
                         qnr = QuestionnaireResponse.query.get(q.qnr_id)
-                        qnr.qb_id = rp_flyweight.nxt_qbd.qb_id
+                        assert qnr.authored >= rp_flyweight.nxt_start
+                        # when RP dates don't align, attempt an advance
+                        if qnr.authored > rp_flyweight.nxt_exp:
+                            rp_flyweight._advance_nxt()
+                            if qnr.authored > rp_flyweight.nxt_exp:
+                                raise ValueError(f"transition error on qnr {qnr.id}")
```
mcjustin, line 578: @pbugni  looks like a syntax error here

mcjustin (changes requested):

portal/models/qb_timeline.py
pbugni: how so?

pbugni: the double check is intentional - might need to advance one.  if that doesn't catch us up, we're in an error state.

mcjustin: Sorry, I likely misread my pylint there w/ the "f"

mcjustin:
@pbugni  I think you said that part of this was resolved with https://jira.movember.com/browse/TN-2869 ? Does it merit another story?

portal/models/questionnaire_response.py
```diff
@@ -77,6 +77,14 @@ def default_authored(context):
...
     )
 
+    @property
+    def qb_id(self):
+        raise ValueError('questionnaire_bank_id referenced by wrong name `qb_id`')
+
+    @qb_id.setter
+    def qb_id(self, value):
+        raise ValueError('questionnaire_bank_id assignment to wrong name `qb_id`')
+
```
ivan-c, line 87: :safety_vest:

ivan-c (approved):

pbugni:
> @pbugni I think you said that part of this was resolved with https://jira.movember.com/browse/TN-2869 ? Does it merit another story?

@mcjustin That story may be related, but we don't have a reproduction or know that it's fixed, yet.  It's true, we don't really have a story capturing what the `qb_id` vs `questionnaire_bank_id` assignment during transition caused.  Feel free to open a story.

mcjustin:
@pbugni  new story for this:  https://jira.movember.com/browse/TN-2872

portal/migrations/versions/f93d9608c372_.py
```diff
@@ -32,7 +31,7 @@ def visit_name(qb_id, iteration):
     raise RuntimeError("not possible")
 
 
-def existing_qnr_deets(patient_id, missing_qb_id):
+def existing_qnr_deets(session, patient_id, missing_qb_id):
```
ivan-c, line 34: >deets

:laughing:

ivan-c (approved):
Looks good!

PR #3892, opened 2020-12-17
TN-2577 update frontend cookie testing code

achen2401:
sub-task for : https://jira.movember.com/browse/TN-2577
When frontend is testing if cookies are enabled in a browser, add secure cookie attribute to test cookie so that it won't be rejected by the browser.

### DIFF
--- a/portal/static/js/src/CookieMonster.js
+++ b/portal/static/js/src/CookieMonster.js
@@ -16,8 +16,8 @@ import Utility from "./modules/Utility.js";
         this.setCookie({"expires":d.toGMTString(), ...this.getSameSiteAttribute()});
     };
     CookieMonster.prototype.getSameSiteAttribute = function() {
-        //Do we need this set specifically to environment or ?
-        //if (String(this.settings["SYSTEM_TYPE"]).toLowerCase() === "development") return {};
+        //Do not set this for development environment
+        if (String(this.settings["SYSTEM_TYPE"]).toLowerCase() === "development") return false;
         return {
             "SameSite":  "None; Secure"
         }
@@ -188,7 +188,7 @@ import Utility from "./modules/Utility.js";
 
     };
     CookieMonster.prototype.getSettings = function (callback) {
-        Utility.newHttpRequest("/api/settings", false, callback);
+        Utility.newHttpRequest("/api/settings", {cache: true}, callback);
     }
     CookieMonster.prototype.init = function() {
         var self = this;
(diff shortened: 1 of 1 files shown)


portal/static/js/src/CookieMonster.js
```diff
@@ -11,13 +12,28 @@ import {getUrlParameter} from "./modules/Utility.js";
...
         var d = new Date();
         d.setTime(d.getTime() - (1000*60*60*24)); //set the time to the past. 1000 milliseonds = 1 second
-        var expires = "expires=" + d.toGMTString(); //compose the expirartion date
-        window.document.cookie = this.testCookieName+"="+"; "+expires;//set the cookie with name and the expiration date
+        //set the cookie with name and the expiration date
+        this.setCookie({"expires":d.toGMTString(), ...this.getSameSiteAttribute()});
+    };
+    CookieMonster.prototype.getSameSiteAttribute = function() {
+        //Do we need this set specifically to environment or ?
+        //if (String(this.settings["SYSTEM_TYPE"]).toLowerCase() === "development") return {};
```
achen2401, line 20: @ivan-c Ivan, in the story, you mentioned, `conditionally add Secure cookie attribute for non-development deploys`, do we not need it for other environments, e.g. development?

ivan-c: > @ivan-c Ivan, in the story, you mentioned, `conditionally add Secure cookie attribute for non-development deploys`, do we not need it for other environments, e.g. development?

Correct. If a cookie has the `Secure` attribute, it is required to be transmitted over https only. We do not want to set the `Secure` attribute on cookies for development deployments, otherwise no cookies will be set, and you will not be able to login into your development deploy.

We have the same functionality in the [flask backend here](https://github.com/uwcirg/truenth-portal/blob/develop/portal/config/config.py#L185-L186), just need it replicated in the frontend.

achen2401: thank you, @ivan-c, for clarifying, I did just put the conditional check for environment back now.

ivan-c (approved):
Looks great, thanks!

FYI I tried testing locally (docker image tag `feature-update-frontend-cookie-detection-code`) but I'm seeing the old version of `CookieMonster.js`. I'm going to assume it's some kind of caching issue.

achen2401:
thanks, @ivan-c yeah, it might be cached version of cookieMonster.js you are seeing.  I did test out (in FF) and my changes appear to be working.
This is what I observe:
In development system type when no SameSite secure attribute is set on the cookie:
Screen Shot 2020-12-17 at 1 45 37 PM
vs in production system type when SameSite secure attribute is set on the cookie:
Screen Shot 2020-12-17 at 1 44 06 PM

PR #3911, opened 2020-12-23
TN-2771 update sub-study patient list column names and order

achen2401:
per https://jira.movember.com/browse/TN-2771
Update column name and order for sub-study patient list.
Fix clinician action status from "due" to "required" as per specs.
See final specs here:  https://movember.app.box.com/s/fv78bk00lpru8scm99fex1ec09ocbtlv


pbugni (approved):
👍

PR #3916, opened 2020-12-29
Withdrawn sanity check should log error, not halt process

pbugni:
Couldn't generate an acceptance report due to a single user's mis-filed QNRs.

Generate an error log, don't halt the process via exception.

### DIFF
--- a/portal/models/qb_status.py
+++ b/portal/models/qb_status.py
@@ -143,8 +143,8 @@ class QB_Status(object):
                 self.overall_status != OverallStatus.withdrawn):
             from flask import current_app
             current_app.logger.error(
-                "Unexpected state %s, user %d should be withdrawn", 
-                    self.overall_status, self.user.id)
+                "Unexpected state %s, user %d should be withdrawn",
+                self.overall_status, self.user.id)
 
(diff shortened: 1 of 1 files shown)


mcjustin (approved):

portal/models/qb_status.py
```diff
@@ -141,9 +141,10 @@ def _status_from_current(self, cur_qbd):
         # Withdrawn sanity check
         if self.withdrawn_by(self.as_of_date) and (
                 self.overall_status != OverallStatus.withdrawn):
-            raise RuntimeError(
-                "Unexpected state {}, should be withdrawn".format(
-                    self.overall_status))
+            from flask import current_app
+            current_app.logger.error(
+                "Unexpected state {}, user {} should be withdrawn".format(
+                    self.overall_status, self.user.id))
```
ivan-c, line 147:
```suggestion
            current_app.logger.error(
                "Unexpected state %s, user %d should be withdrawn", 
                    self.overall_status, self.user.id)
```

ivan-c (approved):
Looks good! minor logging nitpick but no blockers

PR #3945, opened 2021-01-13
TN-2910 Removing `authored` column from QuestionnaireResponse

pbugni:
Removing `authored` column from QuestionnaireResponse as it duplicates (and drifts) from the single point of truth in QuestionnaireResponse.document['authored'].


portal/models/questionnaire_response.py
```diff
@@ -428,9 +413,9 @@ def qnrs(self):
             QuestionnaireResponse.status,
             QuestionnaireResponse.document[
                 ('questionnaire', 'reference')].label('instrument_id'),
-            QuestionnaireResponse.authored,
+            QuestionnaireResponse.document['authored'].label('authored'),
             QuestionnaireResponse.encounter_id).order_by(
-            QuestionnaireResponse.authored)
+            QuestionnaireResponse.document['authored'])
```
ivan-c, line 418: If I understand correctly, this is using the string methods (`__lt__()`/`__gt__()`) for alphabetical sorting.

Are we operating under the assumption that all datetimes are GMT-normalized? Do we need to cast it to a datetime-like object before sorting (so timezones could be normalized)? eg would `2020-11-05T09:15:30-08:00` correctly come before `2020-11-05T08:15:30-05:00`?

ivan-c (approved):
Looks great! Had a few questions but no blockers

portal/models/questionnaire_response.py
```diff
@@ -95,8 +83,9 @@ def qb_id(self, value):
 
     def __str__(self):
         """Print friendly format for logging, etc."""
+        authored = self.document and self.document.get('authored') or ''
```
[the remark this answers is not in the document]
pbugni: it's just used for logging, etc.  None doesn't look so nice :)

portal/models/questionnaire_response.py
pbugni: good point - i don't think it's a real world problem, but best to consider.

pbugni: wow, that was way harder than i expected.  in the end, i found the sort excessively cumbersome and instead, raise an exception if we ever find a situation where the resulting sort order is incorrect.

PR #3960, opened 2021-01-27
create global for MAX_ATTEMPTS to simplify debugging

pbugni:
In general, we use 3 attempts from front-end requests, should there be a blip that prevents an API request from responding.

When debugging code, this can be a hassle, as multiple threads concurrently hit breakpoints confusing the user.

Adding an obvious constant to make temporary changes to the number of attempts easy to adjust for dev purposes.


portal/static/js/src/modules/TnthAjax.js
```diff
@@ -1133,7 +1134,7 @@ export default { /*global $ */
             return false;
         }
         params = params || {};
-        this.sendRequest("/api/user/" + userId + "/questionnaire_bank", "GET", userId, {data: {as_of_date: completionDate}, sync: params.sync}, function(data) {
+        this.sendRequest("/api/user/" + userId + "/questionnaire_bank", "GET", userId, {data: {as_of_date: completionDate}, sync: params.sync, max_attempts: MAX_ATTEMPTS}, function(data) {
```
pbugni, line 1137: @achen2401 is this change necessary?  Wouldn't we pick up the correct `default` without including, as defined on line #21?

achen2401: @pbugni You are right, I was just thinking that. I think it is safe to remove the change here.

achen2401 (approved):

PR #3978, opened 2021-02-04
TN-2828-IRONMAN-ORG-QUT-REMOVE-PROTOCOL ...

mcjustin:
... QUT should never have been assigned a protocol, it never has patients - it's 3 children have patients, and they need to manage their protocol individually.

### DIFF
--- a/portal/config/eproms/Organization.json
+++ b/portal/config/eproms/Organization.json
@@ -378,12 +378,6 @@
         {
           "timezone": "Australia/Brisbane",
           "url": "http://hl7.org/fhir/StructureDefinition/user-timezone"
-        },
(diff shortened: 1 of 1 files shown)


mcjustin:
@pbugni  does this seem safe? As it stands, it's very confusing.

pbugni (approved):
Yes, it's safe and a good idea.
It will force the reprocessing of all child organization's patients with qb_timeline data (some 32 patients), but no problems discovered in the process.

mcjustin:
@ivan-c  I'll defer to you on merging this to master/develop...

ivan-c (approved):

PR #3980, opened 2021-02-04
TN-2919 CLI preview-site-update 

pbugni:
Add new CLI `preview-site-update` to present expected changes for patients on a Research Protocol change.

Example run:
```flask preview-site-update --retired "2021-02-01 13:00:00" --org_id 146102```

Results from said example available on parent Jira issue:  https://jira.movember.com/secure/attachment/99896/APC-RPv5-changes.txt

Not intended to be run on prod - checks config to confirm.

### DIFF
--- a/manage.py
+++ b/manage.py
@@ -431,12 +431,14 @@ def no_email(email, actor):
     suppress_email(email, actor)
 
 
-@click.option('--qnr_id', help="Questionnaire Response ID")
+@click.option('--qnr_id', help="Questionnaire Response ID", required=True)
 @click.option(
     '--authored',
+    required=True,
     help="new datetime for qnr authored, format example: 2019-04-09 15:14:43")
 @click.option(
     '--actor',
+    required=True,
     help='email address of user taking this action, for audit trail'
 )
 @app.cli.command()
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: manage.py [hunk 1]
```diff
@@ -586,3 +586,153 @@ def merge_users(src_id, tgt_id, actor):
         user_id=acting_user.id,
         subject_id=tgt_id)
     print(message)
+
+
+@click.option('--org_id', help="Organization (site) ID")
+@click.option(
+    '--retired',
+    help="datetime for site's current protocol expiration,"
+         " format example: 2019-04-09 15:14:43")
```
ivan-c, line 595:
```suggestion
@click.option('--org_id', help="Organization (site) ID", required=True)
@click.option(
    '--retired',
    required=True,
    help="datetime for site's current protocol expiration,"
         " format example: 2019-04-09 15:14:43")
```

I think both these are required, so `required=True` should be passed [as a keyword arg](https://click.palletsprojects.com/en/7.x/options/#basic-value-options)

>By default, options are not required, however to make an option required, simply pass in required=True as an argument to the decorator.

ivan-c: Regular (positional) [arguments](https://click.palletsprojects.com/en/7.x/arguments/) might make sense too, if there's some obvious ordering

ivan-c (approved):
Looks good! Minor nitpick but ready to merge

manage.py
pbugni: Yes, positional would make more POSIX sense, given they're required, but the ordering isn't obvious.  I like flagging them as required.

PR #3983, opened 2021-02-08
TN-2468 correct QBs due date to reflect desired `target_date`

pbugni:
Long history with the confusion between `start/available` and `due` dates for a QB.  The current implementation treats them ambiguously, maintaining both a `start` and `due` value on the QB itself, but calling the `start/available` date as `due` in the QB timeline and within all status checks.

TN-2468 addresses the confusion this creates for users, as they might get a reminder email for a 3 month prom that suggests it's `due` when it first becomes `available`.

At this point, this PR corrects:
- all EMPRO patient QBs, to have a due date 2 weeks after the start.
- all recurring IRONMAN QBs, to have a due date such that the original "visit" date is respected.  For example, a v3 3 month that first recurs at "3 months -15 days" now has a due of 15 days.
- the `due_date` variables used by the templates for AssessmentEngine cards now pick up the `target_date`, synonymous with the intent of `due_date` but not the same as `start/available`.

### DIFF
--- a/portal/config/eproms/QuestionnaireBank.json
+++ b/portal/config/eproms/QuestionnaireBank.json
@@ -262,7 +262,7 @@
     },
     {
       "classification": "recurring",
-      "due": "{\"days\": 0}",
+      "due": "{\"days\": 15}",
       "expired": "{\"months\": 3, \"days\": -2}",
       "name": "IRONMAN_recurring_3mo_pattern",
       "overdue": "{\"days\": 16}",
@@ -298,7 +298,7 @@
     },
     {
       "classification": "recurring",
-      "due": "{\"days\": 0}",
+      "due": "{\"days\": 15}",
       "expired": "{\"months\": 3, \"days\": -2}",
       "name": "IRONMAN_v3_recurring_3mo_pattern",
       "overdue": "{\"days\": 16}",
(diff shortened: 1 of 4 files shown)


portal/config/eproms/QuestionnaireBank.json
```diff
@@ -262,7 +262,7 @@
     },
     {
       "classification": "recurring",
-      "due": "{\"days\": 0}",
+      "due": "{\"days\": 15}",
```
mcjustin, line 265: Pretty fundamental change, it seems likely to have unintended consequences

achen2401: @pbugni Changes checked out on the UI, but just want to make sure, per Justin's comment, will the change here 
 (and elsewhere in the QuestionnaireBank.json) affect the global study?

pbugni: As discussed, this is very confusing as it is currently implemented.  The right thing to do is to address the internal state, as we use `due` and `start` as synonyms in some places and discrete values in others.

The change above alters the QuestionnaireBank definition for each respected case to have a different `due` date than before.  But those values are NOT used when looking at the `due` (better known as start or available) value for a timeline, assessment, etc.

I'm willing to pull together a refactor PR that addresses this, but it's a fair amount of work and change:
``` find portal -name \*.py| xargs grep due| grep -iv overdue| wc -l
121
```

mcjustin: Per our convo this morning, I'm content w/ this change

ivan-c (approved):

PR #3986, opened 2021-02-09
adherence report UI on patient list - export error display fix

achen2401:
Adherence report UI on patient list:
Issue:  Currently, the UI doesn't display error detail and dialog disappears promptly when error happens in the process of exporting.
Fix: display and _persist_ more discrete error detail from the ajax call to export adherence report data when it fails.


ivan-c (approved):

pbugni:
Thank you @achen2401

PR #3990, opened 2021-02-11
TN-2924 exclude weekends for EMPRO reminder schedules.

pbugni:
When calculating if it's time to send a reminder email to staff for EMPRO triggered patient follow ups, we need to exclude weekends.

Added new date_tools utility function `weekday_delta()` to calculate time delta between two dates, omitting weekend days.

### DIFF
--- a/portal/date_tools.py
+++ b/portal/date_tools.py
@@ -1,5 +1,5 @@
 """Module for additional datetime tools/utilities"""
-from datetime import date, datetime
+from datetime import date, datetime, timedelta
 import json
 
 from dateutil import parser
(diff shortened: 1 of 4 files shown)


portal/date_tools.py
```diff
@@ -172,3 +172,21 @@ def utcnow_sans_micro():
     """
     now = datetime.utcnow()
     return now.replace(microsecond=0)
+
+
+def weekday_delta(start, end):
```
ivan-c, line 177: Will this function mutate `start`? Is that a problem?

ivan-c (approved):
Looks good!

portal/date_tools.py
pbugni: reasonable concern, but i don't believe it's modifying `start` or `end` - that's the purpose of using `i` in this loop as the mutable form.

PR #3992, opened 2021-02-11
Add critical USER_UNAUTHORIZED_ENDPOINT and enable clinicians to generate adherence reports

pbugni:
Clinicians were blocked from accessing `/api/report/questionnaire_status` (an easy fix) but this was difficult to track as we didn't have a `USER_UNAUTHORIZED_ENDPOINT` configured, and would therefore simply redirect to '' and return a 301 status rather than a 401 as needed for the front end to be aware of the error situation.

### DIFF
--- a/portal/views/auth.py
+++ b/portal/views/auth.py
@@ -1025,3 +1025,21 @@ def authorize(*args, **kwargs):
         return redirect('/')
     # See "hardwired" note in docstring above
     return True
+
+
+@auth.route('/unauthorized_endpoint')
+@oauth.require_oauth()
+def unauthorized_endpoint():
+    """Redirection endpoint for logged-in users w/o adequate role/permission
+
+    Flask-User can be configured to point to a view such as this on an event
+    such as a user being logged in but lacking adequate role, say the view
+    is decorated with @roles_required().
+
+    By default, such a case redirects to '', and fails to return an adequate
+    status code, but rather a 302.
+
+    Most importantly, raise a 401 so clients can catch this scenario.
+
+    """
+    abort(401)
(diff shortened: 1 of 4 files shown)


achen2401 (approved):
Thanks, Paul, for the fix.  The changes look good to me - I didn't encounter error when exporting report data as a user with just the clinician role.  Also, I tested the error scenario (e.g., unauthorized use), that looks good as well as error is displayed on the frontend (see screenshot): 

Screen Shot 2021-02-11 at 2 39 34 PM

portal/views/portal.py
```diff
@@ -141,6 +141,8 @@ def report_slow_queries(response):
 
     """
     if current_app.config.get("LOG_SLOW_RESPONSES"):
+        if not hasattr(g, 'start_request_time'):
+            return response
```
ivan-c, line 145: This doesn't look related to the bugfix. If I understand correctly, this seems to exit early if the request is not being timed, but isn't that controlled by the above config?

portal/views/auth.py
```diff
@@ -1025,3 +1025,21 @@ def authorize(*args, **kwargs):
         return redirect('/')
     # See "hardwired" note in docstring above
     return True
+
+
+@auth.route('/unauthorized_endpoint')
+@oauth.require_oauth()
```
ivan-c, line 1031: Do we need this decorator for an endpoint that is effectively informational? Worried it'll cause a redirect loop if flask-user is sending unauthenticated users here

ivan-c (approved):
Looks good! Had some questions, but ready to merge

portal/views/portal.py
pbugni: it's just something that blew up on me when testing - never seen it before, but seemed worthy of adding a safeguard.

portal/views/auth.py
pbugni: that's a different config, USER_UNAUTHENTICATED_ENDPOINT.  i walked through the flask-user code and confirmed we only hit this when the user is authenticated (logged-in) but not authorized.

PR #3993, opened 2021-02-11
TN-2932 action_state should reflect 'withdrawn'

pbugni:
on discovery of a withdrawn EMPRO user, update the `action_state` to withdrawn as well, so the reports display the desired Clinician Status.


ivan-c (approved):
:star: :star:

PR #3997, opened 2021-02-16
Zero consented patients query safeguard

pbugni:
On a system with zero consented users, the query intended to include only consented users must consider the empty set.


ivan-c (approved):

achen2401 (approved):
thank you, Paul

PR #3998, opened 2021-02-17
Add `T` to datetime format string in failing migration on truenth-demo

pbugni:
Potential fix on truenth-demo where the `T` separating date & time in `document->>'authored'` (as specified by ISO 8601) is killing the SQL `TO_TIMESTAMP` parse.

### DIFF
--- a/portal/migrations/versions/68a25f790d27_.py
+++ b/portal/migrations/versions/68a25f790d27_.py
@@ -6,6 +6,7 @@ Create Date: 2020-12-08 14:17:28.279106
 
 """
 from alembic import op
+from flask import current_app
 import sqlalchemy as sa
 
 
@@ -15,6 +16,8 @@ down_revision = '7fd6b3abfec2'
 
 
 def upgrade():
+    if current_app.config.get('GIL'):
+        return
     op.add_column(
         'trigger_states',
         sa.Column('visit_month', sa.Integer()))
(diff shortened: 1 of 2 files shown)


portal/migrations/versions/68a25f790d27_.py
```diff
@@ -15,6 +16,8 @@
 
 
 def upgrade():
+    if current_app.config.get('GIL'):
```
ivan-c, line 19: I think `PERSISTENCE_DIR` is the top-level config that sets `GIL`, but don't we need this to run on demo TNUSA, where `PERSISTENCE_DIR` is configured as `gil`?

ivan-c (changes requested):
Minor config issue, looks good otherwise!

portal/migrations/versions/68a25f790d27_.py
pbugni: We happen to set both, verified in `https://demo.us.truenth.org/settings`.  I've just used this one in a number of like migrations.

ivan-c (approved):

portal/migrations/versions/68a25f790d27_.py
ivan-c: Sorry, misunderstood that was for the triggers table, not the authored datetime fixup

PR #3999, opened 2021-02-17
TN-2933 call the correct research study api on a staff profile page

achen2401:
Address story:  https://jira.movember.com/browse/TN-2933

- Issue:  An ajax call is fired off to the **patient** research study API on a **staff** profile page, resulting in a 400 error, i.e. incorrect role.
- Background for the ajax api call:  The call to the research study api is mainly relevant in the context of the patient profile page where certain UI elements can only be seen by the EMPRO patients.
- Fix:  Correctly call the research study API for staff on a staff profile page if it is determined that the subject is not a patient.

Implemented as a hot fix.

### DIFF
--- a/portal/static/js/src/profile.js
+++ b/portal/static/js/src/profile.js
@@ -570,7 +570,8 @@ export default (function() {
                 }
             },
             setSubjectResearchStudies: function() {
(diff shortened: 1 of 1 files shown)


pbugni (approved):
👍

ivan-c (approved):

PR #4006, opened 2021-03-09
TN-2927 multiple clinicians for EMPRO patients

pbugni:
- Patients can now have any number of clinicians.
- `primary_investigator` (PI) role added
- PI added as clinician on EMPRO consent


portal/models/qb_status.py
```diff
@@ -523,7 +523,7 @@ def patient_research_study_status(patient, ignore_QB_status=False):
             'errors': [],
         }
         results[rs] = rs_status
-        if rs == EMPRO_RS_ID and patient.clinician_id is None:
+        if rs == EMPRO_RS_ID and len(patient.clinicians) == 0:
```
ivan-c, line 526: If these are SQLA models, should they use the `.count()` method?
```suggestion
        if rs == EMPRO_RS_ID and patient.clinicians.count() == 0:
```

portal/models/user.py
```diff
@@ -1174,6 +1179,30 @@ def update_consents(self, consent_list, acting_user):
...
+        self.check_consents()
+
+    def check_consents(self):
+        """Hook method for application of consent related rules"""
+
+        # For EMPRO, automatically add the PI on consent
+        from .user_consent import latest_consent
+        from .research_study import EMPRO_RS_ID
+        consent = latest_consent(self, EMPRO_RS_ID)
+        if consent and len(self.clinicians) == 0:
```
ivan-c, line 1191:
```suggestion
        if consent and self.clinicians.count() == 0:
```

tests/test_demographics.py
```diff
@@ -334,6 +342,31 @@ def test_demographics_duplicate_ref(self):
...
+
+        just_the_pi = {
+            "resourceType": "Patient",
+            "careProvider": [
+                {"reference": "Clinician/{}".format(pi_id)}]
+            }
+
+        response = self.client.put(
+            '/api/demographics/%s' % TEST_USER_ID,
+            content_type='application/json', data=json.dumps(just_the_pi))
```
ivan-c, line 363:
```suggestion
            '/api/demographics/%s' % TEST_USER_ID, json=just_the_pi)
```

ivan-c (approved):
Looks good! Had a question and nitpick, but ready to merge

portal/models/qb_status.py
pbugni: it depends on the relationship type.  "dynamic" builds a query, on which you can call `count()` - otherwise the relationship manages the attribute as a python iterable, for which `count()` isn't defined.

PR #4023, opened 2021-03-24
Patch orgtree threading issue

pbugni:
Adding a redis lock around the OrgTree initializer eliminated the notorious and illusive mysterious situation where organizations in the tree would lose their parents.

During one of those rare windows when the symptom of this issue (would raise as `OrgTree.at_and_above_ids()` would find a node that had lost its parent) was reliably reproducible unless flask was run `--without-threads`, opted to try a redis lock around the initializer and found it was no longer reproducible.  The reproduction would happen when viewing /patients/substudy as a JS request starts a new thread that somehow exposed an internal shared resource that is apparently not thread safe.  Probably a SQLAlchemy pool or the like.

I should note we haven't seen this on a test or prod system, only in development (both @achen2401 and I have had windows of time where it would come and go).

### DIFF
--- a/portal/models/organization.py
+++ b/portal/models/organization.py
@@ -15,6 +15,7 @@ from ..database import db
 from ..date_tools import FHIR_datetime
 from ..dict_tools import strip_empties
 from ..system_uri import IETF_LANGUAGE_TAG, SHORTNAME_ID, TRUENTH_RP_EXTENSION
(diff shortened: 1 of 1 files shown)


achen2401 (approved):

ivan-c (approved):
:lock: :thread:

PR #4024, opened 2021-03-24
TN-2955 EMPRO adherence report: add triggers and content access

pbugni:


requirements.txt
```diff
@@ -35,6 +35,7 @@ healthcheck==1.3.3
 idna==2.10                 # via requests
 itsdangerous==1.1.0        # via flask
 jinja2==2.11.2              # via flask, flask-babel, sphinx
+jmespath==0.10.0
```
ivan-c, line 38: not exactly eager to add another dependency for such a shallow lookup, unless you think it'll be more widely useful elsewhere

ivan-c (approved):
Had some reservations about adding another dependency, but no blockers

requirements.txt
pbugni: fair point.  it would be useful elsewhere, and significantly reduces the robust error checking and defaults at each level.  i was surprised to find it wasn't already available.  so, i'd like to have it available, but will remove if you feel strongly?  as you probably know, it's the same JSON Path solution used on other projects like map-api.

ivan-c: I feel pretty strongly that it's not worth the maintenance burden (managing as a dependency, reading changelogs etc) if it can be replaced by 2 chained `get()` dictionary methods. If you're confident it'll be used it more widely, please keep it.

No hard feelings either way- I only have a stronger opinion now that I've had to dig through our dependabot PRs :)

mcjustin (approved):

PR #4026, opened 2021-03-24
TN-2963 use study ID in EMPRO trigger emails to staff.

pbugni:
NB, will only function once liferay templates are updated to use {study_id} - see https://jira.movember.com/browse/TN-2968


achen2401 (commented):
@pbugni We will need to make sure to inform Cam to use the new defined variable, `study_id` in the LF email content, i.e. trigger email content to staff.
I think he is currently using the variable, `patient_id` in the content, see attached:
Screen Shot 2021-03-24 at 4 19 52 PM

PR #4035, opened 2021-03-31
TN-2854, TN-2955, TN-2972: Adherence reports

pbugni:
Number of fixes, enhancements on adherence reports.

- Withdrawn users now include consent dates
- EMPRO columns added for triggers and access (TN-2854)
- Timewarp adjusts audits tracking access (TN-2972)

### DIFF
--- a/portal/models/qbd.py
+++ b/portal/models/qbd.py
@@ -107,15 +107,17 @@ class QBD(object):
         if not query.count():
             # Check indefinite case, which doesn't generate timeline rows
             if self._questionnaire_bank.classification == 'indefinite':
-                query = QuestionnaireResponse.query.filter(
+                found = QuestionnaireResponse.query.filter(
                     QuestionnaireResponse.subject_id == user_id).filter(
                     QuestionnaireResponse.status == 'completed').join(
                     QuestionnaireBank).filter(
-                    QuestionnaireResponse.questionnaire_bank_id == QuestionnaireBank.id).filter(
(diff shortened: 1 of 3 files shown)


portal/models/qbd.py
```diff
@@ -103,5 +105,16 @@ def completed_date(self, user_id):
...
+            if self._questionnaire_bank.classification == 'indefinite':
+                query = QuestionnaireResponse.query.filter(
+                    QuestionnaireResponse.subject_id == user_id).filter(
+                    QuestionnaireResponse.status == 'completed').join(
+                    QuestionnaireBank).filter(
+                    QuestionnaireResponse.questionnaire_bank_id == QuestionnaireBank.id).filter(
+                    QuestionnaireBank.classification == 'indefinite').with_entities(
+                    QuestionnaireResponse.document['authored'].label('authored'))
+                if query.count():
+                    return FHIR_datetime.parse(query.first()[0])
```
ivan-c, line 118: I think using both `query.count()` and `query.first()` will cause 2 queries to be run vs a single one if only `first()` is used.
```suggestion
                first = query.first()
                if first:
                    return FHIR_datetime.parse(first[0])
```

ivan-c (approved):
Minor speculative nitpick; looks good to merge!

portal/models/qbd.py
pbugni: good point!

mcjustin (approved):

PR #4039, opened 2021-04-01
TN-2973 Configure app and celery workers to use JSON logging

pbugni:


ivan-c:
@pbugni this should be ready to re-review/merge whenever you get a chance. thanks again!

PR #4049, opened 2021-04-12
TN-2974 2FA for privledged accounts

pbugni:


achen2401 (approved):

portal/views/auth.py
```diff
@@ -69,6 +79,38 @@
...
+    except (BadRequest, ValueError):
+        abort(404)
+
+    form = KeyForm()
+    if not form.validate_on_submit():
+        return render_template('flask_user/second_factor.html', form=form)
+
+    # Still here implies validation was successful, resume interrupted login
+    session.pop('user_needing_2fa')
+    session['2FA_verified'] = '2FA verified'
```
pbugni, line 109: perhaps as an extra layer of paranoia, this should be the user_id that passed 2FA?

portal/views/auth.py
```diff
@@ -360,7 +402,7 @@ def base64_url_decode(s):
         """url safe base64 decoding method"""
         padding_factor = (4 - len(s) % 4)
         s += "=" * padding_factor
-        return base64.b64decode(unicode(s).translate(
+        return base64.b64decode(s.translate(
```
pbugni, line 405: not related to this PR - just a broken line of code left over from the python2 days.

portal/models/login.py
```diff
@@ -20,6 +23,39 @@ def login_user(user, auth_method=None):
...
+        # log user back out, in case a flow already promoted them
+        flask_user_logout()
+
+        token = user.generate_otp()
+        session['user_needing_2fa'] = user.id
+        session['pending_auth_method'] = auth_method
+        current_app.logger.debug(f"OTP for {user.id}: {token}")
+        email = EmailMessage(
+            subject="TrueNTH Access Token",
+            body=f"One Time Authentication Token: {token}",
```
ivan-c, line 50: I think we need i18n for staff, but I'm not 100% sure. @mcjustin ?

portal/models/user.py
```diff
@@ -304,6 +311,9 @@ class User(db.Model, UserMixin):
         db.Column(db.Integer, default=0, nullable=False)
     last_password_verification_failure = db.Column(db.DateTime, nullable=True)
 
+    # For 2FA
+    otp_secret = db.Column(db.String(16), default=generate_random_secret)
```
ivan-c, line 315: Naive question: why not use the app's `SECRET_KEY` instead of generating a secret for every user?...

portal/models/user.py
```diff
@@ -469,6 +479,28 @@ def current_encounter(
             return initiate_encounter(self, auth_method=existing.auth_method)
         return existing
 
+    TOKEN_LEN = 4
```
ivan-c, line 482: Prefer a more unique name, just in case we need to make this configurable...
```suggestion
    TOTP_TOKEN_LEN = 4
```

portal/models/user.py
```diff
@@ -469,6 +479,28 @@ def current_encounter(
...
+    TOKEN_LIFE = 30*60
```
ivan-c, line 483:
```suggestion
    TOTP_TOKEN_LIFE = 30*60
```

portal/templates/flask_user/second_factor.html
```diff
@@ -0,0 +1,16 @@
+{% extends "layout.html" %}
+{% block main %}
+{% from "flask_user/_macros.html" import render_field %}
+
+    <h3 class="tnth-headline">{{ _("Two Factor Athentication") }}</h3><br />
+
+    <p>{{ _("You should receive an email with a token.  Enter below for access.") }}</p>
```
ivan-c, line 7: Just a little more explicit (acknowledge Mo may have more language tweaks)
```suggestion
    <p>{{ _("You should receive an email with a token. Enter the token below for access.") }}</p>
```

portal/templates/flask_user/second_factor.html
```diff
@@ -0,0 +1,16 @@
...
+
+    <form action="/2fa/verify" method="post">
```
ivan-c, line 9:
```suggestion
    <form action="{{ url_for('two_factor_auth') }}" method="post">
```

portal/views/auth.py
```diff
@@ -69,6 +79,38 @@
...
+
+    def validate_key(form, field):
+        """Use form validation to verify token"""
+        user = unchecked_get_user(session.get('user_needing_2fa'))
+
+        if user.validate_otp(field.data):
+            current_app.logger.debug(f"{user.id} passed 2FA with valid token")
+        else:
+            current_app.logger.debug(f"{user.id} failed 2FA token validation")
+            raise ValidationError("invalid token")
```
ivan-c, line 93:
```suggestion
            raise ValidationError("invalid 2FA token")
```

requirements.txt
```diff
@@ -40,6 +40,7 @@ kombu==4.6.6              # pyup: < 5.0 # via celery
 mako==1.1.0               # via alembic
 markupsafe==1.1.1           # via jinja2, mako
 oauthlib==2.1.0 # pyup: < 3.0.0  # pin until flask-oauthlib, flask-dance, requests-oauthlib upgraded
+onetimepass==1.0.1
```
ivan-c, line 43: Please add to `setup.cfg` as well

ivan-c (approved):
Looks great! Seems like a good compromise working around flask-user. Had a few nitpicks but no blockers

portal/models/login.py
mcjustin: Yes, we'll need i18n here. Current emails to staff are indeed i18n'd (as is the content on the website for them).

portal/models/user.py
pbugni: security.  without this, you could potentially use a key obtained from one account on a second, provided you hit the time window.

ivan-c: Ah, thanks for the explanation- I assumed the tokens had the user_id/email inside the token.

portal/models/login.py
```diff
@@ -20,6 +24,39 @@ def login_user(user, auth_method=None):
...
+        # log user back out, in case a flow already promoted them
+        flask_user_logout()
+
+        token = user.generate_otp()
+        session['user_needing_2fa'] = user.id
+        session['pending_auth_method'] = auth_method
+        current_app.logger.debug(f"OTP for {user.id}: {token}")
+        email = EmailMessage(
+            subject=_("TrueNTH Access Token"),
+            body=_(f"One Time Authentication Token: {token}"),
```
ivan-c, line 51: @pbugni Just FYI this is causing a broken build
```suggestion
            body=_("One Time Authentication Token: %s", token),
```

portal/config/config.py
```diff
@@ -103,6 +103,7 @@ class BaseConfig(object):
         ('assessment_cache_region', 60*60*2),
         ('reporting_cache_region', 60*60*12)]
     SEND_FILE_MAX_AGE_DEFAULT = 60 * 60  # 1 hour, in seconds
+    ENABLE_2FA = os.environ.get('ENABLE_2FA', None)
```
pbugni, line 106: @ivan-c is there a docker file in which we need to mention this var?  shouldn't be on anywhere but -test for now...

ivan-c: We have a [`portal.env` "template"](https://github.com/uwcirg/truenth-portal/blob/develop/docker/portal.env.default) we could add it to, but I'm not entirely convinced we should add every value, effectively documenting them twice.

I might say this is important enough to do that for though.  :man_shrugging:

pbugni: agreed - no need for that level of detail.  thanks

PR #4050, opened 2021-04-12
JSON Logging Fixups

ivan-c:
* Use the same module-global variable as `json_logging` ([`_current_framework`](https://github.com/bobbui/json-logging-python/blob/master/json_logging/__init__.py#L29)) to detect logging initialization
* Read celery loglevel from flask app, or `celery` CLI option

### DIFF
--- a/portal/factories/app.py
+++ b/portal/factories/app.py
@@ -246,7 +246,7 @@ def configure_logging(app):  # pragma: no cover
         return
 
     # Configure for JSON logging
-    if json_logging._request_util is None:
+    if json_logging._current_framework is None:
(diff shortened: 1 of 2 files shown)


portal/factories/app.py
```diff
@@ -246,7 +246,7 @@ def configure_logging(app):  # pragma: no cover
         return
 
     # Configure for JSON logging
-    if json_logging._request_util is None:
+    if json_logging._current_framework is None:
```
pbugni, line 249: better choice. thx

pbugni (approved):
well done.  can't wait to see how it works out w/ elastic.

PR #4051, opened 2021-04-12
Update pytest

ivan-c:
Update pytest to latest working version. 

NB newer versions of py.test seem to be more prone to timing out ([see CI failure rate](https://github.com/uwcirg/truenth-portal/commits/fixup/update-pytest_stepwise)). If this continues to be a problem, we should revert this change.


pbugni (approved):
Let's give it a go!

PR #4055, opened 2021-04-15
TN-2989 correctly manage 'next' in 2FA

pbugni:
2FA was being thwarted by unauthenticated request to restricted resources.  Difficult to wire in with flask-user, but found a way by patching the `make_safe_url()` function of flask-user.

Replaced the `token` term with `code` as suggested in parent story TN-2974

### DIFF
--- a/portal/models/login.py
+++ b/portal/models/login.py
@@ -49,13 +49,13 @@ def login_user(user, auth_method=None):
         current_app.logger.debug(f"2FA OTP for {user.id}: {code}")
         email = EmailMessage(
             subject=_("TrueNTH Access Code"),
-            body=_("One Time Authentication Code: %(code)s", code=code),
+            body=_("One-time access code: %(code)06d", code=code),
             recipients=user.email,
             sender=current_app.config['MAIL_DEFAULT_SENDER'],
             user_id=user.id)
         if current_app.config.get("MAIL_SUPPRESS_SEND"):
             # Dump to console for easy access
-            print(f"!!!!!!!! 2FA Code: {code} !!!!!!!!")
+            print(email.body)
         else:
             email.send_message()
 
--- a/portal/models/login.py
+++ b/portal/models/login.py
@@ -43,17 +43,22 @@ def login_user(user, auth_method=None):
         # log user back out, in case a flow already promoted them
         flask_user_logout()
 
-        token = user.generate_otp()
+        code = user.generate_otp()
         session['user_needing_2fa'] = user.id
         session['pending_auth_method'] = auth_method
-        current_app.logger.debug(f"OTP for {user.id}: {token}")
+        current_app.logger.debug(f"2FA OTP for {user.id}: {code}")
         email = EmailMessage(
-            subject=_("TrueNTH Access Token"),
-            body=_("One Time Authentication Token: %s" % token),
+            subject=_("TrueNTH Access Code"),
+            body=_("One Time Authentication Code: %s" % code),
             recipients=user.email,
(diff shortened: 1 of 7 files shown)


portal/models/login.py
```diff
@@ -43,17 +43,22 @@ def login_user(user, auth_method=None):
...
+        code = user.generate_otp()
         session['user_needing_2fa'] = user.id
         session['pending_auth_method'] = auth_method
-        current_app.logger.debug(f"OTP for {user.id}: {token}")
+        current_app.logger.debug(f"2FA OTP for {user.id}: {code}")
         email = EmailMessage(
-            subject=_("TrueNTH Access Token"),
-            body=_("One Time Authentication Token: %s" % token),
+            subject=_("TrueNTH Access Code"),
+            body=_("One Time Authentication Code: %s" % code),
```
ivan-c, line 52: I don't think this is lazily evaluated, meaning babel won't be able to extract it...

```suggestion
            body=_("One Time Authentication Code: %(code)s", code=code),
```

portal/models/login.py
```diff
@@ -43,17 +43,22 @@ def login_user(user, auth_method=None):
...
             recipients=user.email,
             sender=current_app.config['MAIL_DEFAULT_SENDER'],
             user_id=user.id)
-        email.send_message()
+        if current_app.config.get("DEV_SYSTEM_WO_EMAIL", None):
```
ivan-c, line 56: I think we could reuse an existing config value here...
```suggestion
        if current_app.config.get("MAIL_SUPPRESS_SEND"):
```

ivan-c (changes requested):
One minor i18n blocker; ready to merge otherwise

portal/templates/flask_user/second_factor.html
```diff
@@ -1,16 +1,22 @@
...
-
-    <form action="{{ url_for('auth.two_factor_auth') }}" method="post">
-        {{ form.hidden_tag() }}
-        {{ render_field(form.key, label=_("Token"), tabindex=5) }}
-
-        <button type="submit" class="btn btn-tnth-primary">{{ _("Validate Token") }}</button>
-    </form>
-
+    <h3 class="tnth-headline">{{ _("Two Factor Athentication") }}</h3>
+    <p>{{ _("You should receive an email with a 4 digit code. Enter the code below for access.") }}</p>
```
ivan-c, line 5: Just in case we flip-flop a few more times on the number of characters :) 
```suggestion
    <p>{{ _("You should receive an email with a %d digit code. Enter the code below for access.", 4) }}</p>
```

pbugni: agreed.  looks like this already got fixed?

portal/templates/flask_user/second_factor.html
```diff
@@ -1,16 +1,22 @@
...
-
-    <form action="{{ url_for('auth.two_factor_auth') }}" method="post">
-        {{ form.hidden_tag() }}
-        {{ render_field(form.key, label=_("Token"), tabindex=5) }}
-
-        <button type="submit" class="btn btn-tnth-primary">{{ _("Validate Token") }}</button>
-    </form>
-
+    <h3 class="tnth-headline">{{ _("Two Factor Athentication") }}</h3>
+    <p>{{ _("You should receive an email with a %(code)d digit code. Enter the code below for access.", code=6) }}</p>
```
ivan-c, line 5:
```suggestion
    <p>{{ _("You should receive an email with a %(length)d digit code. Enter the code below for access.", length=6) }}</p>
```

ivan-c (approved):

PR #4061, opened 2021-04-22
TN-2994 correct QB visit date overlaps

pbugni:
Eliminate overlapping QB problems by shifting the expiration date of the previous QB to be just before the start of the next QB.

- Click task `find-overlaps()` will seek out and report all overlapping QBs in the active patients, and optionally correct.
- Any overlaps found while building up a user's timeline are eliminated by moving the expiration date, provided QNRs or other events won't be clobbered.
- Additional safeguards added to the timeline build process, to confirm we don't insert new rows containing overlaps.

### DIFF
--- a/manage.py
+++ b/manage.py
@@ -736,8 +736,9 @@ def preview_site_update(org_id, retired):
 
     # Update the org's research protocol as requested - assume to the latest
     previous_rp = organization.research_protocols[-1]
-    latest_rp = ResearchProtocol.query.order_by(
-        ResearchProtocol.id.desc()).first()
+    assert previous_rp.name == 'IRONMAN v3'
+    latest_rp = ResearchProtocol.query.filter(
+        ResearchProtocol.name == 'IRONMAN v5').one()
     previous_org_rp = OrganizationResearchProtocol.query.filter(
         OrganizationResearchProtocol.research_protocol_id ==
         previous_rp.id).filter(
--- a/manage.py
+++ b/manage.py
@@ -34,7 +34,11 @@ from portal.models.i18n_utils import (
 )
 from portal.models.intervention import add_static_interventions
 from portal.models.organization import add_static_organization
-from portal.models.qb_timeline import invalidate_users_QBT
+from portal.models.qb_timeline import (
+    QBT,
+    invalidate_users_QBT,
+    update_users_QBT,
+)
 from portal.models.questionnaire_bank import (
     QuestionnaireBank,
     add_static_questionnaire_bank,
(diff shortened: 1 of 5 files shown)


portal/factories/app.py
```diff
@@ -260,8 +260,13 @@ def configure_logging(app):  # pragma: no cover
...
 
     # Configure Error Emails for high level log messages, only in prod mode
+    # and with email enabled
+
     if not any((
-            app.debug, app.testing, not app.config.get('ERROR_SENDTO_EMAIL'))):
+            app.debug,
+            app.testing,
+            not app.config.get('ERROR_SENDTO_EMAIL'),
+            app.config.get('MAIL_SUPPRESS_SEND'))):
```
ivan-c, line 269: I think this check will disable exception emails if `MAIL_SUPPRESS_SEND` is enabled, which might be useful in the staging environment. 

Not a huge deal either way if we're going to work towards generalizing exception emails with ELK.

portal/models/qb_timeline.py
```diff
@@ -736,6 +762,77 @@ def invalidate_users_QBT(user_id, research_study_id):
...
+    for row in qbt_rows:
+        # Confirm expected order
+        if last_at:
+            assert row.at >= last_at
+
+        key = f"{row.qb_id}:{row.qb_iteration}"
+        if previous_key and previous_key != key:
+            # just moved to next visit, confirm it's novel
+            if key in seen:
+                if not (key in reported_on and previous_key in reported_on):
```
ivan-c, line 796: Minor issue-some of these could be inverted into guard clauses (with `continue`) to make it less indented

ivan-c (approved):
Looks good! Had a few questions but no blockers

manage.py
```diff
@@ -604,6 +608,91 @@ def merge_users(src_id, tgt_id, actor):
...
 
 
+def capture_patient_state(patient_id):
+    """Call to capture QBT and QNR state for patient, used for before/after"""
+    qnrs = QuestionnaireResponse.qnr_state(patient_id)
+    tl = QBT.timeline_state(patient_id)
+    return {'qnrs': qnrs, 'timeline': tl}
+
+
+def present_before_after_state(user_id, external_study_id, before_state):
```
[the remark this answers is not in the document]
pbugni: I thought about that - thus far, only used here.  If scope grows, I'll refactor as suggested.

manage.py
```diff
@@ -604,6 +608,91 @@ def merge_users(src_id, tgt_id, actor):
...
+                qbt_rows, cli_presentation=True) and correct_overlaps:
+            # Reprocess w/ adjusting expired, report differences
+            b4 = capture_patient_state(patient.id)
+            if reprocess_qnrs:
+                # Extends runtime and makes for noisy audit logs.
+                # Furthermore in practice no QNRs require updates as
+                # expiration isn't moved if QNRs were posted in the overlap.
+                QuestionnaireResponse.purge_qb_relationship(
+                    subject_id=patient.id,
+                    research_study_id=0,
```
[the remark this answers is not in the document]
pbugni: EMPRO is (so far) free from all of this sort of confusion.  We'll need a significant re-factoring should that come into play (EMPRO research protocols also introducing overlapping issues on change).

portal/models/qb_timeline.py
pbugni: not easy, given the variables set following that clause.  punting for now, but thanks for the suggestion, i like your eye toward better!

mcjustin (approved):

PR #4067, opened 2021-04-26
TN-2998 allow timewarp on demo systems.

pbugni:
Was a hotfix, reverted to standard flow as time allows for this to go out with the current sprint.

### DIFF
--- a/portal/views/patient.py
+++ b/portal/views/patient.py
@@ -472,7 +472,8 @@ def patient_timewarp(patient_id, days):
     from portal.models.questionnaire_response import QuestionnaireResponse
     from portal.models.user_consent import UserConsent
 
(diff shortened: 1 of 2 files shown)


portal/views/patient.py
```diff
@@ -472,7 +472,8 @@ def patient_timewarp(patient_id, days):
     from portal.models.questionnaire_response import QuestionnaireResponse
     from portal.models.user_consent import UserConsent
 
-    if current_app.config['SYSTEM_TYPE'] not in ('development', 'testing'):
+    if current_app.config['SYSTEM_TYPE'] not in (
+            'demo', 'development', 'testing'):
```
ivan-c, line 476: Why not just "not production"?
```suggestion
    if current_app.config['SYSTEM_TYPE'] != "production":
```

ivan-c (approved):
Looks good!

PR #4076, opened 2021-05-04
TN-3007 Tolerate redundant identifier sets on same user

pbugni:


tests/test_identifier.py
```diff
@@ -41,6 +41,15 @@ def testPOST(test_user, login, client):
     user = User.query.get(TEST_USER_ID)
     assert len(user.identifiers) == expected
 
+    # Insert the same, expect idempotent reaction
+    response = client.post(
+        '/api/user/{}/identifier'.format(TEST_USER_ID),
+        content_type='application/json', data=json.dumps(data))
+    assert response.status_code == 200
```
ivan-c, line 48: I've come to acknowledge there are some uses for f-strings :wink: 
```suggestion
        f'/api/user/{TEST_USER_ID}/identifier',
        json=data)
    assert response.status_code == 200
```

ivan-c (approved):
Looks good! Minor testing nitpick but ready to merge

tests/test_identifier.py
pbugni: i love it!  but this is one of thousands of test code strings that could use such an enhancement - just caught in the c&p from the test code above.
let's punt for now - don't see it worth delaying for another build cycle.

PR #4078, opened 2021-05-06
TN-3008 hide deleted user identifiers

pbugni:
To avoid any potential conflicts, should a user be deleted with an identifier wanting use elsewhere, rename the identifier with a `-deleted` suffix.

Includes migration to clean up existing case and code to do so moving forward.


portal/migrations/versions/ebb5fee8122b_.py
```diff
@@ -0,0 +1,55 @@
...
+        "SELECT user_id FROM user_identifiers JOIN users ON users.id = user_id "
+        "WHERE deleted_id IS NOT NULL").fetchall()
+
+    fixes = []
+    for user_id in patients:
+        ident_id, value = conn.execute(text(
+            "SELECT identifiers.id, value FROM identifiers JOIN "
+            "user_identifiers ON identifiers.id = identifier_id WHERE "
+            "user_id=:user_id"), user_id=user_id[0]).fetchone()
+        # Only concerned with IDs that match the MR format (nnn-nn-nnn)
```
ivan-c, line 37: Always appreciate a real example alongside a regex!

ivan-c (approved):
Looks great!

PR #4107, opened 2021-06-14
TN-3039 ironman i18n add Canadian English coding en_CA

mcjustin:
https://jira.movember.com/browse/TN-3039


ivan-c (changes requested):
Can you change the target branch to `master` so we can hotfix this onto demos if need be?

ivan-c (approved):

PR #4121, opened 2021-07-13
TN-3027 Adherence report entry method work

pbugni:
- restore "online" default to entry method (when not paper, interview assisted or staff authenticated)
- add entry method `staff_authenticated` for the login as patient scenario
- calculate all the entry methods for all questionnaire responses on a given visit and when more than one exists, return the one most frequently used and a ratio of that one to all, i.e. `paper 4:5`

### DIFF
--- a/portal/models/questionnaire_response.py
+++ b/portal/models/questionnaire_response.py
@@ -639,9 +639,11 @@ class QNR_results(object):
         # return only one, namely the most frequent
         # if not all were the same as most frequent, include ratio
         max_key = max(found, key=lambda key: found[key])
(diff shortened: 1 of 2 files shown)


mcjustin:
@pbugni great - I've just asked Theresa about including the ratio in the report.

mcjustin:
@pbugni  per email today, Theresa doesn't want the ratio included in the report.

mcjustin (changes requested):
@pbugni per email today, Theresa doesn't want the ratio included in the report.

mcjustin (approved):

PR #4126, opened 2021-07-20
TN-3048 /api/patient/<>/research_study extended for `staff_eligible`

pbugni:
Front end was erroneously disabling the control for a staff follow up on EMPRO trigger, once the patient has subsequent global study work pending.

The `/api/patient/<>/research_study` API now includes an additional `staff_eligible` boolean.  Only when that is false does the control for staff follow up need to be disabled.

### DIFF
--- a/portal/models/qb_status.py
+++ b/portal/models/qb_status.py
@@ -510,7 +510,7 @@ def patient_research_study_status(patient, ignore_QB_status=False):
          have been met, such as assigned clinician if applicable.
      - ready: set True or False based on complex rules.  True means pending
          work user can immediately do.  NOT determined w/ ``ignore_QB_status``
-     - staff_eligible: if false, staff controls should be disabled.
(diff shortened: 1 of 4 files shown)


portal/models/qb_status.py
```diff
@@ -510,6 +510,8 @@ def patient_research_study_status(patient, ignore_QB_status=False):
          have been met, such as assigned clinician if applicable.
      - ready: set True or False based on complex rules.  True means pending
          work user can immediately do.  NOT determined w/ ``ignore_QB_status``
+     - staff_eligible: if false, staff controls should be disabled.
```
mcjustin, line 513: @pbugni  my only concern is the name of this variable and others in this PR... this is not limited to the staff role. It is for any role that can complete the intervention, which is I believe staff, staff_admin, and clinician. Maybe use `intervention_qnr_available`?

pbugni: 👍

mcjustin (approved):

PR #4130, opened 2021-08-04
TN-3049 capture successful 2FA auth

pbugni:


portal/views/auth.py
```diff
@@ -90,6 +90,9 @@ def validate_key(form, field):
 
         if user.validate_otp(field.data):
             current_app.logger.debug(f"{user.id} passed 2FA with valid token")
+            auditable_event(
+                message="Successful 2FA verification", context="login",
+                user_id=user.id, subject_id=user.id)
         else:
             current_app.logger.debug(f"{user.id} failed 2FA token validation")
             raise ValidationError("Invalid access code")
```
mcjustin, line 98: @pbugni  I wonder if we want to add entries for failures too - I find the identity verification failure entries very helpful. Granted, this is a simpler challenge, but people might try to use old codes. Defer to you, skip if you think it would be noise.

pbugni: very reasonable - we can handle a little noise, in the interest of making the system easier to support!

mcjustin (approved):

PR #4131, opened 2021-08-09
TN-3060 add ironman org William Jennings Bryan Dorn VAMC

mcjustin:
https://jira.movember.com/browse/TN-3060

### DIFF
--- a/portal/config/eproms/Organization.json
+++ b/portal/config/eproms/Organization.json
@@ -2289,6 +2289,34 @@
     {
       "extension": [
         {
+          "timezone": "America/New_York",
+          "url": "http://hl7.org/fhir/StructureDefinition/user-timezone"
+        },
+        {
+          "research_protocols": [
+            {"name": "IRONMAN v3"}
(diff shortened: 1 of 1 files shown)


portal/config/eproms/Organization.json
```diff
@@ -2286,6 +2286,34 @@
...
     },
+    {
+      "extension": [
+        {
+          "timezone": "America/New_York",
+          "url": "http://hl7.org/fhir/StructureDefinition/user-timezone"
+        },
+        {
+          "research_protocols": [
+            {"name": "IRONMAN v3"}
```
pbugni, line 2297: From the linked jira issue:
`Protocol Version: 4`
Beyond my love for magic numbers not lining up, doesn't 4 mean `IRONMAN v5` in this case?

mcjustin: No... there were no changes necessary on our end between v3 and v4, so from our perspective they are equivalent.

PR #4136, opened 2021-08-18
TN-3061 EMPRO training material

achen2401:
Address: https://jira.movember.com/browse/TN-3061
EMPRO training material

changes include
- add new endpoint for accessing the EMPRO training material
- a new template for the training material content, which is fetched from LR

### DIFF
--- a/portal/eproms/views.py
+++ b/portal/eproms/views.py
@@ -402,7 +402,7 @@ def resources():
 
 
 @eproms.route('/empro-resources', methods=['GET'])
-@roles_required([ROLE.STAFF.value, ROLE.STAFF_ADMIN.value])
+@roles_required([ROLE.STAFF.value, ROLE.STAFF_ADMIN.value, ROLE.CLINICIAN.value])
 @oauth.require_oauth()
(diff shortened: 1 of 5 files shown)


portal/eproms/views.py
```diff
@@ -401,6 +401,26 @@ def resources():
...
+    if not org:
+        abort(400, 'user must belong to an organization')
+    resources_data = get_any_tag_data("empro-training-material")
+    results = resources_data['results']
+    if len(results) > 0:
+        for item in results:
+            content = asset_by_uuid(item['uuid'])
+            item['content'] = content
+    else:
+        abort(400, 'resources not found')
```
ivan-c, line 419: You could pull the `resources not found` case into a guard clause this way:
```suggestion
    if len(results) == 0:
        abort(400, 'resources not found')
        
    for item in results:
        content = asset_by_uuid(item['uuid'])
        item['content'] = content
```

ivan-c (approved):
Looks great! One minor nitpick but ready to merge

pbugni (approved):

PR #4137, opened 2021-08-25
TN-3066 switch email to TSL on port 587

pbugni:
alter mail transport to pickup DKIM sig and hopefully reduce spam score.

### DIFF
--- a/portal/config/config.py
+++ b/portal/config/config.py
@@ -83,6 +83,7 @@ class BaseConfig(object):
     MAIL_USERNAME = os.environ.get('MAIL_USERNAME')
     MAIL_PASSWORD = os.environ.get('MAIL_PASSWORD')
     MAIL_USE_SSL = os.environ.get('MAIL_USE_SSL')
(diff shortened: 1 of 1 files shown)


portal/config/config.py
```diff
@@ -83,6 +83,7 @@ class BaseConfig(object):
     MAIL_USERNAME = os.environ.get('MAIL_USERNAME')
     MAIL_PASSWORD = os.environ.get('MAIL_PASSWORD')
     MAIL_USE_SSL = os.environ.get('MAIL_USE_SSL')
+    MAIL_USE_TLS = os.environ.get('MAIL_USE_TLS')
```
ivan-c, line 86: These might need some boolean casting?
```suggestion
    MAIL_USE_SSL = os.environ.get('MAIL_USE_SSL', 'false').lower == "true"
    MAIL_USE_TLS = os.environ.get('MAIL_USE_TLS', 'false').lower == "true"
```

ivan-c (approved):
Had a question; looks good otherwise

PR #4144, opened 2021-10-05
TN-3077 2FA was interrupting staff account initiation.

pbugni:
Solve problems with new staff user's coming in on invite link, verifying name & dob, promoting auth details into invited account, etc., by suspending 2FA during this bootstrap process.

NB - this means the initial new staff user interaction WILL require them to validate (name & dob) but will NOT require them 2FA.  It would be a significant change to alter that and so, I hope this is acceptable.

### DIFF
--- a/portal/models/login.py
+++ b/portal/models/login.py
@@ -50,6 +50,7 @@ def login_user(user, auth_method=None):
             current_app.config.get("ENABLE_2FA") and
             not current_app.testing and
             not getattr(getattr(request, 'oauth', None), 'user', None) and
+            not user.has_role(ROLE.WRITE_ONLY.value) and
             user.has_role(
                     ROLE.ADMIN.value,
                     ROLE.ANALYST.value,
--- a/portal/views/auth.py
+++ b/portal/views/auth.py
@@ -579,7 +579,12 @@ def next_after_login():
         logout(prevent_redirect=True, reason='reverting to invited account')
         invited_user.promote_to_registered(user)
         db.session.commit()
+
+        # on this brand new promoted account, must fake/suppress 2fa or we'll
+        # repeat the cycle as part of the `login_user()` call
+        session['2FA_verified'] = '2FA verified'
         login_user(invited_user, 'password_authenticated')
+
         if preserve_next_across_sessions:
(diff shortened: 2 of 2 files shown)


mcjustin:
@pbugni  can you please remind me what the lifetime is of invite URLs and 2FA codes?


> ...
> NB - this means the initial new staff user interaction WILL require them to validate (name & dob) but will NOT require them 2FA. It would be a significant change to alter that and so, I hope this is acceptable.

pbugni:
@mcjustin 

For the access tokens (url_tokens) we include in emails, they're good for [30 days](https://github.com/uwcirg/truenth-portal/blob/develop/portal/config/config.py#L116).

For the 2FA 6 digit one time use token, [30 minutes](https://github.com/uwcirg/truenth-portal/blob/develop/portal/models/user.py#L483)

mcjustin:
@pbugni  Thanks for the details. 
1. Do passwords reset URLs use the same 30 day expiry? 
2. Based on https://github.com/uwcirg/truenth-portal/pull/4085, I think the pw reset flow includes 2FA before they can proceed post-reset, is that correct?

pbugni:
@mcjustin the reset-password functionality is built into flask-user.  The expiration on those defaults to 2 days (can be overridden w/ config:

```
lib/python3.7/site-packages/flask_user/settings.py:    um.reset_password_expiration   = sd('USER_RESET_PASSWORD_EXPIRATION',  2*24*3600)   # 2 days
```

The password reset flow includes the challenge (first, last, DOB), but does NOT include 2FA.

mcjustin:
@pbugni  please merge, since this is a blocking error for new staff.

I'm going to create another story for the longer term, to consider requiring a re-login anytime the user _sets_ their pw (aka registers), as we do after resets.  [Flow w/ screenshots here](https://docs.google.com/document/d/1Hr5OieJwTCnySTnvcvNbol4M-jR1NmjVrViUUGlyR-g/edit#bookmark=id.smako4iqhdi0). This, so we have 2FA, since these invite emails can be out in the wild for 30 days.

mcjustin (approved):

PR #4152, opened 2021-10-28
TN-3043-switch-canadian-sites-i18n-2

mcjustin:
https://jira.movember.com/browse/TN-3043
No migration for existing users as there is no effect on what they'll see in content anywhere.


pbugni (approved):
Looks good.  Just rebased w/ master to pick up necessary change for tests.

PR #4155, opened 2021-11-06
TN-3099 UI roles update bugfix

achen2401:
https://jira.movember.com/browse/TN-3099
Fix includes:
When a new role is checked/removed from the roles UI, an existing set of roles is queried first via API to ensure freshness before role(s) are updated.

### DIFF
--- a/portal/static/js/src/profile.js
+++ b/portal/static/js/src/profile.js
@@ -2759,9 +2759,6 @@ export default (function() {
             },
             updateRolesData: function(event) {
                 var visibleRoles = $("#rolesGroup input:checkbox:checked:visible");
-                var roles = $("#rolesGroup input:checkbox:checked").map(function() {
-                    return {name: $(this).val()};
-                }).get();
                 /*
                  * check if a role is selected
                  */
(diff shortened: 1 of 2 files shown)


portal/static/js/src/profile.js
```diff
@@ -2771,12 +2768,38 @@ export default (function() {
                     $(".put-roles-error").html("A role must be selected.");
                     return false;
                 }
+                var isChecked = $(event.target).is(":checked");
+                var checkedRole = event.target.value;
```
mcjustin, line 2772: Perhaps a better name for this is 'changedRole', since the event could be the user either checking it, or unchecking it? That would make this easier to read...

pbugni (approved):

portal/static/js/src/profile.js
```diff
@@ -2771,12 +2768,38 @@ export default (function() {
...
+                        });
+                        if (!isChecked) {
+                            //removed from existing role list
+                            dataRoles = dataRoles.filter(function(role){
+                                return role.name !== changedRole;
+                            });
+                            roles = dataRoles;
+                        } else {
+                            //combine checked role with existing roles
+                            roles = [{name: changedRole}, ...dataRoles];
```
ivan-c, line 2792: Does this array of combined roles need to be de-duplicated?

ivan-c (approved):
Looks great! Thanks for fixing an issue I've run into so many times!! :tada:

portal/static/js/src/profile.js
achen2401: thank you, Ivan, for catching this!  I did just fix it.

PR #4158, opened 2021-11-10
TN-3099 followup bugfix

achen2401:
Followup bug fix to this story:
https://jira.movember.com/browse/TN-3099

Found and address two issues when testing:

- when calling API for current roles, need to make sure that they are not coming from cache, or cached storage (e.g. sessionStorage)
- need to check/uncheck roles on the UI to reflect true current roles of the user


ivan-c (approved):
:tada:

PR #4159, opened 2021-11-15
TN-3101 correct default (prior to initiation) status for EMPRO questionnaires

pbugni:
For EMPRO patients, specifically those shown in the /substudy list, use a default status of "Not yet available" rather than the global study default "expired".

Note - this is intended to go out as a hotfix.

### DIFF
--- a/portal/models/reporting.py
+++ b/portal/models/reporting.py
@@ -113,7 +113,11 @@ def adherence_report(
         # if no current, try previous (as current may be expired)
         last_viable = qb_stats.current_qbd(
             even_if_withdrawn=True) or qb_stats.prev_qbd
-        if last_viable:
+        if not last_viable:
+            # Global study default pre-started is Expired.  See TN-3101
+            if row['status'] == "Expired":
+                row['status'] = "Not Yet Available"
+        else:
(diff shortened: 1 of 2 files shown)


mcjustin (approved):
Looks good to me, only quibble is that the comment at https://github.com/uwcirg/truenth-portal/pull/4159/files#diff-a47de2fe5678b4f011261b24fdeb4af69ea131a992497d799dabad800af22772R1259 isn't keeping pace

portal/models/reporting.py
```diff
@@ -113,7 +113,11 @@ def adherence_report(
         # if no current, try previous (as current may be expired)
         last_viable = qb_stats.current_qbd(
             even_if_withdrawn=True) or qb_stats.prev_qbd
-        if last_viable:
+        if not last_viable:
+            # Global study default pre-started is Expired.  See TN-3101
```
ivan-c, line 117: Found my nitpick! :neckbeard: 
```suggestion
            # Global study default pre-started is Expired. See TN-3101
```

ivan-c (approved):
Looks good!

PR #4161, opened 2021-11-29
TN-3107 restore deactivate UI for non-patient accounts on admin page

achen2401:
https://jira.movember.com/browse/TN-3107
restore deactivation UI for non-patient account on admin page -  allowing a user with admin role to deactivate any non-patient account


mcjustin:
@pbugni  can you please remind me where access to the `/admin` page is governed on the backend?

achen2401:
@mcjustin I see that only users with admin role can access `/admin` page, I checked [here](https://github.com/uwcirg/truenth-portal/blob/af253e8bfd800285a4597f1f5d73b0e84060ce02/portal/views/portal.py#L614)

mcjustin (approved):

PR #4163, opened 2021-12-14
TN-3109 relax assertion for single in-process QNR

pbugni:
log error but continue, when more than a single in-progress QNR is located.

@mcjustin I also verified with this in place:
- we will get (a lot) of error messages when this situation is present
- the staff user can view the respective QNR results from the patient's profile.


mcjustin (approved):
Makes sense to me (noting PEP 8 complaint of course)

ivan-c (approved):

PR #4165, opened 2022-01-05
Bugfix/patient list sorting error

achen2401:
https://jira.movember.com/browse/TN-3115
address JS error occurred when sorting columns (e.g. consent date, study id) that use custom sort function, which possible affected the rendering of patient list table

Changes include:

- remove script tag linking to `https://cdn.rawgit.com/myadzel/6405e60256df579eda8c/raw/e24a756e168cb82d0798685fd3069a75f191783f/alphanum.js` as it is no longer present
- create `alphanum` function used by bootstrap extension lib, that was present in the aforementioned JS file that is no longer available  See error:
Screen Shot 2022-01-05 at 9 10 30 AM
-  create date sorting function that consentDate column used as custom sort function `tnthTables.dateSorter` (tnthTables.js probably existed at some point, but I can't seem to find when it was removed)

### DIFF
--- a/portal/templates/admin/admin_base.html
+++ b/portal/templates/admin/admin_base.html
@@ -1,8 +1,8 @@
 {%- extends "layout.html" -%}
 {% block additional_scripts %}
 <script src="https://cdnjs.cloudflare.com/ajax/libs/jasny-bootstrap/3.1.3/js/jasny-bootstrap.min.js"></script>
+<script src="{{ url_for('static', filename='js/dist/bootstrapTableExtensions.bundle.js') }}"></script>
 <script src="https://cdnjs.cloudflare.com/ajax/libs/bootstrap-table/1.11.1/extensions/export/bootstrap-table-export.min.js"></script>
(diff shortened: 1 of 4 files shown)


portal/templates/admin/admin_base.html
```diff
@@ -1,8 +1,8 @@
 {%- extends "layout.html" -%}
 {% block additional_scripts %}
 <script src="https://cdnjs.cloudflare.com/ajax/libs/jasny-bootstrap/3.1.3/js/jasny-bootstrap.min.js"></script>
+<script src="{{ url_for('static', filename='js/dist/bootstrapTableExtensions.bundle.js') }}"></script>
```
mcjustin, line 4: naive question: are there any potential conflicts between this and the js file in line 8?

mcjustin: Ah, never mind - I see that line 8 is for datatables !

ivan-c (approved):

PR #4167, opened 2022-01-13
TN-3120 Adherence report - improve usability

achen2401:
https://jira.movember.com/browse/TN-3120

Changes include:

- address progress popup disappearing by remove code that hid the export popup on window re-size
- abort/clean up pending export ajax requests on error and prior to each new export
- abort export request if export status stays in PENDING status for more than 1 minute  (shouldn't be in PENDING that long, no?)
- allow access to the most recent export (as a link) (see screenshot):
Screen Shot 2022-01-12 at 4 30 38 PM

### DIFF
--- a/portal/static/js/src/admin.js
+++ b/portal/static/js/src/admin.js
@@ -319,6 +319,7 @@ import {EPROMS_MAIN_STUDY_ID, EPROMS_SUBSTUDY_ID} from "./data/common/consts.js"
                         if (exportStatus === "SUCCESS") {
                             setTimeout(function() {
                                 let resultUrl = statusUrl.replace("/status", "");
+                                console.log("resultUrl ", resultUrl);
                                 self.setCacheReportInfo(resultUrl);
                                 window.location.assign(resultUrl);
                             }.bind(self), 50); //wait a bit before retrieving results
--- a/portal/static/js/src/admin.js
+++ b/portal/static/js/src/admin.js
@@ -232,7 +232,6 @@ import {EPROMS_MAIN_STUDY_ID, EPROMS_SUBSTUDY_ID} from "./data/common/consts.js"
                         return false;
                     }
                     let html = $("#exportReportPopoverWrapper").html();
-                    const DELAY_INTERVAL = 50;
                     $("#adminTableContainer .fixed-table-toolbar .columns-right").append(html);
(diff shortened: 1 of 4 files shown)


mcjustin:
@achen2401  I've seen these reports take 10 minutes and longer. I think we shouldn't abort the export based on a timer.
```
"abort export request if export status stays in PENDING status for more than 1 minute (shouldn't be in PENDING that long, no?)"
```

achen2401:
@mcjustin I should clarify that the abort code that I added will only execute if the export request is stuck at `PENDING` status or 0% for more than 1 minute without ever advancing to `PROGRESS` status, which is when the completion percentage starts to visibly increase.  This scenario I observed in my own instance -  an export that is **stuck at PENDING status** will spin off an infinite series of requests unless it is aborted.

portal/static/js/src/admin.js
```diff
@@ -187,6 +189,12 @@ import {EPROMS_MAIN_STUDY_ID, EPROMS_SUBSTUDY_ID} from "./data/common/consts.js"
                 $("#exportReportContainer").removeClass("open").popover("show");
                 $("#btnExportReport").attr("disabled", true);
                 $(".exportReport__status").addClass("active");
+                this.clearExportReportTimeoutID();
+                this.exportReportProgressTime = new Date();
+                let pastInfo = this.getCacheReportInfo();
+                if (pastInfo) {
+                    $(".exportReport__history").html(`<a href="${pastInfo.url}" target="_blank">View last result exported on ${tnthDates.formatDateString(pastInfo.date, "iso")}</a>`);
```
ivan-c, line 196:
```
View last result exported on ${tnthDates.formatDateString(pastInfo.date, "iso")}</a>`);
```
Do we need to make this string translatable? I'm not sure if site staff view adherence reports

mcjustin (approved):
All this sounds good @achen2401 , just noting that we should need to play with it on staging with a prod-level of realistic data @ivan-c  fyi

ivan-c (approved):
Looks great! Had a minor question about i18n, but ready to merge

mcjustin:
I tried it on staging - works well for me, merge when ready. Can you please add the screenshot and notes about other fxnality to the issue https://jira.movember.com/browse/TN-3120

PR #4184, opened 2022-02-16
Allow pre-existing extension(s) in questionnaire responses.

pbugni:
The pre-existing assertion (present as the storage of our internal extensions was destructive) was preventing the necessary use of extensions in QuestionnaireResponse for out-of-band submissions.

Now checking for presence of `extension`s and extending with internal data rather than replace.


ivan-c (approved):
Looks great, thank you!

PR #4189, opened 2022-03-14
add csrf token lifetime element in template

achen2401:
https://jira.movember.com/browse/TN-3137
Changes include:
- Add JS code to check validity of CSRF token and to handle when CSRF token expires.
- In the event that a CSRF token is considered expired, A dialog will display informing the user as such and refresh the browser page seconds after (User can also refresh the page but clicking on the refresh button).  The dialog is per screenshot:
expiryDialog
- add check for the validity of CSRF token before executing an ajax request

Other changes:
- move frontend code for handling timeout and logout in one place, i.e. `SessionMonitor.js`
- fix JS bug,(add check for null element) when filtering/sorting patient list (see fix in `bootstrap_table_filter_control.js`)


mcjustin (approved):

pbugni (approved):
Wonderful change - thank you for addressing this long standing issue!
Anything we can do to quiet those `semgrep-app` warnings?

portal/factories/app.py
```diff
@@ -171,6 +171,7 @@ def configure_jinja(app):
     app.jinja_env.globals.update(app_text=app_text)
     app.jinja_env.globals.update(ROLE=ROLE)
     app.jinja_env.globals.update(INTERVENTION=INTERVENTION)
+    app.jinja_env.globals.update(WTF_CSRF_TIME_LIMIT=app.config.get("WTF_CSRF_TIME_LIMIT"))
```
ivan-c, line 174: I'm wary about adding a global template var that may only be needed in a handful of templates. 

Do you expect you'll need to use this template var in more templates in the future? Would it make sense to explicitly pass it to the 2 changed templates in this PR?

portal/static/js/flask_user/CsrfTokenChecker.js
```diff
@@ -0,0 +1,196 @@
...
+        modalTitleElement.classList.add("modal-title");
+        modalTitleElement.appendChild(document.createTextNode(i18next.t("Login Attempt Expired")));
+        modalHeaderElement.appendChild(modalTitleElement);
+        //body
+        var modalBodyElement = (document.createElement("div"));
+        modalBodyElement.classList.add("modal-body");
+        //body text
+        var modalBodyContentElement = (document.createElement("div"));
+        modalBodyContentElement.classList.add("text-center");
+        modalBodyContentElement.appendChild(document.createTextNode(i18next.t("Your login attempt timed out.  Refreshing the page...")));
```
ivan-c, line 62:
```suggestion
        modalBodyContentElement.appendChild(document.createTextNode(i18next.t("Your login attempt timed out. Refreshing the page...")));
```

ivan-c (approved):
Looks good! Few minor nitpicks, but ready to merge

portal/static/js/flask_user/CsrfTokenChecker.js
```diff
@@ -0,0 +1,164 @@
...
+                '</div>' +
+            '</div>' +
+        '</div>';
+        var modalElement = document.createElement("div");
+        modalElement.setAttribute("id", this.modalElementId);
+        modalElement.setAttribute("tabIndex", -1);
+        modalElement.setAttribute("role", "dialog");
+        modalElement.classList.add("modal");
+        modalElement.classList.add("fade");
+        modalElement.innerHTML = contentHTML;
```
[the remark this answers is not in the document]
ivan-c: Resolve these if you've done the busy work of fixing them! :)

portal/factories/app.py
achen2401: The two changed templates are the base templates from which many other templates are **extending** from.  I _**think**_ this approach is cleaner than passing the variable to all the ones that need them.

PR #4191, opened 2022-03-24
IRONN-45 add CLI to delete post withdrawn qnrs

pbugni:
As per support story IRONN-45, questionnaire responses submitted beyond the visit when a user is withdrawn need to be deleted from the system.

A rather delicate, complicated process, new CLI function `remove-post-withdrawn-qnrs` does just that, including an audit row to track the change.


mcjustin (commented):
I think I follow what you're doing here... as I read this and the support ticket, this could be run on any of the patients in the list that you and Alyssa have been passing back and forth, and it will do the right thing (take action on only a subset)?

mcjustin (approved):

PR #4205, opened 2022-05-19
TN-3157 potential fix; enhance trigger states logging

pbugni:
TN-3157 happened due to a failure to persist the correct visit month in a trigger_states transition.

This PR includes a one line code change (shifting the `db.session.commit()` out of a first month only clause in `initiate_trigger()`, so we always persist new or changed trigger_states rows from that function), as they should always be saved, and may potentially be the source bug behind the issue.

All other changes are to better capture trigger_states transitions and persistence in a manner that should be easier to locate in the logs, should the problem persist.

### DIFF
--- a/portal/trigger_states/empro_states.py
+++ b/portal/trigger_states/empro_states.py
@@ -92,10 +92,10 @@ def enter_user_trigger_critical_section(user_id):
     sm = EMPRO_state(ts)
     sm.begin_process()
     # Record the historical transformation via insert.
-    current_app.logger.debug(
-        "record state change to 'inprocess' from "
-        f"enter_user_trigger_critical_section({user_id})")
     ts.insert(from_copy=True)
+    current_app.logger.debug(
+        "persist-trigger_states-new from enter_user_trigger_critical_section()"
+        f" {ts}")
 
     # Now 'inprocess', obtain the lock to be freed at the conclusion
     # of `evaluate_triggers()`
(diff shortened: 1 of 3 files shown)


portal/trigger_states/empro_states.py
```diff
@@ -92,10 +92,10 @@ def enter_user_trigger_critical_section(user_id):
     sm = EMPRO_state(ts)
     sm.begin_process()
     # Record the historical transformation via insert.
-    current_app.logger.debug(
-        "record state change to 'inprocess' from "
-        f"enter_user_trigger_critical_section({user_id})")
     ts.insert(from_copy=True)
+    current_app.logger.debug(
+        "persist-trigger_states-new from enter_user_trigger_critical_section()"
```
ivan-c, line 97: >persist-trigger_states-new

Are the long "tokens" meant to be used as an easy way to search ELK fields, ie a workaround for exact matches? Might be worth adding a TODO to figure out ELK exact matches, but maybe you expect to cut this logging out later

ivan-c (approved):
Looks good! Had a nitpick question but ready to merge

portal/trigger_states/empro_states.py
pbugni: I believe you can do exact matches with the `field.keyword` selector in ELK - the "tokens" you refer to are just me trying to be consistent with spelling, so a grep will hit them all.  Yes, cutting this down later would be reasonable, as would tuning logging to not include debug.

mcjustin (approved):

PR #4207, opened 2022-05-23
scope error on logging statement, interrupting EMPRO trigger task

pbugni:
The log statement isn't nested w/i the for loop, but one level out at the point of commit, causing these errors seen in the docker logs:

```
celeryworkerslow_1  | {"written_at": "2022-05-23T16:01:00.019Z", "written_ts": 1653321660019708000, "msg": "Task portal.tasks.process_triggers_task[a274233b-17a6-455d-80dd-a4a72cf353b7] succeeded in 0.014829255640506744s: 'Unexpected exception in `process_triggers_task` on 148 : local variable \\'ts\\' referenced before assignment'", "type": "log", "logger": "celery.app.trace", "thread": "MainThread", "level": "INFO", "module": "trace", "line_no": 124, "correlation_id": "d10db52c-d896-11ec-9b28-0242ac120006"}
```

Was a bit of a mystery as to why this didn't get picked up in our log watcher - presumably as the log level shows as INFO - which doesn't agree with [the source code logging such as ERROR](https://github.com/uwcirg/truenth-portal/blob/develop/portal/tasks.py#L81).  Looks like the "local" `logger` was improperly configured and such log messages were all going in the bit bucket.  Second commit here makes consistent use of `current_app.logger` known to be functional.

### DIFF
--- a/portal/tasks.py
+++ b/portal/tasks.py
@@ -12,7 +12,6 @@ from functools import wraps
 import json
 from traceback import format_exc
 
-from celery.utils.log import get_task_logger
 from flask import current_app
 import redis
 from requests import Request, Session, post
@@ -49,8 +48,6 @@ from .models.user import User, UserRoles
 #   rdb.set_trace()
 # Follow instructions from celery console, i.e. telnet 127.0.0.1 6900
 
-logger = get_task_logger(__name__)
-
 celery = create_celery(create_app())
 LOW_PRIORITY = 'low_priority'
 
@@ -63,7 +60,7 @@ def scheduled_task(func):
 
         if not manual_run and job_id and not check_active(job_id):
             message = "Job id `{}` inactive.".format(job_id)
-            logger.debug(message)
+            current_app.logger.debug(message)
             return message
 
         try:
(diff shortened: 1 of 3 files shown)


ivan-c (approved):

mcjustin (commented):
@pbugni  I see the logging mech being replaced here, but not a change for "The log statement isn't nested w/i the for loop, but one level out at the point of commit, causing these errors seen in the docker logs..."

pbugni:
> @pbugni I see the logging mech being replaced here, but not a change for "The log statement isn't nested w/i the for loop, but one level out at the point of commit, causing these errors seen in the docker logs..."

@mcjustin see first commit in module `empro_states.py`

portal/tasks.py
```diff
@@ -49,8 +48,6 @@
 #   rdb.set_trace()
 # Follow instructions from celery console, i.e. telnet 127.0.0.1 6900
 
-logger = get_task_logger(__name__)
```
ivan-c, line 52: I think removing this line is causing an import to break here in [`celery_worker.py`](https://github.com/uwcirg/truenth-portal/blob/v22.5.2.2/portal/celery_worker.py#L27)

[See CI error](https://gitlab.com/uwcirg/truenth-portal/-/jobs/2495593993#L92)

ivan-c (changes requested):
Think I found an import error, looks good otherwise

ivan-c (approved):

mcjustin (approved):
Thanks @pbugni  for reminding me that python scoping is controlled by whitespace.

pbugni:
> 

Right, the point of that whitespace is to move the logging statement within the `for ts in ...` scope starting on line 309.  The error was caused by that being outside of the scope, due to indent level.

PR #4212, opened 2022-06-07
Resorting celery task logger imports

pbugni:
Resorting celery task logger imports, in an attempt to see if this may restore celerybeat health.

(yes, this is a wild guess at the moment, but the only obvious change since updates to the scheduled_jobs stopped functioning and celerybeat started reporting errors)


ivan-c (commented):
Looks good (assuming it works on eproms-test)!

PR #4213, opened 2022-06-09
TN-3157 found a more plausible reason for losing track of visit_month

pbugni:
Reworked the asynchronous locks to more reliably prevent interruption or race conditions when updating a user's trigger_stats from various threads (jobs, user actions, etc).

Caught another incorrect visit_month scenario with enhanced logging, in the act.  Almost certain the source is behind the trigger_states.insert(from_copy=True) calls, which makes use of SQLAlchemy's `make_transient()` method.

Using `make_transient()` coupled with `db.session.commit()` can lead to unexpected loss of object results.
Now intentionally reloading after any such calls.

See also https://github.com/sqlalchemy/sqlalchemy/issues/3640

### DIFF
--- a/portal/trigger_states/empro_states.py
+++ b/portal/trigger_states/empro_states.py
@@ -134,7 +134,8 @@ def users_trigger_state(user_id):
     # if semaphore is locked for user, confirm "inprocess"
     semaphore = TimeoutLock(key=EMPRO_LOCK_KEY.format(user_id=user_id))
     if semaphore and ts.state != "inprocess":
-        current_app.logger.error(f"found {ts.state} w/ semaphore locked?!")
+        current_app.logger.error(
+            f"found {ts.state} for user {ts.user_id} w/ semaphore locked?!")
     return ts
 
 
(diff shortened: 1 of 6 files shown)


portal/trigger_states/empro_states.py
```diff
@@ -117,16 +125,16 @@ def users_trigger_state(user_id):
...
     ts = TriggerState.query.filter(
         TriggerState.user_id == user_id).order_by(
         TriggerState.timestamp.desc()).first()
     if not ts:
         ts = TriggerState(user_id=user_id, state='unstarted')
+
+    # if semaphore is locked for user, confirm "inprocess"
+    semaphore = TimeoutLock(key=EMPRO_LOCK_KEY.format(user_id=user_id))
+    if semaphore and ts.state != "inprocess":
+        current_app.logger.error(f"found {ts.state} w/ semaphore locked?!")
```
ivan-c, line 137: Would it worthwhile to include the `user_id` in the log message body?

ivan-c (approved):
Looks good!

portal/trigger_states/empro_states.py
pbugni: sure, wise addition - thanks!

mcjustin (approved):

PR #4218, opened 2022-07-08
elongate the handling time for an ajax request, also elongate the

achen2401:
Address recent occurrences of timed out Ajax call errors as seen in logs.

Changes include:

- Elongate the time that an Ajax request can take before terminating to **5** seconds (it was 3 seconds previously)
- Elongate the wait time between each Ajax request retry to **5** seconds (it was 2)


mcjustin (approved):

ivan-c (approved):
Thank you!

PR #4219, opened 2022-07-14
IRONN-79 Add a new site in Switzerland

mcjustin:
https://jira.movember.com/browse/IRONN-79 . This is not the first site in Switzerland, and the languages are used at other Swiss sites.

### DIFF
--- a/portal/config/eproms/Organization.json
+++ b/portal/config/eproms/Organization.json
@@ -2717,6 +2717,51 @@
           "valueCodeableConcept": {
             "coding": [
               {
+                "code": "de_CH",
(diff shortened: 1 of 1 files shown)


mcjustin:
Marking this as draft while I see if they need this as a hotfix... The SIV date of 7/19 probably means that we need it as a hotfix this week...

mcjustin:
@ivan-c  PCCTC says that next Friday's release would be soon enough, so this is ready for your review, and we'll merge to develop.

ivan-c (approved):

PR #4227, opened 2022-07-25
debugging redis/celery healthcheck issue - adding logging only

pbugni:
WIP - don't merge.


pbugni:
https://jira.movember.com/browse/TN-3177 captures issue.  rolling out this logging only change "fixed" it, but so did forcing a restart of all services.

closing PR.

PR #4228, opened 2022-07-25
TN-3178 don't start indef prior to consent

pbugni:
Now that we allow future consent dates, need to prevent the "special" indefinite QB from being started prior to the passage of the consent moment.


mcjustin (approved):
I don't fully understand how these work, but this should be easy to test all the same...
```q.relative_start > self.as_of_date:```

PR #4229, opened 2022-08-01
Traefik Ingress

ivan-c:
- Add docker-compose override to allow config via Traefik
  - Set `SERVER_NAME` in `.env` to match DNS Hostname (eg `eproms.ubu.ivanc.dev.cirg.uw.edu`)
- Add docker-compose override for exposing port (current apache deploy method)
  - `ports:` section will be removed from docker-compose.yaml at a later date

### DIFF
--- a/docker/docker-compose.traefik-ingress.yaml
+++ b/docker/docker-compose.traefik-ingress.yaml
@@ -1,0 +1,26 @@
+# docker-compose ingress overrides for traefik
+---
+version: "3.9"
+services:
+  web:
+    labels:
+      - "traefik.enable=true"
+      # Traefik will route requests with Host matching the SERVER_NAME environment variable (see .env)
+      - "traefik.http.routers.portal-${COMPOSE_PROJECT_NAME}.rule=Host(`${SERVER_NAME}`)"
+
+      - "traefik.http.routers.portal-${COMPOSE_PROJECT_NAME}.entrypoints=websecure"
+      - "traefik.http.routers.portal-${COMPOSE_PROJECT_NAME}.tls=true"
+      - "traefik.http.routers.portal-${COMPOSE_PROJECT_NAME}.tls.certresolver=letsencrypt"
+    networks:
+      - ingress
+      # TODO rename to internal
+      # internal network
+      - default
+    environment:
+      PREFERRED_URL_SCHEME: https
+
+networks:
+  # ingress network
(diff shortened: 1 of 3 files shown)


docker/docker-compose.traefik-ingress.yaml
```diff
@@ -0,0 +1,26 @@
...
+      - "traefik.enable=true"
+      # Traefik will route requests with Host matching the SERVER_NAME environment variable (see .env)
+      - "traefik.http.routers.portal-${COMPOSE_PROJECT_NAME}.rule=Host(`${SERVER_NAME}`)"
+
+      - "traefik.http.routers.portal-${COMPOSE_PROJECT_NAME}.entrypoints=websecure"
+      - "traefik.http.routers.portal-${COMPOSE_PROJECT_NAME}.tls=true"
+      - "traefik.http.routers.portal-${COMPOSE_PROJECT_NAME}.tls.certresolver=letsencrypt"
+    networks:
+      - ingress
+      # TODO rename to internal
```
pbugni, line 16: why is this left as a TODO?  likely to break current deploys?

pbugni (approved):
nice.

docker/docker-compose.traefik-ingress.yaml
ivan-c: > why is this left as a TODO? likely to break current deploys?

Yup, we'll need to bring the docker-compose project down before making this change (otherwise we'll have 2 running projects)

(By default, docker-compose uses `default` as the network name)

pbugni: Makes sense, sounds challenging to schedule w/ all of our deploys.

ivan-c: The easiest way to roll that change out is to: take a backup, wipe the dc project, deploy the change, and restore the DB. 99% of that process is scripted, so restoring the psql dump would be the largest contributor to downtime (disk IO).

I regularly do this with eproms-stage, and it doesn't take more than a minute to load the prod psql dump

ivan-c:
@jhowe-uw 

Can you take a look at this when you get a chance? I've got it deployed on `eproms.ubu.ivanc.dev.cirg.uw.edu` if you want to take a look at how it's set up.

Thanks!

jhowe-uw (approved):

PR #4231, opened 2022-08-09
Feature/audit request extend timeout

achen2401:
address issue https://jira.movember.com/browse/IRONN-86
Extend timeout for audit Ajax request, noting that some can take up to almost 30 seconds.  For example see: https://eproms.truenth.org/profile/4 which has more than 3000 audit records.

Changes include:

- extend maximum timeout for an audit Ajax request to 5 minutes to confront accounts that may have tons of audit records

### DIFF
--- a/portal/static/js/src/modules/TnthAjax.js
+++ b/portal/static/js/src/modules/TnthAjax.js
@@ -78,7 +78,7 @@ export default { /*global $ */
                 (function(self, url, method, userId, params, callback) {
                     setTimeout(function () {
                       self.sendRequest(url, method, userId, params, callback);
-                    }, RETRY_INTERVAL); //retry after 5
+                    }, RETRY_INTERVAL); //retry after 5 seconds
                 })(self, url, method, userId, params, callback);
             } else {
                 fieldHelper.showError(targetField);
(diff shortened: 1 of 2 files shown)


portal/static/js/src/modules/TnthAjax.js
```diff
@@ -22,8 +22,9 @@ export default { /*global $ */
     },
     "sendRequest": function(url, method, userId, params, callback) {
         if (!url) { return false; }
-        var REQUEST_TIMEOUT_INTERVAL = 5000;
-        var defaultParams = {type: method ? method : "GET", url: url, attempts: 0, max_attempts: MAX_ATTEMPTS, contentType: "application/json; charset=utf-8", dataType: "json", sync: false, timeout: 5000, data: null, useWorker: false, async: true};
+        var REQUEST_TIMEOUT_INTERVAL = 10 * 1000; // default timed out at 10 seconds
+        var RETRY_INTERVAL = 5000; // wait 5 seconds before retry
```
pbugni, line 26: does it make sense to retry after 5 seconds, when a request is given 10 to timeout?
seems like this might lead to stacking requests?  i'm not sure, but using the same interval for timeout and retries makes sense to me.

achen2401: thank you @pbugni that makes sense. I fixed it.

ivan-c (approved):

PR #4236, opened 2022-08-22
TN-3136

pbugni:
WIP - draft at this time.


portal/views/clinician.py
```diff
@@ -15,15 +15,16 @@
 clinician_api = Blueprint('clinician_api', __name__)
 
 
-def clinician_query(acting_user, org_filter=None):
+def clinician_query(acting_user, org_filter=None, include_staff=False):
```
mcjustin, line 18: Seems odd for this function to optionally query for users who have the staff role but not clinician or primary_investigator (the latter two are rather closely related, so that makes sense). Def not a blocker from my pov, I expect that refactoring something like this to be more general is prob not worth it.

mcjustin (approved):

mcjustin (approved):
Noting that the report as emailed looks good to me, except that I have yet to see one with the "Clinician Survey Completion Date" column populated, but I expect that's because we don't have recent tests in eproms-test (this report is only for the current EMPRO timepoint).

pbugni:
We should now be ready to merge, and I'd like to soon in preparation for Friday's rollout.  Anna has submitted several change requests which have all been closed out - see TN-3136

portal/config/eproms/ScheduledJob.json
```diff
@@ -5,13 +5,27 @@
...
+      "active": true,
+      "args": null,
+      "id": 6,
+      "kwargs": {
+        "research_study_id": 1,
+        "org_id": 20000
+      },
+      "name": "EMPRO site summary email",
+      "resourceType": "ScheduledJob",
+      "schedule": "r 20 20 * *",
```
ivan-c, line 26: [At 20:rr on day-of-month 20](https://crontab.guru/#1_20_20_*_*)

ie, random minute, at 8pm on the 20th of every month :+1:

ivan-c (approved):

PR #4241, opened 2022-09-19
IRONN-100 text change for error when IRONMAN questionnaire is pending

achen2401:
As requested from https://jira.movember.com/browse/IRONN-100
Update error text displayed in clinician intervention questionnaire section in patient profile when IRONMAN questionnaire is pending for user.

### DIFF
--- a/portal/models/qb_status.py
+++ b/portal/models/qb_status.py
@@ -571,7 +571,7 @@ def patient_research_study_status(patient, ignore_QB_status=False):
             if results[0]['ready']:
                 # Clear ready status when base has pending work
                 rs_status['ready'] = False
-                rs_status['errors'].append('Pending work in base study')
+                rs_status['errors'].append('Participant has pending questionnaire in main IRONMAN study')
             elif not patient.email_ready():
                 # Avoid errors from automated emails, that is, email required
                 rs_status['ready'] = False


mcjustin (approved):
@achen2401  sounds like this is just what Anna asked for. Just flagging @pbugni  here, since I _think_ it's rare for the API to return errors in expected flows like this... ack that there are a lot of business rules around this feature, I'm fine to leave the backend as-is.

```
For context, the text, `Pending work in base study` comes from [here](https://github.com/uwcirg/truenth-portal/blob/41143f7aa5bc868ed377aa1b948c485f1ef51ddf/portal/models/qb_status.py#L574)  and it is displayed if the [api ](https://eproms.truenth.org/api/patient/2175/research_study)returns error (see under errors).  That can be easily changed as requested.  Looking at the code, it does look like the error is generated if there is still questionnaire(s) pending for the base study. [see here](https://github.com/uwcirg/truenth-portal/blob/41143f7aa5bc868ed377aa1b948c485f1ef51ddf/portal/models/qb_status.py#L565)

However, I don't believe that will prevent a hard trigger from being actioned. I tested this myself in the Test environment.  See api output [here](https://eproms-test.cirg.washington.edu/api/patient/3982/research_study) for patient 2175 in Test - you see the error in the api but the post intervention questionnaire is still available to be filled in...
```

PR #4242, opened 2022-09-20
TN-3188 hide "pending work in base study" error in clinician intervention questionnaire

achen2401:
address https://jira.movember.com/browse/TN-3188
Use `intervention_qnr_eligible` flag returned from `/api/patient/[id]/research_study` to determine whether to display error(s) in the clinician intervention questionnaire section of the patient profile page.
**Note** the flag is TRUE even if there is pending work in base study.  See example [here](https://eproms-test.cirg.washington.edu/api/patient/3982/research_study).  So that specific error will not display on the UI.


mcjustin (approved):

ivan-c (approved):

pbugni (approved):
💯

PR #4245, opened 2022-09-27
TN-3190 correct display of <deleted> clinicians

pbugni:
Although primarily a symptom of fake test data (due to outdated assignment of plain staff as clinicians), clean up how deleted clinicians are displayed in the `/patients/substudy` list.
 
1. without a current clinician, use empty string, not `<deleted>`.
2. Include text should we ever again find a clinician w/o a "name in map", as picked up from test-only data on test system.

### DIFF
--- a/portal/views/patients.py
+++ b/portal/views/patients.py
@@ -78,8 +78,8 @@ def render_patients_list(
             patient.current_qb = qb_status['visit_name']
             if research_study_id == EMPRO_RS_ID:
                 patient.clinician = '; '.join(
-                    (clinician_name_map.get(c.id, "<deleted>") for c in
-                     patient.clinicians)) or "<deleted>"
+                    (clinician_name_map.get(c.id, "not in map") for c in
+                     patient.clinicians)) or ""
                 patient.action_state = qb_status['action_state'].title() \
                     if qb_status['action_state'] else ""
             patients_list.append(patient)


portal/views/patients.py
```diff
@@ -78,8 +78,8 @@ def render_patients_list(
             patient.current_qb = qb_status['visit_name']
             if research_study_id == EMPRO_RS_ID:
                 patient.clinician = '; '.join(
-                    (clinician_name_map.get(c.id, "<deleted>") for c in
-                     patient.clinicians)) or "<deleted>"
+                    (clinician_name_map.get(c.id, "not in map") for c in
+                     patient.clinicians)) or ""
```
mcjustin, line 82: I'm having trouble squaring this with your comment:
`without a current clinician, use empty string, not "".`
Sorry, my python is weak...

pbugni: line 81, we attempt to look up the clinician id (`c.id`) in the name map, providing a default of "not in map" should the lookup fail - which should never happen.  (it was happening as we had non clinicians in the test data, and if it returns, i want a clear indication of the problem.)
line 82, if the entire clause left of the `or` is logically false, i.e. the patient had zero clinicians, we'll get back the empty string from the right hand side of the `or` expression.

mcjustin (approved):

PR #4259, opened 2022-11-01
TN-3198 prevent PI from being removed from patient's assigned clinician list

achen2401:
see https://jira.movember.com/browse/TN-3199

Changes include:

- remove delete icon button from PI on the assigned clinician list in patient profile (see screenshot):
Screen Shot 2022-10-31 at 5 51 10 PM
- disabled clinicians selector element _while_ a selected clinician is being added/removed to ensure that data is updated before another selection occurs.


mcjustin (approved):

pbugni (approved):
special thanks for disabling controls while we wait for the change request to be processed - quite possibly the source of the problem!

PR #4262, opened 2022-11-09
UI bug - PI deletion on treating clinician section in profile

achen2401:
Follow up to https://jira.movember.com/browse/TN-3198
Issue found:
User can still delete PI in the Treating Clinician section even though the delete icon isn't showing for UI element.

Cause:
The **on click** event to delete a clinician is attached to the the parent UI element that contains the delete icon. So clicking anywhere on it will invoke the delete clinician event even if the delete icon isn't showing.

Fix:
Attach on click event to delete clinician to the delete icon button itself


mcjustin (approved):
Looks good to me... not clear why https://gitlab.com/uwcirg/truenth-portal/-/pipelines/689297593 failed for it.

achen2401:
thanks, I re-ran the test and that passed.

PR #4272, opened 2022-12-09
TN-3202 fix for complicated EMPRO trigger dates moving

pbugni:
if subsequent global visit has become due after user started EMPRO work (and previously there wasn't a global completed within 4 weeks of EMPRO consent).

very isolated change - only impacts EMPRO trigger-date[1] calculations, and then only if all of the above criteria are met.
See comments in code.

[1] trigger-date: term for the initial start of the study, not related to an EMPRO "triggered status"


pbugni:
Note also, the comments at the tail end of TN-3202, in which I hand tested every patient affected by this change.

mcjustin (approved):

PR #4277, opened 2022-12-28
default 5 year expiration on user_consents coming into play

pbugni:
the use of user_consents.expired is not being managed and erroneously hiding consent rows.  delete checks against this column as hot fix for now.


pbugni:
superseded by https://github.com/uwcirg/truenth-portal/pull/4278 (this one picks up a change in develop we don't want in the hotfix).

PR #4282, opened 2022-12-30
Updated elgohr/Publish-Docker-Github-Action to a supported version (v5)

elgohr:
elgohr/Publish-Docker-Github-Action@master is not supported anymore

### DIFF
--- a/.github/workflows/build-deliver.yaml
+++ b/.github/workflows/build-deliver.yaml
@@ -17,7 +17,7 @@ jobs:
 
       - name: Publish to GitHub Container Registry
         # TODO: pin to hash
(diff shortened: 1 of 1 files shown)


ivan-c:
Thank you for the heads-up (and amazing GitHub Action)!

We don't have an automated tool (eg [Renovate](https://github.com/renovatebot/renovate) to maintain our version pins, so I'd prefer to always use the latest version, rather than pinning a specific version  (will address in a follow-up commit).

There's a chance we'll need to "rollback" to an old pinned version, but we haven't encountered any breaking changes yet (our usage of this Action is pretty simple).

PR #4289, opened 2023-01-24
TN-3217 PI role account creation - prevent invalid role combo

achen2401:
https://jira.movember.com/browse/TN-3217

Description of the changes:

- When the role of Primary Investigator is checked - in account creation page or profile page, the UI will display warning about other required role(s), e.g. Staff, Admin Staff, and Clinician (see example screenshot).
- The user will be prevented from continuing unless ONE of those required roles is checked.

Example screenshot:
Screen Shot 2023-01-24 at 12 04 49 PM


mcjustin (approved):
Looks right to me...

pbugni (approved):
Well done.  Thanks @achen2401

PR #4291, opened 2023-02-03
Bugfix/pi in profile page

achen2401:
https://jira.movember.com/browse/TN-3217
Follow up fix to the comment
Scenario tested:
- User **unchecked** all staff roles (staff, admin staff, clinician and primary investigator roles)
- check only the Primary Investigator role on the profile page.

The fix is to display warning to user, note this will still result in Primary Investigator role being select, what to do here?  Do you auto-check one of the required roles? (i.e. staff, admin staff or clinician)?

### DIFF
--- a/portal/static/js/src/profile.js
+++ b/portal/static/js/src/profile.js
@@ -2938,9 +2938,19 @@ export default (function() {
                 $(event.target).prop("checked", true);
                 return;
               }
-              var visibleRoles = $(
-                "#rolesGroup input:checkbox:checked:visible"
-              );
+              // if primary investigator is checked
+              if ($(event.target).val() === "primary_investigator" &&
+                  $(event.target).is(":checked")) {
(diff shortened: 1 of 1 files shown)


sierramatice:
@achen2401 I do want to clarify that the system did not actually save the PI only state, it just appeared to from a UI standpoint. The saved state was still PI + the previous role!

achen2401:
thanks @sierramatice I did fix that in this PR - the primary investigator role will be saved (without previous role(s)) in the scenario.

mcjustin (approved):
Per my message in slack:
I'd like this profile page to act the same as the acct creation page, which per https://github.com/uwcirg/truenth-portal/pull/4289  :
"The user will be prevented from continuing unless ONE of those required roles is checked."
So in this case, prevent selection of the PI role unless one of the other required roles is selected.

... approving this w/ the understanding that this is what the code now does.

ivan-c (approved):

PR #4298, opened 2023-02-28
TN-3140 Catch second submission of a known QNR dyad in the act

pbugni:
as per TN-3140, need to prevent duplicate QNRs, that is, ones that match on visit, a function of (questionnaire_bank_id, qb_iteration), and instrument for a given patient.

given the complexity of source data - interventions including AE submitting such QNRs, don't raise but rather log the error to trigger a notification.

### DIFF
--- a/portal/models/questionnaire_response.py
+++ b/portal/models/questionnaire_response.py
@@ -180,6 +180,22 @@ class QuestionnaireResponse(db.Model):
                 subject_id=self.subject_id, user_id=acting_user_id,
                 context='assessment', comment=msg)
             db.session.add(audit)
+            # TN-3140 multiple QNRs found for visit/questionnaire dyad
+            # Generate an error for alerts, should this look to be a fresh
+            # duplicate.  Ignore if we don't have a valid questionnaire bank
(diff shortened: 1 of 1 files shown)


pbugni:
Complicated reproduction, as this is trying to capture an unexpected situation.

Done on eproms-test:
1. complete all questionnaires for a patient.
2. use SQL to move one of the submitted responses out of the way - i.e. assign a different subject_id temporarily
3. hit the /timeline&purge=true for user to update status
4. return to AE to take the questionnaire moved out of the way in step 2
5. BEFORE submitting from AE, reverse the move in step 2
6. submit from AE, and note in the logs:
```
 {"written_at": "2023-03-10T03:34:54.078Z", "written_ts": 1678419294078849000, "msg": "Second submission for an existing QNR dyad received. Patient: 4089, QNR 4017", "type": "log", "logger": "flask.app", "thread": "MainThread", "level": "ERROR", "module": "questionnaire_response", "line_no": 198, "correlation_id": "859510f0-bef4-11ed-8731-0242ac130008"}
 ```

mcjustin (approved):

PR #4301, opened 2023-03-27
IRONN-172 new API to rescore an interrupted scoring on any given QNR

pbugni:
API at: `/api/questionnaire_response/<int:qnr_id>/$rescore`

only available to users with `admin` role.


ivan-c (approved):
Looks great, appreciate the helpful notes in the function docstrings!

mcjustin (approved):

PR #4303, opened 2023-03-31
Update broken codesystem URL

ivan-c:
Update broken codesystem URL to include FHIR version


pbugni (approved):
Thank you @ivan-c .  Let's hope this URL has a longer life (it should).

PR #4305, opened 2023-04-10
TN-3226 new celery job to intentionally raise exceptions

pbugni:
NB, this job is not scheduled, but can be manually triggered anytime we need to verify handling and routing of errors.


ivan-c (approved):
Looks great, thank you!! Will merge once tests pass, unless you beat me to it ;)

PR #4312, opened 2023-04-27
add check to ensure user is still authorized to prevent stale session

achen2401:
address https://jira.movember.com/browse/IRONN-176
and follow up to meeting, see [slack convo](https://cirg.slack.com/archives/C5EU4L7PF/p1682532067179999)
I believe the current behavior is:

- Frontend receives when session will`expires_in` information from the backend.
- Frontend calculates the session lifetime.
- Frontend pops up a warning a minute before the session expires.
- Frontend logs out the user when the session expires.

This change to add extra check, in case for some reason, the previous steps fail to ensure session freshness, via call to `/api/me`.  If the API call returns `401` unauthorized status code - the user will be logged out

### DIFF
--- a/portal/static/js/portal_wrapper/SessionMonitor.js
+++ b/portal/static/js/portal_wrapper/SessionMonitor.js
@@ -44,6 +44,17 @@ var SessionMonitorObj = function() { /* global $ */
                                 "X-CSRFToken": o
                             };
                         }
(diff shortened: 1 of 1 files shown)


pbugni (approved):
💯 Thank you!

mcjustin (approved):

PR #4328, opened 2023-06-30
Test Deadlock Workaround

ivan-c:
Split tests causing deadlocks to separate test suite (tox "env") invocation


pbugni:
@ivan-c I'd suggest you kill this branch - the source of the problem was found to be something of a deadlock between the test harness and the background jobs it was waiting on, which didn't have the assumed system_user in the database.

thank you for the attempt to help!

PR #4330, opened 2023-07-12
generate `site.cfg` for celery workers

pbugni:
The site_persistence machinery generates a site specific (eproms vs USA) config file for consumption on start-up.

Previously, only the `web` service was executing `flask sync` which generates this file for consumption.

New `generate-site-cfg` CLI included in the respective `celery` launch commands does exactly that, so the jobs have the same configuration as the web service.

### DIFF
--- a/docker/docker-compose.dev.yaml
+++ b/docker/docker-compose.dev.yaml
@@ -17,3 +17,17 @@ services:
         target: /mnt/code
         type: bind
         read_only: true
+
+  celeryworker:
+    volumes:
+      - source: ../
+        target: /mnt/code
+        type: bind
+        read_only: true
+
+  celeryworkerslow:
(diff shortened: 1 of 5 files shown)


docker/docker-compose.dev.yaml
```diff
@@ -17,3 +17,17 @@ services:
         target: /mnt/code
         type: bind
         read_only: true
+
+  celeryworker:
+    volumes:
+      - source: ../
+        target: /mnt/code
+        type: bind
+        read_only: true
```
ivan-c, line 26: I think this may cause problems if `site.cfg` is expected to be written somewhere in there. I've sometimes had to comment out the same line for the portal, when `flask sync` is invoked

PR #4332, opened 2023-07-25
threaded adherence cache production error

pbugni:
add guard and premature exit should adherence cache be working on timeline data purged in another thread.


ivan-c (approved):
Looks good, thanks!!

Applied to prod on 2023-07-26 and haven't seen any exceptions since

PR #4338, opened 2023-09-14
TN-3250, incorrect type in check for withdrawn from study

pbugni:
The view used to determine which questionnaires (assessments) a given patient needs to submit, was incorrectly comparing an `OverallStatus` instance with the string `Withdrawn`.  Thus, patients withdrawn from a given study would incorrectly be administered assessments.

Extending this hotfix to include exclusion of questionnaire responses after withdrawn visits - actually now excluding ALL questionnaire responses that aren't associated with a valid questionnaire bank (i.e. a visit).

### DIFF
--- a/portal/eproms/templates/eproms/assessment_engine/ae_complete.html
+++ b/portal/eproms/templates/eproms/assessment_engine/ae_complete.html
@@ -38,17 +38,24 @@
                 <div class="text-content">{{_("No questionnaire is due.")}}</div>
             {%- endcall -%}
         {%- else -%}
-            <!-- substudy due card -->
-            {{empro_due(
-                substudy_assessment_is_due=substudy_assessment_is_due,
-                assessment_status=assessment_status,
-                OverallStatus=OverallStatus,
-                substudy_assessment_status=substudy_assessment_status,
-                substudy_due_date=substudy_due_date,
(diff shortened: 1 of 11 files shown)


mcjustin:
> The view used to determine which questionnaires (assessments) a given patient needs to submit, was incorrectly comparing an `OverallStatus` instance with the string `Withdrawn`. Thus, patients withdrawn from a given study would incorrectly be administered assessments if any had been left `in-progress`.

@pbugni  that might explain the first QR that patient 5596 completed post-withdrawal (#65241), but it doesn't *seem* to explain what happened with their QR 72971 completed 2023-07-06, many months after their withdrawal in Oct '22. 
https://movember.atlassian.net/browse/TN-3250?focusedCommentId=246743

pbugni:
@mcjustin sorry for the confusion in my initial comment, the in-progress scenario is not the only one that would incorrectly have been given the chance, I had just watched that scenario in the debugger.  The check was in general, for a user being withdrawn from a study.

mcjustin (approved):

ivan-c (approved):

mcjustin (approved):

pbugni:
i'd like to rename the branch (from hotfix/x to feature/x) as we're not releasing as hotfix, but that would mean closing out this PR and losing the history, so, we'll live with it.

PR #4339, opened 2023-09-15
TN-3259 research data export - UI not showing the most current download link

achen2401:
Address story: [TN-3259](https://movember.atlassian.net/browse/TN-3259?atlOrigin=eyJpIjoiM2Q1MzQ2ZjE0YzY1NDBjMDk0OTM2OWMxMWMzMTc3ODAiLCJwIjoiamlyYS1zbGFjay1pbnQifQ)
See also Slack convo [here](https://cirg.slack.com/archives/C5EU4L7PF/p1694733386101599)
Changes include:
- Whenever an data export is initiated on the UI, the frontend will cache the celery task URL associated with the export
- Address the scenario where the export is left running in the background, e.g. as when user switches to a different browser tab - when data export window receives focus again, e.g. as when the user comes back from a different browser tab, the frontend will now check if the download has finished by calling the celery task URL cached from the previous step and update the download link on UI if it has finished successfully.
- other change - enhance error checking

### DIFF
--- a/portal/static/js/src/components/ExportInstruments.vue
+++ b/portal/static/js/src/components/ExportInstruments.vue
@@ -116,8 +116,8 @@
                 //watch for when study changes
                 //reset last exported item link as it is specific to each study
                 this.handleSetExportHistory();
-                //reset export error
-                this.resetExportError();
+                //reset export display info
+                this.resetExportInfoUI();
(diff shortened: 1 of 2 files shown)


mcjustin (approved):

pbugni (approved):
thank you!

PR #4342, opened 2023-10-03
Fixup Docker Image Building

ivan-c:
- Update debian base image from oldoldstable (buster) to oldstable (bullseye)
  - current stable (bookworm) fails to build

We'll probably want to switch to our general (simpler) approach for building python images and remove [`dh-virtualenv`](https://github.com/spotify/dh-virtualenv), instead of upgrading to bookworm


pbugni (approved):
bullseye!  🎯

PR #4343, opened 2023-10-06
IRONN-210 allow for change of withdrawal dates

pbugni:
This change allows for an update to an existing withdrawn consent.  Previously, a 404 was generated if the user didn't have a valid matching consent when a withdrawal was requested.

It is required that the requested withdrawal date is:
- greater than the last valid consent for the research study
- less than 1 day in the future (to allow for any timezone details)

### DIFF
--- a/portal/views/user.py
+++ b/portal/views/user.py
@@ -1,5 +1,5 @@
 """User API view functions"""
-from datetime import datetime
+from datetime import datetime, timedelta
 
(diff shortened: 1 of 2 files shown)


pbugni:
grr, the select for review didn't go through last night, sorry 'bout that.

mcjustin (approved):

PR #4346, opened 2023-10-11
Fixup python3 path

ivan-c:
- Update version-specific python3 paths to bookworm release of python3
- fixes #4342

### DIFF
--- a/debian/portal.triggers
+++ b/debian/portal.triggers
@@ -1,4 +1,4 @@
-# Register interest in Python interpreter changes (Python 2 for now); and
+# Register interest in Python interpreter changes (Python 3 for now); and
 # don't make the Python package dependent on the virtualenv package
 # processing (noawait)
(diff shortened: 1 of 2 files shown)


debian/portal.triggers
```diff
@@ -1,7 +1,7 @@
 # Register interest in Python interpreter changes (Python 2 for now); and
```
pbugni, line 1: dated comment (Python 2), but not urgent.

pbugni (approved):
would rather see a variable used for python version, but i assume this is not easy/available, or you would have done so.

debian/portal.triggers
ivan-c, line 1:
```suggestion
# Register interest in Python interpreter changes (Python 3 for now); and
```

ivan-c: Thanks! fixed in 8be5066

PR #4349, opened 2023-11-08
TN-3267 Remove Zulu

mcjustin:
Only Groote Schuur Hospital has been using it to date.  https://movember.atlassian.net/browse/TN-3267 .

See the description at that story, we actually removed Zulu in 2021 but earlier this year were asked to use it for Groote Schuur. 

I think we'll need a migration to switch any user from Zulu to English, like we did for orgs in 2021 https://github.com/uwcirg/truenth-portal/pull/4086 (and we'll need to run that existing migration again)

PR #4353, opened 2023-12-05
migration to remove Zulu locale.

pbugni:
Migration for PR #4349

### DIFF
--- a/portal/migrations/versions/d1f3ed8d16ef_.py
+++ b/portal/migrations/versions/d1f3ed8d16ef_.py
@@ -1,0 +1,43 @@
+"""remove Zulu locale
+
+Revision ID: d1f3ed8d16ef
+Revises: 80c3b1e96c45
+Create Date: 2023-12-05 14:09:10.442328
+
+"""
+from alembic import op
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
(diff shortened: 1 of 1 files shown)


mcjustin (approved):

portal/migrations/versions/d1f3ed8d16ef_.py
```diff
@@ -0,0 +1,43 @@
...
+        "delete from codeable_concept_codings where coding_id in "
+        "(select id from codings where code = 'zu_ZA')")
+
+    # remove Zulu from any orgs
+    op.execute(
+        "delete from organization_locales where coding_id = (select id from codings where code = 'zu_ZA')"
+    )
+
+    # remove the Zulu coding
+    op.execute("delete from codings where code = 'zu_ZA'")
```
ivan-c, line 38: No strong feelings, but I thought we planned on removing Zulu only from prod sites, but this will remove the ability to select Zulu entirely (even on Test Sites)

I'm also not 100% sure this will work, since [Zulu is still in Coding.json](https://github.com/uwcirg/truenth-portal/blob/v23.10.12.2/portal/config/eproms/Coding.json#L6117-L6122) and will be reloaded at next `flask sync`

ivan-c (approved):
Looks good, but think there may be issues with the JSON files vs DB

PR #4354, opened 2023-12-12
IRONN-225 EMPRO month 12 expired vs. not available

pbugni:
fix for https://movember.atlassian.net/browse/IRONN-225

EMPRO qb_status expired status is ambiguous - can mean "not yet available" or "expired".  Added comparison with consent to correct in adherence report.

### DIFF
--- a/portal/models/reporting.py
+++ b/portal/models/reporting.py
@@ -168,8 +168,8 @@ def single_patient_adherence_data(patient, as_of_date, research_study_id):
             ts_reporting = TriggerStatesReporting(patient_id=patient.id)
             empro_row_detail(row, ts_reporting)
 
-        # latest is only valid for a week, unless the user withdrew
-        valid_for = 500 if row['status'] in ('Expired', 'Withdrawn') else 7
+        # latest is only valid for a day, unless the user withdrew
+        valid_for = 30 if row['status'] in ('Expired', 'Withdrawn') else 1
         AdherenceData.persist(
             patient_id=patient.id,
             rs_id_visit=rs_visit,
@@ -181,9 +181,10 @@ def single_patient_adherence_data(patient, as_of_date, research_study_id):
     for qbd, status in qb_stats.older_qbds(last_viable):
         rs_visit = AdherenceData.rs_visit_string(
             research_study_id, visit_name(qbd))
-        # once we find cached_data, the rest of the user's history is good
+        # once we find cached_data, the rest of the user's history is likely
+        # good, but best to verify nothing is stale
         if AdherenceData.fetch(patient_id=patient.id, rs_id_visit=rs_visit):
-            break
+            continue
 
         historic = row.copy()
         historic['status'] = status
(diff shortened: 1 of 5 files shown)


pbugni:
Realizing we'll need a migration to catch cached adherence data for the affected patients.

portal/models/reporting.py
```diff
@@ -152,10 +152,14 @@ def empro_row_detail(row, ts_reporting):
     # build up data until we find valid cache for patient's history
     status = str(qb_stats.overall_status)
     row = patient_data(patient)
+    row["status"] = status
     if status == "Expired" and research_study_id == EMPRO_RS_ID:
-        row["status"] = "Not Yet Available"
-    else:
-        row["status"] = status
```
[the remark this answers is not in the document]
pbugni: No, `status` will contain that value as well as others like `Withdrawn` and will fall through to the `else` clause.

mcjustin (approved):

portal/models/reporting.py
```diff
@@ -152,10 +152,14 @@ def empro_row_detail(row, ts_reporting):
...
+        # Expired status ambiguous for EMPRO - either not available
+        # due to complex business rules around start or walked off
+        # the end.  Assume if consent + 1year > now, it's the former.
+        consent = datetime.strptime(row["consent"], "%d-%b-%Y %H:%M:%S")
+        if consent + timedelta(days=365) > as_of_date:
```
mcjustin, line 161: Per https://promowiki.movember.com/pages/viewpage.action?pageId=72941826#SequencingofPROMs-EMPROQUESTIONNAIRE , the actual 12 month expiration date can be significantly later than this.

1. *"If the subject consents to the IRONMAN and EMPRO studies on the same day: the EMPRO questionnaire is made available for four weeks from the date the subject completes their baseline IRONMAN questionnaire"*. So, this cutoff date can be up to 4 weeks too early in this situation.
2. *"If the subject consents to the EMPRO study after consenting to the IRONMAN study... (and) If the subject consents to the EMPRO study more than four weeks after completing an IRONMAN questionnaire: The EMPRO questionnaires are made available for four weeks from the date the subject completes their next IRONMAN questionnaire"* - this may be even longer than four weeks later, as I read it.

Am I reading this correctly? Would we report "Expired" to soon in these situations?

If so, can we look to the timeline for a more accurate date here? If not, we'll need to explain to PCCTC how to interpret this.

pbugni: hmm, yes, you are correct.  i'll need to work that in, thanks for remembering this detail @mcjustin

pbugni: also of note, found in testing scenarios where `row["consent"] == None` need to account for this as well.

pbugni:
@mcjustin I just requested a fresh review after fixing the issue you found above.  testing revealed other details i resolved as well.

mcjustin (approved):

portal/models/reporting.py
mcjustin: As I read it, you've addressed this - thanks

PR #4363, opened 2024-02-14
Adherence to always report withdrawal and completed after

pbugni:
Adherence report changes:
- always include a row for withdrawal, and use 'completed date' column for date of withdrawal
- include completed visits even if they are completed after date of withdrawal

qb_timeline changes:
- include rows after withdrawal for implementation of the above

migration changes:
- now that we must rebuild the timeline and adherence data for every withdrawn user, it takes far too long to run and would render the service unavailable until the migration completes.  therefore, it only removes the timelines and adherence data for that set of users, and will rebuild when first requested.

NB: this PR merges into another branch, such that all may be a single update on the develop branch.


portal/migrations/versions/3c871e710277_.py
```diff
@@ -152,49 +157,68 @@ def upgrade():
...
-            b4_state = capture_patient_state(patient_id)
-            update_users_QBT(
-                patient_id,
-                research_study_id=study_id,
-                invalidate_existing=True)
-            _, _, _, any_changes = present_before_after_state(
-                patient_id, study_id, b4_state)
-            if not any_changes:
-                continue
+            if slow_report_details:
```
mcjustin, line 160: I don't see this ever being set to true...

mcjustin (approved):
Thanks, just had one question out of curiosity.

portal/migrations/versions/3c871e710277_.py
pbugni: > I don't see this ever being set to true...

no, that was just a convenient way to fence out that code, in case we wanted to review the noisy output.  that's exactly how i generated the files attached to IRONN-210 - but given that we're now hitting every withdrawn patient, it's an overwhelming amount of data (and more than doubles the time).

mcjustin: 👍

portal/config/eproms/Questionnaire.json
```diff
@@ -5851,6 +5851,42 @@
                 "display": "Other",
                 "code": "irondemog_v3.26.8"
               }
+            },
```
ivan-c, line 5854: (this was in #4364 )

ivan-c (approved):

PR #4364, opened 2024-03-04
TN-3264 Add options to irondemog_v3

mcjustin:
TN-3264 Adds 6 new options to irondemog_v3.26 (ethnic group)

### DIFF
--- a/portal/config/eproms/Questionnaire.json
+++ b/portal/config/eproms/Questionnaire.json
@@ -5851,6 +5851,42 @@
                 "display": "Other",
                 "code": "irondemog_v3.26.8"
               }
+            },
+            {
+              "valueCoding": {
+                "display": "African",
+                "code": "irondemog_v3.26.9"
+              }
+            },
+            {
+              "valueCoding": {
+                "display": "Black",
+                "code": "irondemog_v3.26.10"
+              }
+            },
+            {
+              "valueCoding": {
+                "display": "Coloured",
+                "code": "irondemog_v3.26.11"
+              }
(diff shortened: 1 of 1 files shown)


mcjustin:
@ivan-c  is there anything beyond the usual automated deploy to test that needs to be done for this?

mcjustin:
> @ivan-c is there anything beyond the usual automated deploy to test that needs to be done for this?

Never mind, I see that it was picked up at https://eproms-test.cirg.washington.edu/api/questionnaire 👍

portal/config/eproms/Questionnaire.json
```diff
@@ -5851,6 +5851,42 @@
                 "display": "Other",
                 "code": "irondemog_v3.26.8"
               }
+            },
+            {
+              "valueCoding": {
+                "display": "African",
```
mcjustin, line 5857: @ivan-c  I'm removing this "African" option from the AE, as the only sites that use it changed their mind and no longer want it (https://movember.atlassian.net/browse/TN-3285). I expect that I should leave it here, since it has been live for some months and we likely have some data for it. Is that the case?

mcjustin: Thanks @ivan-c  Reading that, I'm wondering if I shouldn't have removed the option from the AE a few weeks ago https://gitlab.cirg.washington.edu/svn/dhair2/-/merge_requests/838/diffs , but you did approve that. Should I restore that (with a condition to either always exclude it, or set sequence = 0)?

PR #4365, opened 2024-03-04
move all redis session creation to factory

pbugni:
This started as an attempt to mock redis for consistent testing outside of normal behavior.

That attempt failed thus far, but to capture all redis session initialization behind a single function still seems of value, for future work and easier debugging.

This PR includes a fence around the cache clearing redis communication necessary in a patient's timeline rebuild, to only hit redis when not under test.  Necessary as otherwise, we experience a deadlock situation where tables can't be dropped between test runs, as postgres believes a deadlock condition exists.

### DIFF
--- a/portal/models/qb_timeline.py
+++ b/portal/models/qb_timeline.py
@@ -4,7 +4,6 @@ from time import sleep
 
 from dateutil.relativedelta import relativedelta
 from flask import current_app
-import redis
 from redis.exceptions import ConnectionError
 from sqlalchemy.types import Enum as SQLA_Enum
 from werkzeug.exceptions import BadRequest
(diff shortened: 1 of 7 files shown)


portal/migrations/versions/3c871e710277_.py
```diff
@@ -0,0 +1,225 @@
+"""Correct user_consent regression issues raised by PR #4343
```
ivan-c, line 1: (this was in #4363)

portal/models/qb_timeline.py
```diff
@@ -786,6 +761,15 @@ def invalidate_users_QBT(user_id, research_study_id):
         for ad in adh_data:
             db.session.delete(ad)
 
+        if not current_app.config.get("TESTING", False):
+            # clear the timeout lock as well, since we need a refresh
+            # after deletion of the adherence data
```
ivan-c, line 766:
```suggestion
            # after deletion of the adherence data
            # otherwise, we experience a deadlock situation where tables can't be dropped 
            # between test runs, as postgres believes a deadlock condition exists
```

ivan-c (approved):
looks good, thanks!

PR #4366, opened 2024-03-06
confirm staff have valid email before attempt to send

pbugni:
found additional cases where test system may be attempting to send EMPRO alert emails to user's w/o a valid email address (staff).

added detailed logging where failure is occurring, as well.


ivan-c (approved):
looks great, thank you!!

PR #4367, opened 2024-03-07
log missing cur_qbd situation

pbugni:
the following was recently found on elk:

```
File "/opt/venvs/portal/lib/python3.9/site-packages/portal/models/qb_status.py", line 102, in _sync_timeline
    if cur_index > 0:
TypeError: '>' not supported between instances of 'NoneType' and 'int'
```
pencil tracing the code, can't think of a logical explanation for the combination of variables that would cause this.  modified code to stop raising exceptions, be more robust and log an error if this happens again with more detail.


ivan-c (approved):
thanks!

PR #4369, opened 2024-03-12
Increase max line length for linters

ivan-c:
- Increase max line length allowed by linting tools

[Some precedents established by others](https://knox.codes/posts/line-length-limits)


pbugni (approved):
thank you

PR #4372, opened 2024-03-25
Feature/opt out post tx report UI

achen2401:
https://movember.atlassian.net/browse/TN-3252
Opt out changes to post intervention report section for clinician:
Per [specs](https://movember.atlassian.net/wiki/spaces/ISS/pages/105744158/Trigger+Logic)

Changes include:
- Not including hard trigger(s) dismissed by the participant in domains requiring action from the clinician (see example screenshot, note "general pain" is not included):
PostInterventionOptOut
- The questionnaire is not available in the total opt out scenario, i.e. the participant has opted out from all hard triggers:
PostInterventionOptOut_NoQ

Other change:
per spec, the participant will be presented opt out modal when a hard trigger for a domain is triggered **3** months in a row


mcjustin (approved):

portal/static/js/src/empro.js
```diff
@@ -488,10 +488,10 @@ emproObj.prototype.processTriggerData = function (data) {
       self.mappedDomains.push(mappedDomain);
     }
     for (let q in data.triggers.domain[key]) {
-      // if sequence count > 1, that means it has been asked before
+      // if sequence count >= 3, that means it has been triggered previously
```
pbugni, line 491: if sequence count >= 3, the user can choose to opt_out of respective domain

pbugni (approved):
Looks great!  Minor comment suggestion.  Thanks

PR #4373, opened 2024-03-26
stop emailing withdrawn

pbugni:
Discovered withdrawn users continue to be sent reminder emails.

- This PR patches a number of incorrect usages of `user.email_ready()` (as it returns a tuple, a simple `if user.email_ready()` check is misleading.
- masks the email address of all withdrawn users, to make certain they aren't sent any additional messages (via a migration)
- blocks another path to hand trigger sending reminders.  I doubt this was being used, but no reason to leave door open


portal/migrations/versions/5caf794c70a7_.py
```diff
@@ -0,0 +1,125 @@
...
+            confirm_withdrawn_row(patient, 0)
+        if withdrawal_e:
+            confirm_withdrawn_row(patient, 1)
+
+        # check for users consented for both but only withdrawn from one
+        if (consent_g and consent_e) and not (withdrawal_g and withdrawal_e):
+            print(f"user {patient_id} consented for both, but only withdrawn from one")
+            continue
+
+        mask_withdrawn_user_email(session, patient, admin)
```
mcjustin, line 115: @pbugni  in slack you wrote "[12 hours ago](https://cirg.slack.com/archives/C5MH9NVKQ/p1711431751401259?thread_ts=1711382681.633649&cid=C5MH9NVKQ)
_one hack that would buy more time: i could write a tiny migration to alter the email address of all withdrawn users, so we’d be certain to not send anything to them.  this would be reversible, and enable us to turn the job back on._"

My read was that you were going to do this out of band as a temporary stopgap. There is the possibility that consent will be restored for some of these - the system needs to back this out if that happens.

pbugni: @mcjustin yes - please see commit 9e46f28 below.  now applying the mask (and removing it) on all consent changes when applicable.

mcjustin (approved):

ivan-c (approved):
Looks good! Appreciate the high-level functional calls in the migration to helpers. Ran on stage with the latest prod data and looked good

PR #4382, opened 2024-06-04
TN-3291 implement caching of research data, for performance improvement

pbugni:
Reports from the field - the generation of research_data has become unreliable due to time it takes to build report.

Now retaining a table of cached values, one row per questionnaire response, for a responsive report generation.

Primary concern will be with cache-invalidation, especially on an org change of research protocol or a user's consent date change.

There is now a scheduled job that runs daily, to seek out any overlooked QNRs missing from this new cache table `research_data`.  A dry run of the prod db:
```
"found 92748 questionnaire responses missing from research_data cache"
[...]
""Task portal.tasks.cache_research_data_task[f87d7130-d2b2-4413-a461-61cf3ebc9594] succeeded in 32897.385058208994s"
```
thereafter, it will generally find zero, as we immediately update on QNR put/post.


portal/models/questionnaire_response.py
```diff
@@ -370,7 +370,8 @@ def quote_double_quote(value):
                         answer['valueCoding'].get('text')
                     )
 
-                    text_and_coded_answers.append({'valueString': text_answer})
+                    if text_answer is not None:
```
pbugni, line 373: @ivan-c this change was necessary to get tests to pass.  i can't imagine it's be work ever saving a `valueString` with a `None`, would it?  those were showing up as differences in the following test:
```
 py.test tests/test_assessment_engine.py::TestAssessmentEngine::test_update_assessment
 ```

ivan-c: > i can't imagine it would ever be worth saving a `valueString` with a value of `None` in a questionnaire response, would it?

probably not- `None`/null isn't the same type as valueString anyway

ivan-c (approved):
Looks good!

PR #4388, opened 2024-07-03
Upgrade Celery

ivan-c:
- Upgrade celery & dependencies to next major version
  - Modified `setup.cfg` and used `pip-compile` to generate a new set of dependencies
- Upgrade version of python used for testing to match container

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -20,7 +20,7 @@ before_script:
 # test templates
 .test_template: &test_definition
   # track debian stable version of python3
-  image: python:3.7
+  image: python:3.9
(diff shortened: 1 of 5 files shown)


pbugni (approved):
Looks clean.  Interesting/inconsistent line breaks around `via` notes, but insignificant.

Thank you for doing this!

ivan-c:
>Interesting/inconsistent line breaks around via notes, but insignificant

I think pip-tools has updated their output to break that down into separate lines. thought about rewriting the whitespace to match the old format, but feels like we'll want to rebuild the whole requirements.txt eventually

PR #4393, opened 2024-07-18
Feature/wait for processed triggers

achen2401:
fix based on Slack convo [here](https://cirg.slack.com/archives/C5MH9NVKQ/p1721261271064199)

- expand frontend retry attempts (will wait for at least 2 minutes and 35 seconds for the triggers to get processed)
- add new hourglass indicator icon (see screenshot), the timing of sand dripping down to the end should coincides with the wait time
Screenshot 2024-07-18 at 4 55 13 PM

- add audit logging for when Optout modal is presented, when Optout modal is dismissed, when maximum retry attempt is reached for retrieving triggers


portal/static/js/src/modules/TnthAjax.js
```diff
@@ -373,6 +373,12 @@ export default { /*global $ */
                 setTimeout(function() {
                     this.getSubStudyTriggers(userId, params, callback);
                 }.bind(this), 1500*params.retryAttempt);
+                if (params.retryAttempt === params.maxTryAttempts) {
+                    this.postAuditLog(userId, {
+                        context: "access",
```
pbugni, line 378: i don't see a great category in the audit/context column, but this isn't a controlled vocab.

perhaps "timeout" or "error" would make more sense.   or stick with "assessment" as that's what is being collected here, so to speak.

pbugni (approved):
thank you!

mcjustin (approved):

PR #4394, opened 2024-07-18
logging to help discover performance problems

pbugni:
series of debug logging statements added to `extract_observations`, to help isolate the performance issue.


pbugni:
done with this branch.  only used to isolate missing indexes causing slowdown

PR #4395, opened 2024-07-30
Update SqlAlchemy

ivan-c:
- Update SqlAlchemy to latest version 1.4 for `postgresql_opt` support
- Disable failing test (timing out)


pbugni (approved):
Nice work, @ivan-c .  I might also run `flask db migrate` with this code in place against an existing db, to confirm the system is in sync, and doesn't think those embedded ENUMs need any change (I hope not, but you know...)

PR #4396, opened 2024-08-07
additional database indexes as per TN-3308

pbugni:
Using the table in [TN-3308](https://movember.atlassian.net/browse/TN-3308) as a guide, pencil traced the code to find the common queries on each respective table.

I'd then assemble a query using `flask shell` to get the exact SQLAlchemy syntax, and then attempt before and after `explain $query` times, to confirm it would be of value.

Note the few in the migration that are directly added via SQL `exec` as the SQLAlchemy syntax didn't support (at least I couldn't find a way).


mcjustin (approved):

ivan-c (approved):
looks good, thanks!

I re-ran the `seq_vs_idx` query again and most of the highest-ranked offenders are dealt with in this PR, except `user_observations`, which is now #1

pbugni:
Thank you @ivan-c , great catch on the user_observations table.

PR #4398, opened 2024-08-15
upgrade default postgresql version

ivan-c:
upgrade default postgresql version from 13 to 16


ivan-c:
@pbugni gonna merge this, I've made sure `POSTGRES_VERSION` is overridden on prod, so we don't accidentally upgrade when this rolls out

PR #4408, opened 2024-09-26
TN-3292 fix performance of patient list

achen2401:
TN-3292, enhance performance of patients list, by moving to server-side pagination.  Each respective page now requested via ajax


ivan-c (approved):
Looks great, thanks for keeping the tests working!!

PR #4409, opened 2024-10-03
TN-3319 extend wait time for adherence report export task

achen2401:
https://movember.atlassian.net/browse/TN-3319?atlOrigin=eyJpIjoiOTYwODgxZDM5MTBhNGRkODhlNGViYTRhNDJiMjQwYzQiLCJwIjoiaiJ9

- increase wait time for celery task to start (progressing from PENDING status) from 1 minute to 5 minutes.
- log as error if waiting for more than 5 minutes.

screenshot for when 5 minutes have reached:
Screenshot 2024-10-03 at 3 09 40 PM

### DIFF
--- a/portal/static/js/src/admin.js
+++ b/portal/static/js/src/admin.js
@@ -342,6 +342,8 @@ import {EPROMS_MAIN_STUDY_ID, EPROMS_SUBSTUDY_ID} from "./data/common/consts.js"
                                     "error": true,
                                     "message": i18next.t("More than 5 minutes spent in pending status.")
                                 });
+                                //log error
+                                tnthAjax.reportError(self.userId, window.location.pathname, "Request to export report data failed. More than 5 minutes spent in pending status.");
                                 return;
                             }
                         }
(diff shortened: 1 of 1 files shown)


portal/static/js/src/admin.js
```diff
@@ -334,13 +334,13 @@ import {EPROMS_MAIN_STUDY_ID, EPROMS_SUBSTUDY_ID} from "./data/common/consts.js"
...
-                            if (passedTime > 60) {
-                                //more than a minute passed and the task is still in PENDING status
+                            if (passedTime > 300) {
+                                //more than 5 minutes passed and the task is still in PENDING status
                                 //never advanced to PROGRESS to start the export process
                                 //abort
                                 self.onAfterExportReportData({
                                     "error": true,
-                                    "message": i18next.t("More than a minute spent in pending status.")
+                                    "message": i18next.t("More than 5 minutes spent in pending status.")
```
pbugni, line 343: @achen2401 would it be possible to log an error when this happens, as we should be alerted if it's taking more than 5 mins?

pbugni (changes requested):

portal/static/js/src/admin.js
achen2401: thank you @pbugni.  Logging added.

pbugni (approved):

PR #4411, opened 2024-10-14
Fixup/table preference

achen2401:
Address a stream of 400 errors (CSRF token is missing) related to saving patientlist table preferences.
Cause:
I think the error stemmed from when an ajax request was first made to retrieve patient data, which is precedes by a call to save table preferences, the document DOM was not ready (which includes the value of CSRF token in an hidden element) - hence the missing CSRF token error.
Fix:
To add check for DOM readiness before the request to save table preferences is made.


mcjustin (approved):
Thanks for identifying and digging into this @achen2401 , I've had problems with these preferences in the past.

ivan-c (approved):

PR #4413, opened 2024-10-22
fix frontend error handing of patient list ajax error

achen2401:
part of  https://movember.atlassian.net/browse/TN-3324

Currently UI is a blank screen when the ajax call to retrieve patient list data encounters error:
Screenshot 2024-10-22 at 11 08 58 AM

Added fix to error handling with error displaying on screen and allowing user to change/clear filters on the UI if need to
Screenshot 2024-10-22 at 11 13 29 AM

### DIFF
--- a/portal/static/js/src/admin.js
+++ b/portal/static/js/src/admin.js
@@ -143,6 +143,12 @@ let requestTimerId = 0;
           errorElement.innerHTML = errorMessage;
         }
       },
+      clearError: function() {
+        var errorElement = document.getElementById("admin-table-error-message");
+        if (errorElement) {
+          errorElement.innerHTML = "";
+        }
+      },
       injectDependencies: function () {
         var self = this;
         window.portalModules =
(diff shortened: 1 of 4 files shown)


mcjustin:
> part of https://movember.atlassian.net/browse/TN-3324
> 
> Currently UI is a blank screen when the ajax call to retrieve patient list data encounters error...
> Added fix to error handling with error displaying on screen and allowing user to change/clear filters on the UI if need to ...

Thanks @achen2401 . In your screenshot, it looks like the user wouldn't be able to clear the birthdate filter, which was the one causing the problem at https://movember.atlassian.net/browse/TN-3324 . I think this change would still be an improvement if that's the case, but I just want to clarify.

achen2401:
> In your screenshot, it looks like the user wouldn't be able to clear the birthdate filter, which was the one causing the problem at https://movember.atlassian.net/browse/TN-3324 . I think this change would still be an improvement if that's the case, but I just want to clarify.
@pbugni right, the user won't be able to clear the birthdate filter, but can see that error has occurred.  This is an improvement, as  the user can then try to clear other filter(s) (in the case of the screenshot, the study ID field), then the patient list data will be returned since the filters have been modified.

PR #4416, opened 2024-10-23
Fixes found in failing testing details

pbugni:
Addressed a number of warnings in the tests, and a few random non-repeatable failures.  Tests look to be more stable again.

This does include the elimination of "integration_tests" which used selenium to attempt some basic browser login and such verification tests.  They are fragile, and of little use given our required test plan before any release.


ivan-c (approved):
Looks good!

I may have to revive the browser tests sometime soon, but I'll likely invoke them separately from the rest of the test suite (since LiveServerTestCase was so fragile)

PR #4419, opened 2024-10-24
robust handling of unsortable patient list attributes (develop target)

pbugni:
Same PR as #4414, but this one correctly merges into develop.


pbugni:
not requesting reviews, as the same code was reviewed in #4414.

PR #4423, opened 2024-11-08
IRONN-270 hotfix. 

pbugni:
See also https://movember.atlassian.net/browse/IRONN-270
 
The `TimeoutLock` used to prevent concurrent EMPRO trigger states overwrites was expiring prior to completion.

Elaborate chain of events needed to reproduce the problem, but the existing 60 second timeout for any single user's updates to their trigger_states->triggers data was inadequate when other jobs (such as `update_patients_task`) came in before completion of trigger processing.

The end result was an area of code intended to be locked in a critical section, would obtain a fresh lock even if one was in use, when it had been longer than the expiration time.

This PR gives more than adequate space (2 hours rather than 60 seconds).

### DIFF
--- a/portal/trigger_states/empro_states.py
+++ b/portal/trigger_states/empro_states.py
@@ -26,6 +26,7 @@ from ..models.user import User
 from ..timeout_lock import TimeoutLock, LockTimeout
 
 EMPRO_LOCK_KEY = "empro-trigger-state-lock-{user_id}"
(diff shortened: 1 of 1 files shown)


mcjustin (approved):

ivan-c (approved):
Looks good; not a big deal, but should merge into master if it's a hotfix

ivan-c (approved):

PR #4424, opened 2024-11-20
IRONN-264 visit_month was being cached in translated form.

pbugni:
looking up visit_month now always returns english version, which may be cached.  `translate_visit_month()` used by front end where needed.


ivan-c (approved):
Looks good!

PR #4433, opened 2025-03-05
Downgrade setuptools-scm

ivan-c:
Pin setuptools-scm to older version to prevent conflict with setuptools

>ERROR: setuptools==44.1.1 is used in combination with setuptools-scm>=8.x


pbugni (approved):
Thank you for resolving this!

PR #4438, opened 2025-04-02
IRONN-285 add clinician role for accessing patients/page page

achen2401:
address https://movember.atlassian.net/browse/IRONN-285?atlOrigin=eyJpIjoiMzI0NjQxMWI1ZmNmNDQ5Nzg5NjZlNTVjNGMxYTY3ZDciLCJwIjoiaiJ9&linkSource=email

- allow user with Clinician role only to access the `patients/page` page (this is the API endpoint that the frontend hits to retrieve records for each page of the patients list)

### DIFF
--- a/portal/views/patients.py
+++ b/portal/views/patients.py
@@ -140,6 +140,7 @@ def sort_query(query, sort_column, direction):
 
 @patients.route("/page", methods=["GET"])
 @roles_required([
+    ROLE.CLINICIAN.value,
     ROLE.INTERVENTION_STAFF.value,
     ROLE.STAFF.value,
     ROLE.STAFF_ADMIN.value])


mcjustin:
@achen2401  can you please grant the same to `PRIMARY_INVESTIGATOR`?

pbugni:
Hmm, I'm not so sure about this fix.

We didn't previously give clinicians (or PIs, who i believe are also always clinicians) access to the patient list, which is what this does.

Did that requirement change?

achen2401:
> @achen2401 can you please grant the same to `PRIMARY_INVESTIGATOR`?

@mcjustin I think `clinician` role probably suffices as when an account is granted the PRIMARY_INVESTIGATOR role, the code will check whether the Clinician, Staff or Staff Admin role is also added:

Screenshot 2025-04-02 at 4 31 06 PM

pbugni (approved):
Good fix, thank you @achen2401 .

When setting up server side pagination, the roles used for top level `/patients` view were copied - we missed the need for `/patients/substudy` to be able to request a page of patients when the user only had the `CLINICIAN` role.

mcjustin (approved):

PR #4440, opened 2025-04-29
Upgrade to docker compose v2

ivan-c:
`docker-compose` v1 is deprecated; switch invocations to `docker compose`


pbugni (approved):
Nice job.  Appreciate the additional cleanup.

Given the volume change and potential loss... should we remove `docker-compose` from the standard path on prod systems?  Probably overkill, just paranoid.

ivan-c:
> Given the volume change and potential loss... should we remove `docker-compose` from the standard path on prod systems? Probably overkill, just paranoid.

@pbugni 
probably a good idea, but unfortunately docker-compose v1 is installed and managed by puppet. I took a quick look but couldn't figure out how it was configured per VM. will see how feasible that is for newer VMs at least

PR #4443, opened 2025-06-25
adherence vs trigger states check

pbugni:
New CLI entry point `adherence_cache_test` available, to check current status.  Reports outliers to STDOUT.

Example output:
```
Differences for 9191 adherence | trigger states
        month 6:         Completed != triggered (Matching state pending clinician follow-up)
```

This feature came to pass in the process of debugging and looking for additional issues from https://movember.atlassian.net/browse/IRONN-287


portal/trigger_states/adherence_cache_validation.py
```diff
@@ -0,0 +1,162 @@
...
+    def report(self, message):
+        if not self.report_called:
+            self.report_called = True
+            print(f"Differences for {self.patient_id} adherence | trigger states")
+        print(message)
+
+    def show_differences(self):
+        if self.patient_id == 9090:
+            pass
+            #import pdb; pdb.set_trace()
```
ivan-c, line 89: leftover debugging?
```suggestion
```

portal/trigger_states/adherence_cache_validation.py
```diff
@@ -0,0 +1,162 @@
...
+                continue
+            if ad and not ts:
+                if ad in ('Due', 'Expired', 'Overdue'):
+                    # trigger states often never started w/o user intervention
+                    continue
+                if ad == 'Withdrawn':
+                    # trigger states often not generated after withdrawal
+                    continue
+                if self.patient_id == 5903:
+                    # one of a kind, missing trigger_states rows for visit 11
```
ivan-c, line 105: instead of referencing this issue by patient ID, would it be possible to ensure there's no more than 1 of these?

mcjustin (approved):
As I read it `adherence_cache_test` reports on all patients, which is what I expect tho your output only shows one - if it's only one let's talk

pbugni:
> As I read it `adherence_cache_test` reports on all patients, which is what I expect tho your output only shows one - if it's only one let's talk

it evaluates all EMPRO patients.  that example output was complete at the time, as it only found a single patient with adherence cache data that didn't match expectations from the trigger_states table.

PR #4447, opened 2025-07-25
Add frontend handling of  auth/csrf token errors in request

achen2401:
see [Slack convo](https://cirg.slack.com/archives/C5EU4L7PF/p1753460068004999)
- catch unauthorized or CSRF token related error before re-attempting an ajax request


mcjustin (approved):
Thanks for the quick work @achen2401 !  Have you tested it locally?

PR #4450, opened 2025-08-15
research-data validation (IRONN-293)

pbugni:
Discovered in [IRONN-293](https://movember.atlassian.net/browse/IRONN-293), some 0.6% of applicable questionnaire responses were missing from the `research_data` table, which is used to generate research reports.

A new command line utility will generate a report of any rows missing, and if asked to `reprocess` correct any errors found.

Example run from our standard docker deploy:
```
dc exec web flask research-data-cache-test --reprocess
```

### DIFF
--- a/manage.py
+++ b/manage.py
@@ -111,7 +111,7 @@ def upgrade_db():
 @click.option(
     '--research_study_id', '-s',
     type=click.IntRange(0,1), default=0,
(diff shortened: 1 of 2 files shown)


mcjustin (approved):
Sounds great, just flagging https://github.com/uwcirg/truenth-portal/pull/4449#issuecomment-3192435998

PR #4480, opened 2026-02-24
use file rather than system memory for report building

pbugni:
the `research_report_task` appears to be the culprit behind the production reboots experienced lately, when the [VM runs low on memory](https://app.shortcut.com/cirg-sysadmin/story/1109/truenth-prod-server-reboots-2025-12-26-2026-01-09#activity-1137).

these research reports are massive and growing, including all questionnaire data from the life of the service.  the current generation process pulls all the data into memory and runs a number of in memory transformations before streaming the result as a file to the user.

refactoring the process to build out the reports directly on the file system, for streaming from there rather than exhausting memory.


pbugni:
Changed base to `master` for hotfix deploy.

ivan-c (approved):
looks good!

switching the base from develop to master brought in some develop changes. I tried to rebase but gave up. probably won't matter

PR #4491, opened 2026-03-17
Hotfix/port timeline new cache

pbugni:
The `/timeline` view raises server errors as the code hasn't been ported to use the filesystem results from `aggregate_responses()` as altered in #4480 

Corrected via new helper function `bundle_from_aggregate_responses()`, used in both previously refactored tests and the `/timeline` view.


ivan-c (approved):
Looks good!

Tested on eproms-stage and confirmed the exception was gone

PR #4500, opened 2026-04-27
IRONN-316 purge old temp files

pbugni:
As per IRONN-316, avoid production disk issues by purging reports over a week old.


portal/tasks.py
```diff
@@ -427,6 +427,8 @@ def update_tous_task(**kwargs):
 def token_watchdog(**kwargs):
     """Clean up stale tokens and alert service sponsors if nearly expired"""
```
ivan-c:
```suggestion
    """Clean up stale reports and tokens. alert service sponsors if nearly expired"""
```

ivan-c (approved):

PR #4512, opened 2026-07-10
Bugfix/package json fix

achen2401:
- Reverting [commit](https://github.com/uwcirg/truenth-portal/commit/67c832b37ea1693e3bb36fa566f9fea219587047) that upgrades babel from 7 to 8 as it causes runtime errors when I was running my instance.
- Add `terser-webpack-plugin` as dependency based on the change seen in package-lock.json
 the plugin was internally available with webpack previously:
```
"node_modules/webpack": {
  "version": "5.105.0",
  ...
  "dependencies": {
    ...
    "terser-webpack-plugin": "^5.3.16",
    ...
  }
```
but now resolved to:
```
"node_modules/terser-webpack-plugin": {
  "version": "5.4.0",
  "resolved": "https://registry.npmjs.org/terser-webpack-plugin/-/terser-webpack-plugin-5.4.0.tgz",
  "dependencies": {
    "@jridgewell/trace-mapping": "^0.3.25",
    "jest-worker": "^27.4.5",
    "schema-utils": "^4.3.0",
    "terser": "^5.31.1"
  },
  "peerDependencies": { "webpack": "^5.1.0" }
}
```

### DIFF
--- a/portal/package.json
+++ b/portal/package.json
@@ -47,6 +47,7 @@
     "path": "^0.12.7",
     "replace-in-file": "^8.3.0",
     "style-loader": "^4.0.0",
+    "terser-webpack-plugin": "^5.6.1",
     "vue": "^2.6.10",
     "vue-loader": "^15.7.0",
     "vue-style-loader": "^4.1.2",
--- a/portal/package.json
+++ b/portal/package.json
@@ -24,7 +24,7 @@
   "devDependencies": {
     "@babel/core": "^7.28.0",
     "@babel/preset-env": "^7.28.0",
-    "babel-loader": "^10.1.1",
+    "babel-loader": "^10.0.0",
     "clean-webpack-plugin": "^4.0.0",
     "core-js": "^3.44.0",
     "css-loader": "^7.1.2",
--- a/portal/package.json
+++ b/portal/package.json
@@ -47,7 +47,6 @@
     "path": "^0.12.7",
     "replace-in-file": "^8.3.0",
     "style-loader": "^4.0.0",
-    "terser-webpack-plugin": "^5.6.1",
     "vue": "^2.6.10",
     "vue-loader": "^15.7.0",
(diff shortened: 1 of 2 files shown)


portal/package.json
```diff
@@ -47,11 +47,12 @@
     "path": "^0.12.7",
     "replace-in-file": "^8.3.0",
     "style-loader": "^4.0.0",
+    "terser-webpack-plugin": "^5.6.1",
```
ivan-c, line 50: minor nitpick- I don't *think* we need `terser-webpack-plugin` here since it's already in `package-lock.json`

If I understand correctly, `terser-webpack-plugin` was pulled in as a dependency of a node package that we use directly, but the frontend code for the portal doesn't use `terser-webpack-plugin` directly. ie  `terser-webpack-plugin` is a transitive/transient dependency


I'm not sure if it's the same for the frontend, but at least for the python libraries, I've been trying to only track the direct dependencies we need, and let the package manager (npm/pip) manage transitive dependencies in the lockfiles (package-lock.json/requirements.txt)

```suggestion
```

ivan-c (approved):
Looks good, thank you!

portal/package.json
achen2401: @ivan-c  I believe terser-webpack-plugin is a phantom dependency here. The webpack.config.js calls require("terser-webpack-plugin") directly see [here](https://github.com/uwcirg/truenth-portal/blob/3be8a8d5b5f4328588cf0f1234aeeab33d677799/portal/webpack.config.js#L6), but it was never listed in the package.json. I think it only worked because webpack itself depends on it internally, and npm happened to place it where our code could reach it. That's not guaranteed — npm's internal placement of borrowed packages can shift for reasons unrelated to anything we changed.  I have added here explicitly so the build won't ever break again.

ivan-c: ah, thanks for the explanation!!

PR #4514, opened 2026-07-15
TN-3365 disable scheduled jobs triggering EMPRO emails.

pbugni:
As per TN-3365, turn off scheduled jobs sending emails for EMPRO.

The jobs have been disabled on production, but this PR is necessary to prevent them from being restarted on a reboot & subsequent deploys.


mcjustin (approved):

pbugni:
NB: no longer a hotfix, as all EMPRO shutdown work more likely to become a single release

PR #4515, opened 2026-07-15
TN-3364 don't allow new EMPRO consents via /user/<id>/consent API

pbugni:
TN-3364 don't allow new EMPRO consents.

Calling the API to add EMPRO consent will result in a 400:

```
<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 3.2 Final//EN">
<title>400 Bad Request</title>
<h1>Bad Request</h1>
<p>`research_study_id` with value 1 not permitted</p>
```

### DIFF
--- a/portal/config/eproms/Questionnaire.json
+++ b/portal/config/eproms/Questionnaire.json
@@ -6986,7 +6986,1348 @@
           "value": "physactivity"
         }
       ],
-      "resourceType": "Questionnaire"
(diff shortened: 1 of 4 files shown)


mcjustin (approved):

pbugni:
NB: no longer a hotfix, as all EMPRO shutdown work more likely to become a single release

